PR #12, opened 2015-10-08
Adding layout structure, CSS, and API Reference Controllers

matthewjohnston4:


src/site/WebController.php
```diff
@@ -18,6 +18,15 @@ protected function getRequiredStringParam(string $key): string {
     return $this->parameters[$key];
   }
 
+  protected function getStringParam(
+    string $key, string $default = ""
+  ): string {
```
fredemmott: would prefer for this to return ?string and possibly rename to getOptionalStringParam

fredemmott:
getSideNav() looks like it should move into a superclass, and it and related functions should be async

fredemmott:
Closing - can this be re-submitted as 3 separate parts?

src/site/APIListController.php
```diff
@@ -0,0 +1,54 @@
+<?hh // strict
+
+use HHVM\UserDocumentation\APIIndex;
+
+enum APIProduct: string as string {
+  HACK = 'hack';
+}
+
+final class APIListController extends WebPageController {
+  protected async function getTitle(): Awaitable<string> {
```
SiebelsTim: This doesn't always return

PR #18, opened 2015-10-21
Tweaks to quickstart

jwatzman:


guides/hack/quickstart.md
```diff
@@ -83,8 +83,6 @@ And here is the output:
 -6 + 6i
 ```
 
-**IMPORTANT NOTE**: In many cases, HHVM will run your code just fine even if the typechecker issues errors. They are currently not 100% correlated. For example, if `toString()` was left to return `bool`, the typechecker would issue an error, but you could run the code successfully in HHVM. The problem is that even though HHVM would run the code, you may get unexpected results, which is why the typechecker is so important.
```
JoelMarcey: I am ok pulling this out. But what I said is true, right? Or are you saying we don't want to give users an "out" and that the typechecker should be the source of truth -- at least in the context of a quick start.

jwatzman: Yes, this statement is true, but I don't think it belongs here. We need to be very careful about how we word this and the mental model we create for people.

PR #19, opened 2015-10-22
BuildSteps, Builders, and build.php changes for new site

matthewjohnston4:
Part 1 of 3 of the new site's code.

### DIFF
--- a/src/api-gen/FunctionMarkdownBuilder.php
+++ b/src/api-gen/FunctionMarkdownBuilder.php
@@ -11,9 +11,13 @@ class FunctionMarkdownBuilder {
   private ?DocBlock $docblock;
 
   public function __construct(
-    string $file,
+    ?string $file = null,
+    ?FunctionDocumentation $method = null,
   ) {
     $this->yaml = \Spyc::YAMLLoad($file);
+    if ($method) {
+      $this->yaml['data'] = $method;
+    }
 
     $comment = $this->yaml['data']['docComment'];
     if ($comment !== null) {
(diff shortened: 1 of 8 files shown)


src/api-gen/FunctionMarkdownBuilder.php
```diff
@@ -11,9 +11,13 @@ class FunctionMarkdownBuilder {
   private ?DocBlock $docblock;
 
   public function __construct(
-    string $file,
+    ?string $file = null,
```
JoelMarcey: Why is this nullable `?string`? We are not really checking for the file being `null` in here? I looked at Spyc::YAMLLoad the propogation of function calls end up here: 

```
function loadFromSource ($input) {
    if (!empty($input) && strpos($input, "\n") === false && file_exists($input))
    return file($input);
    return $this->loadFromString($input);
  }
```

`$input` in this case is our `$file`, and if it is `null` it is going to try to load from a string. But we don't load from strings here right?

in other words, we would always want a valid file here, right?

JoelMarcey: Btw, this doesn't affect me from merging this -- which I am going to do right now :) If we need to change it to be non-nullable, we can do that later.

matthewjohnston4: So the MethodMarkdownBuilder re-uses this FunctionMarkdownBuilder because Functions and Methods have the exact same YAML format, however Methods are contained within the Class files, and don't have their own YAML files. So, existing calls to FunctionMarkdownBuilder proceed with $file as normal, but MethodMarkdownBuilder constructs a FunctionMarkdownBuilder with a null $file, and instead passes along the sub-YAML string as $method. 

There is likely to be a more graceful way of doing this perhaps :)

PR #20, opened 2015-10-22
Adding new layout controllers, tweaking Index files, adding Routers f…

matthewjohnston4:
…or new URLs, and adding a typedef. Part 2/3 of newsite.

Depends on https://github.com/hhvm/user-documentation/pull/19

### DIFF
--- a/src/APIIndex.php
+++ b/src/APIIndex.php
@@ -1,0 +1,71 @@
+<?hh
+
+namespace HHVM\UserDocumentation;
+
+class APIIndex {
+  private static function getIndex(
+  ): Map<string, Map<string, APIClassIndexEntry>> {
+    return require(BuildPaths::APIDOCS_INDEX);
+  }
+
+  public static function getProducts(): ImmVector<string> {
+    return self::getIndex()->keys()->toImmVector();
+  }
+  
+  public static function getReferenceForType(
+    string $type,
+  ): Map<string, APIClassIndexEntry> {
+    $index = self::getIndex();
+    invariant(
+      $index->containsKey($type),
+      '%s is not in the index',
+      $type,
+    );
+    return $index[$type];
+  }
+
+  public static function getFileForAPI(
+    string $type,
+    string $api,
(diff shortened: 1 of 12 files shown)


src/APIIndex.php
```diff
@@ -0,0 +1,71 @@
...
+      $type,
+    );
+    return $index[$type];
+  }
+
+  public static function getFileForAPI(
+    string $type,
+    string $api,
+    ?string $method,
+  ) {
```
JoelMarcey: return type? probably a `: string`

src/APIIndex.php
```diff
@@ -0,0 +1,71 @@
...
+      $type,
+    );
+    invariant(
+      $index[$type]->containsKey($api),
+      'Type %s does not contain reference %s',
+      $type,
+      $api,
+    );
+    invariant(
+      isset($index[$type][$api]['path']),
```
JoelMarcey: I think we can use `array_key_exists` on shapes, right?

Maybe `array_key_exists('path', $index[$type][$api])`?

src/APIIndex.php
```diff
@@ -0,0 +1,71 @@
...
+      'API %s of type %s does not contain a path',
+      $api,
+      $type,
+    );
+    if ($method !== null) {
+      invariant(
+        isset($index[$type][$api]['methods']),
```
JoelMarcey: same.

src/APIIndex.php
SiebelsTim: > I think we can use array_key_exists on shapes, right?

shapes are just a parser alias to array if I remember correctly. 

I think `isset()` is considered evil on a typechecking standpoint, so better not use it..

JoelMarcey: Definitely evil in strict. Yeah, I can get around to changing.

matthewjohnston4: I tried using ->containsKey and got typing errors, but array_key_exists might work

PR #21, opened 2015-10-22
CSS for new site

matthewjohnston4:
Part 3/3 (but not dependent on 1 or 2)


JoelMarcey:
Definitely pulling this in as-is. Maximum trust that this is good. :) We can make nit changes from `master`

PR #23, opened 2015-10-29
Fork 5 threads for rendering API

SiebelsTim:
The previous approach was too consuming and this is much more safe and
configurable.

We currently don't parse command line arguments, which could change number of
threads (and log level by the way). Let's do this sometimes later.

A downside to this is that it's more bulky in the code :(

### DIFF
--- a/src/build/APIHTMLBuildStep.php
+++ b/src/build/APIHTMLBuildStep.php
@@ -16,10 +16,40 @@ final class APIHTMLBuildStep extends BuildStep {
     sort($sources);
 
     $list = Vector { };
+    $bucket = Vector { };
+    $threads = 5;
+    // Prepare buckets which will contain the commands to be executed
+    // inside of a fork. Every fork gets its own bucket to work on.
+    for ($i = 0; $i < $threads; ++$i) {
+      $bucket[] = Vector { };
+    }
+    $i = 0;
     foreach ($sources as $input) {
-      Log::v('.');
-      $output = $this->renderFile($input);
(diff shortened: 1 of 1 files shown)


SiebelsTim:
After all, I have no idea what I'm doing. So could please someone have a look over this?

@JoelMarcey I think you are the only one not being on holiday and working on this repo, right? Additionally, you could test this with the amazon AWS Server :). So that we don't merge something non-functional

JoelMarcey:
Another option to was to increase the RAM we use on AWS when we go into production. I think your other code may work fine then; that said, having something like this, which can be configurable (I wonder instead of hardcoding the threads if we can check and use the number of CPU cores - 1 as the number??). But maybe 5 is ok to start with.

I created a staging area for these type of changes so as not to bring down our production side, so we can test it.

cc @fredemmott

SiebelsTim:
The other code will bomb your cpu though (with ~1400 tasks) and eventually start waiting to have time for this. Specifying threads is much nicer overall - on the deploying side and on the developer side.

fredemmott:
I'm thinking I'm going to make md-render take file names on stdin and output progress on stdout, and use actual threads inside it. Should be more memory-efficient and probably faster even without the threading changes.

SiebelsTim:
That's the best. :) What do you think? Close this then?

fredemmott:
Might not need the threading, just re-using the process is a *huge* win: https://github.com/hhvm/user-documentation/commit/7dcf4a77ba5bbdcd6ba3dd06b1dfef7409092c28

PR #24, opened 2015-11-04
Shoddy search

jaredonline:
This is literally the first time I've ever written Hack, so I'm sure I did lots wrong.

Please be merciless in your feedback =D

Signed-off-by: Jared McFarland <jared.online@gmail.com>


LocalConfig.php.example
```diff
@@ -1,8 +0,0 @@
-<?hh // strict
-
-namespace HHVM\UserDocumentation;
-
-abstract final class LocalConfig {
-  const string BUILD_DIR = __DIR__.'/build';
-  const string HHVM_TREE = '/home/fred/hhvm';
-}
```
fredemmott: ?

md-render/Gemfile.lock
```diff
@@ -35,6 +35,3 @@ DEPENDENCIES
   github-markdown
   html-pipeline (~> 1.0)
   pygments.rb
-
-BUNDLED WITH
-   1.10.6
```
fredemmott: ?

LocalConfig.php.example
```diff
@@ -1,8 +0,0 @@
-<?hh // strict
```
jazzdan: We prolly don't want to delete this. :)

src/site/ParamGetter.php
```diff
@@ -0,0 +1,42 @@
+<?hh
+
+// a wrapper around the 'unstrict' $_GET and $_POST vars
+final class ParamGetter {
```
fredemmott: We have this via the PSR7 request interface, and get*Param functions in WebPageController. See changes to WebController and getRequiredStringParam() in WebPageContentController in https://github.com/hhvm/user-documentation/commit/955072a71b5405d2db52de8e8e145a4014c9f398 for an example

fredemmott: You'll need to extend WebPageController in a similar manner to support $_POST

LocalConfig.php.example
JoelMarcey: Yep. Keep this file.

src/site/SearchController.php
```diff
@@ -0,0 +1,71 @@
...
+use HHVM\UserDocumentation\SearchResultSet;
+
+final class SearchController extends WebPageController {
+  private string $term;
+  private bool $term_set;
+
+  public function __construct(ImmMap<string, string> $params) {
+    parent::__construct($params);
+    $this->term = "";
+    $this->term_set = false;
```
fredemmott: - this constructor is no longer valid - hh_client should complain
- not actually needed - could use static initializers instead
- why $term_set = false instead of ?string $term = null

src/site/SearchController.php
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  private function getListFromResultSet(Map<string, string> $result_set): XHPRoot {
+    $list = <ul />;
+    foreach ($result_set as $name => $path) {
+        $item = <li><a href={$path}>{$name}</a></li>;
+        $list->appendChild($item);
+    }
+    if (count($result_set) === 0) {
+        $item = <li><em>No results for this category</em></li>;
```
fredemmott: Nit: bit weird to still use a `<ul><li>` for this - could some other markup be better?

src/site/SearchController.php
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  protected async function getBody(): Awaitable<XHPRoot> {
+    $results = $this->getSearchResults();
+    return 
+      <x:frag>
+        You searched for "{$this->getSearchTerm()}"
+        <h2>Guides</h2>
+        <h3>Hack</h3>
+            {$this->getListFromResultSet($results->getHackGuides())}
```
fredemmott: - 2 space indentation, not 4
 - not actually a child, shouldn't be indented

src/site/SearchController.php
```diff
@@ -0,0 +1,71 @@
...
+        <h3>Interfaces</h3>
+            {$this->getListFromResultSet($results->getInterfaces())}
+        <h3>Functions</h3>
+            {$this->getListFromResultSet($results->getFunctions())}
+      </x:frag>;
+  }
+
+  private function getSearchTerm(): string {
+    if ($this->term_set) {
+        return $this->term;
```
fredemmott: 2space

src/GuidesIndex.php
```diff
@@ -25,6 +25,24 @@ private static function getSummaries(
     return require(BuildPaths::GUIDES_SUMMARY);
   }
 
+  public static function search(string $term, SearchResultSet &$results): void {
```
JoelMarcey: Couldn't you just return a `SearchResultSet` instead of using a reference?

src/APIIndex.php
```diff
@@ -24,6 +24,17 @@ public static function getReferenceForType(
     return $index[$type];
   }
 
+  public static function search(string $term, SearchResultSet &$results): void {
```
SiebelsTim: Objects are passed by reference by default.

src/GuidesIndex.php
fredemmott: No references in hack.

Alternatively, return a vector of results, or if you have different tyeps of results, return your own searchresultset, and do something like:

$combined_results->addResults(Foo::search($term));

src/site/SearchResultSet.php
```diff
@@ -0,0 +1,60 @@
...
+    private Map<string, string> $hhvm_guides;
+    private Map<string, string> $hack_guides;
+
+    public function __construct() {
+        $this->classes = Map {};
+        $this->functions = Map {};
+        $this->interfaces = Map {};
+        $this->hhvm_guides = Map {};
+        $this->hack_guides = Map {};
+    }
```
fredemmott: prefer static initializers to constructor

src/GuidesIndex.php
JoelMarcey: Or wait, won't this be returned by reference anyway?

src/site/SearchResultSet.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function addAPIResult(string $type, string $name): void {
+        if ($type === "class") {
+            $this->classes[$name] = sprintf("/hack/reference/class/%s/", $name);
```
fredemmott: doesn't need fixing for this, but FYI, I mean to add a URI builder - so something like `GuidePageController::getURIBuilder()->setStringParam('class', $name)->GetURI()` to avoid hardcoded paths

src/site/SearchResultSet.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+        if ($type === "function") {
+            $this->functions[$name] = sprintf("/hack/reference/function/%s/", $name);
+        }
+        if ($type === "interface") {
+            $this->interfaces[$name] = sprintf("/hack/reference/interface/%s/", $name);
+        }
```
fredemmott: has this been tested with namespaced stuff? `HH\Asio\null()` in particular can be fun

JoelMarcey:
Thanks for the pull request @jaredonline !! 

:+1: 

I hope you feel most of the comments are constructive (I hope).

src/GuidesIndex.php
SiebelsTim: Here too. 
Also references are banned in strict mode. (Which this file sadly isn't :()

fredemmott: yeah, objects are byref (in PHP too), so the `&` is unneeded - still, the other suggested approaches here would be cleaner than depending on that.

src/site/SearchController.php
```diff
@@ -0,0 +1,71 @@
...
+  private bool $term_set;
+
+  public function __construct(ImmMap<string, string> $params) {
+    parent::__construct($params);
+    $this->term = "";
+    $this->term_set = false;
+  }
+
+  protected async function getTitle(): Awaitable<string> {
+    return 'Search for shit!';
```
SiebelsTim: Ahem.

fredemmott: :poop:

JoelMarcey: LOL

src/site/SearchResultSet.php
jaredonline: It works fine because the pathing uses `.` instead of `\`, for example: `/hack/reference/function/HH.Asio.null/`

jaredonline:
Hey folks! Sorry it took me so long \= I was having a really hard time getting it to run on my local laptop (outside the custom environment @jazzdan setup for us at Etsy).

I'm pretty sure I've addressed all the feedback at this point.

jazzdan:
https://media.riffsy.com/images/eda65eadf312e50e52406da4f51a2906/raw

.gitignore
```diff
@@ -10,6 +10,7 @@ LocalConfig.php
 *.css.map
 /build/
 /public/www.pid
+/.sass-cache/
```
fredemmott: ?

src/APIIndex.php
```diff
@@ -27,6 +27,23 @@ public static function getReferenceForType(
     return $index[$type];
   }
 
+  public static function search(string $term, SearchResultSet $results): SearchResultSet {
+    // This whole method is UNSAFE
+    $index = Shapes::toArray(self::getIndex());
+    foreach ($index as $key => $value) {
+        if (is_array($value)) {
```
fredemmott: 2space indents

src/APIIndex.php
```diff
@@ -27,6 +27,23 @@ public static function getReferenceForType(
...
+            foreach ($value as $name => $entry) {
+                if (is_string($name) && is_string($key)) {
+                    if (strtolower($name) === strtolower($term) || strpos(strtolower($name), strtolower($term))) {
+                        $results->addAPIResult($key, $name);
```
fredemmott: you might want to also special-case dealing with "HH\" here

src/site/SearchController.php
```diff
@@ -0,0 +1,68 @@
...
+use HHVM\UserDocumentation\APIIndex;
+use HHVM\UserDocumentation\SearchResultSet;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+final class SearchController extends WebPageController {
+  private ?string $term = null;
+
+  public async function getTitle(): Awaitable<string> {
+    return 'Search for shit!';
```
fredemmott: :poop:

src/site/SearchResultSet.php
```diff
@@ -0,0 +1,52 @@
...
+namespace HHVM\UserDocumentation;
+
+final class SearchResultSet {
+    private Map<string, string> $classes = Map {};
+    private Map<string, string> $functions = Map {};
+    private Map<string, string> $interfaces = Map {};
+    private Map<string, string> $hhvm_guides = Map {};
+    private Map<string, string> $hack_guides = Map {};
+
+    public function addAPIResult(string $type, string $name): void {
```
fredemmott: Use APIDefinitionType enum (new) and switch statements (so that the typechecker makes sure that you handle all the enum values)

src/site/SearchResultSet.php
```diff
@@ -0,0 +1,52 @@
+<?hh // strict
+
+namespace HHVM\UserDocumentation;
+
+final class SearchResultSet {
+    private Map<string, string> $classes = Map {};
+    private Map<string, string> $functions = Map {};
+    private Map<string, string> $interfaces = Map {};
+    private Map<string, string> $hhvm_guides = Map {};
+    private Map<string, string> $hack_guides = Map {};
```
fredemmott: could be worth doing some typedefs - eg turn this into `Map<DocumentedAPIDefinitionName, SiteURLPath>` ?

fredemmott:
much cleaner, thanks :)

.gitignore
jaredonline: I think I had an old `.sass-cache` dir laying around and I added this so I didn't accidentally push it. I wasn't sure why the sass stuff all disappeared on me, but it looks like it was removed intentionally. I'll remove my local dir and this line.

src/APIIndex.php
jaredonline: What would the special case be? I haven't run into any issues with it so far.

jaredonline:
Okay, I think I've addressed all the latest round of concerns [=

And I removed the :hankey: talk =p

fredemmott:
Going to do some cleanup/changes, will ref this issue

PR #29, opened 2015-11-05
Fix spelling errors

jwatzman:
Mostly found with aspell:

```
$ for file in `find . -name \*.md | grep -v 99`; do aspell check $file; done
```


JoelMarcey:
Thanks....

PR #42, opened 2015-11-11
Async intro rewrite

fredemmott:
- prettier pictures
- consistent about direction of time
- add 'dependency tree' idea + picture
- remove multithreading section
- make it clearer how IO and CPU operations interact

## Before

screencapture-beta-docs-hhvm-com-hack-async-intro-1447200815800

## After

screencapture-172-25-46-196-8080-hack-async-intro-1447200697496


fredemmott:
cc @JoelMarcey @jwatzman @krakjoe

guides/hack/03-async/01-intro.md
```diff
@@ -1,38 +1,32 @@
...
+![Dependency Tree](/images/async/async-dependency.png)
 
-![Synchronous](/images/synchronous.png)
+Code structured like this gets the most benefit from async.
 
-## Parallel Execution
+## Synchronous/Blocking IO: Sequential Execution
 
-This is an optimum state. We have all tasks running at the same time, concurrently. But, PHP and Hack do not support more than one thread of execution.
+If (like most PHP code) you do not use asynchronous programming, each step will be executed one-after-the-other:
```
jwatzman: Maybe something like "If you are using PHP, which does not have asynchronous programming, each step..." -- basically, PHP can't do async, and this might imply otherwise?

OTOH what I suggest might imply that Hack gets async by default or something... not sure of the right way to phrase this. Food for thought.

guides/hack/03-async/01-intro.md
```diff
@@ -1,38 +1,32 @@
...
 
-![Parallel](/images/parallel.png)
+![Sequential Execution](/images/async/async-sequential.png)
 
 ## Asynchronous Execution
 
-This is what async does. Tasks are executed concurrently in the same execution thread, with respect to each other, interleaving instructions (e.g., I/O) for different tasks back and forth.
+This is what Hack's async functionality is. All PHP/Hack code executes in the main request thread, but I/O does not block it, and multiple I/O or other async tasks can execute concurrently. If your code is constructed as a dependency tree, this will lead to various parts of your code transparently interleaving with each other instead of blocking each other:
```
jwatzman: This phrasing is awkward, especially after removing the "multithreaded" section (the removal of which is good :)). This implies that Hack gets async for free. Maybe something about properly-written async Hack code would look like this?

guides/hack/03-async/01-intro.md
```diff
@@ -1,38 +1,32 @@
...
 
-![Asynchronous](/images/asynchronous.png)
+![Asynchronous](/images/async/async-always-busy.png)
+
+Importantly, the order your code executes is not guaranteed - for example, if the CuRL request for Component A is slow, execution of the same code could look more like this:
```
jwatzman: Nit: be consistent on the casing of curl. You do `cURL` below, which I think is canonical?

guides/hack/03-async/01-intro.md
```diff
@@ -40,13 +34,19 @@ Async is not a panacea, however. While executing two async functions can utilize
...
+
+In the example above, the call to `curl_exec` in `curl_A()` is blocking any other processing. Thus, even though `curl_B()` is an independent call from `curl_A()`, it has to sit around waiting for `curl_A()` to finish before beginning its execution.
+
+![No Async](/images/async/curl-synchronous.png)
+
 @@ intro-examples/async-curl.php @@
 
 In this example, we are calling an async-aware version of `curl_exec()`. Thus, in this case, our waiting state is explicitly allowing other I/O operation tasks in the code to occur.
 
 When `curl_A()` hits a call to `HH\Asio\curl_exec`, depending on, for example, the network latency to retrieve results of the CURL, the async infrastructure (the scheduler) looks for other async tasks that could be run. It finds that `curl_B()` is available to execute, so it starts executing that code. When it hits its `HH\Asio\curl_exec()` call, the process is repeated again, and the scheduler will find that our `curl_exec()` call in `curl_B()` is ready for execution once again.
```
jwatzman: > When `curl_A()` hits a call to `HH\Asio\curl_exec`, depending on, for example, the network latency to retrieve results of the CURL, the async infrastructure (the scheduler) looks for other async tasks that could be run.

This sentence is super awkward and confusing. (Also cases curl in yet a third way.)

guides/hack/03-async/01-intro.md
fredemmott: - I'm thinking I'll add an example (with diagram) like running fibonacci + blocking curl + blocking mysql with `HH\Asio\v` to show that it isn't
 - curl and mysqli support asynchronous execution via polling, and things can be cludged on top of this via generators. I picked this phrasing to avoid a flamewar over this, especially as this concept is getting more popular

guides/hack/03-async/01-intro.md
fredemmott: :like: yep, definitely needs rewording after the removal of multithreading.

guides/hack/03-async/01-intro.md
fredemmott: whoops. Will fix. Also will need to rebuild pictures.

guides/hack/03-async/01-intro.md
fredemmott: Except for the capitalization I'll leave this for another diff - I've not changed this content. The synchronous curl stuff above is a move from the top of the file (it went sync curl => overview of async => async curl )

fredemmott:
screencapture-172-25-46-196-8080-hack-async-intro-1447266176235

fredemmott:
whoops, uploaded curl pictures badly apparently

PR #53, opened 2015-11-12
Reword async curl section

fredemmott:
screen shot 2015-11-11 at 5 00 35 pm

### DIFF
--- a/guides/hack/22-async/01-introduction.md
+++ b/guides/hack/22-async/01-introduction.md
@@ -59,19 +59,20 @@ A naive way to make two cURL requests without async could look like this:
 
 @@ introduction-examples/non-async-curl.php @@
 
-In the example above, the call to `curl_exec` in `curl_A()` is blocking any other processing. Thus, even though `curl_B()` is an independent call from `curl_A()`, it has to sit around waiting for `curl_A()` to finish before beginning its execution.
+In the example above, the call to `curl_exec()` in `curl_A()` is blocking any other processing. Thus, even though `curl_B()` is an independent call from `curl_A()`, it has to sit around waiting for `curl_A()` to finish before beginning its execution:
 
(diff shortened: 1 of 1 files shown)


fredemmott:
also cc @jwatzman

guides/hack/22-async/01-introduction.md
```diff
@@ -38,19 +38,20 @@ A naive way to make two cURL requests without async could look like this:
...
+In the example above, the call to `curl_exec()` in `curl_A()` is blocking any other processing. Thus, even though `curl_B()` is an independent call from `curl_A()`, it has to sit around waiting for `curl_A()` to finish before beginning its execution:
 
 ![No Async](/images/async/curl-synchronous.png)
 
+Fortunately, HHVM provides an async version of `curl_exec()`:
+
 @@ introduction-examples/async-curl.php @@
 
-In this example, we are calling an async-aware version of `curl_exec()`. Thus, in this case, our waiting state is explicitly allowing other I/O operation tasks in the code to occur.
+The async version of `curl_exec()` allows the scheduler to run other code while waiting for a response from cURL. As we're also awaiting a call to `curl_B()`, it is likely that the scheduler will choose to call it, which in turn starts another async `curl_exec()`. If `curl_A()`'s HTTP request has finished, the scheduler will resume execution of the `curl_A()` function - but it's likely that the HTTP requests will take longer than that, so the main thread will be idle, and waiting for either of the HTTP requests to finish:
```
jwatzman: I realize we don't want to commit to any particular scheduler behavior, but the amount of "it's likely that" here I think gets in the way. Can we just preface the whole few sentences with "the most likely scheduler execution is as follows" or something like that? I'm not sure that's any better. But you see what I mean?

fredemmott:
Done suggested changes, rebased

PR #54, opened 2015-11-12
Rewrite XHP one-liner

fredemmott:
I don't think 'XML' is a compelling advertisment nowadays :p

Also, Add missing '.' from shapes summary while I'm looking at this

### DIFF
--- a/guides/hack/24-XHP/xhp-summary.txt
+++ b/guides/hack/24-XHP/xhp-summary.txt
@@ -1,1 +1,1 @@
-Use XML in Hack instead of strings to build your UI
+Build your UI with components, automatic escaping, and XSS protection.
--- a/guides/hack/26-shapes/shapes-summary.txt
+++ b/guides/hack/26-shapes/shapes-summary.txt
@@ -1,1 +1,1 @@
-Adding structure and type-checking to arrays used as structs or records
+Adding structure and type-checking to arrays used as structs or records.


fredemmott:
cc @Swahvay - any better suggestions?

This shows up here: http://beta.docs.hhvm.com/hack/

Swahvay:
Maybe something like: "Advanced UI rendering allowing HTML to be set directly in PHP and custom HTML elements to be defined." How about that?

fredemmott:
IMO that's the biggest advantage of XHP over most other methods, but unless you've already been using it for a while, it's not clear why - so not the best 'advertisement'.

SiebelsTim:
XSS safety and simplicity and/or extensibility is probably one of the best things to say considering advertisement.

fredemmott:
@SiebelsTim : got some suggested text?

SiebelsTim:
mhh. I'm really bad at this. 

---

I like emphasizing the *why* in contrast to the *how* better; which your text in this PR is doing already.

To be honest, having HTML inside of PHP is something PHP developers tried to avoid over the years and thus is not something that sounds appealing at first glance. You have to work with it or read some more about it to acknowledge its power. 

---
> Rendering HTML in a simple and extensible way that is XSS safe by default.

Meh.. I said I'm bad at this :P

jwatzman:
Instead of components, something about *composable* components? Twig is really good at the rest -- templates, escaping, XSS, etc. Where every other templating language I've seen falls down is composability.

Swahvay:
What about...

> An easily reusable and customizable UI system that automatically handles all escaping and XSS protection.

I say reusable instead of composable, simply because I think it's a little easier to understand. "Composable" is sort of an implementation adjective whereas I feel "reusable" is more of a relatable desire.

fredemmott:
I'll probably go for that without 'all' - I don't want people to think `<script>{$foo}</script>` is safe.

Swahvay:
Good point.

PR #59, opened 2015-11-12
Trimming the examples in the Async overview

matthewjohnston4:
I've removed any specifics from these examples. Important to note that these overviews are intended for absolute beginners, hence the dumbed-down English. We want to be less concerned with the technical details in these overviews as with selling the 'why use this' of each feature - that said, technical inaccuracies are obviously to be removed!

PR #61, opened 2015-11-12
Drastically cut async overview, reuse parts of async intro

fredemmott:
Conflicts with #59

 - The old content was wrong
 - I'm not sure there's value in duplicating the how it works stuff - just giving a taster of why it's interesting and what code looks like could be enough
 - Including no technical details means that there's no chance that someone can read /just this page/ and think they understand it. This is good because they almost certainly don't. That's what the async guide is for
 - Reuse examples and images to reduce maintainance work


fredemmott:
cc @matthewjohnston4 @jwatzman 

screencapture-172-25-46-196-8080-hack-overview-async-1447356112632

fredemmott:
I also did some changes to the wording to make it more concise.

fredemmott:
New preview:

screencapture-172-25-46-196-8080-hack-overview-async-1447361895006

guides/hack/00-overview/03-async.md
```diff
@@ -1,85 +1,15 @@
...
+Async allows performance improvements by changing the flow to look like this instead:
 
-function makeGetRequest(string $url): string {
-  $ch = curl_init($url);
-  curl_setopt($ch, CURLOPT_HEADER, 0);
-  $content = curl_exec($ch);
-  curl_close($ch);
-  return $content;
-}
+![Asynchronous execution](/images/async/feature-advert-async.png)
```
matthewjohnston4: Would it be possible to retain the highlighting of the performance improvements?

Also the heading below of #### Further Reading

guides/hack/00-overview/03-async.md
```diff
@@ -1,85 +1,15 @@
 ### Async Programming in Hack
```
JoelMarcey: Why are we starting with `###` instead of `#` or `##` at the very least?

I was careful to try to not to get to 4-levels of headings in any of the content.

guides/hack/00-overview/03-async.md
fredemmott: I'm not sure how to highlight the performance improvements more - it's already the only bit of text before the async diagram.

As for further reading, that just felt a little weird given how little content is on this page now.

matthewjohnston4: Further Reading will be good for consistency across the overviews as it's always there. I don't think it matters how short this page is. 

I would reword the leading sentence to say:

"Hack's Async functionality introduces the potential for performance improvements by changing the flow to look like this instead:"

guides/hack/00-overview/03-async.md
matthewjohnston4: I made a comment about this elsewhere, but we have a problem currently where we're autogenerating an <h2> page title for all guides, and then below that we have the <h1> title from the Markdown - we need to figure out whether we ditch the autogen title completely, or downgrade the manual titles to <h2> and upgrade the autogen to <h1>

JoelMarcey: So this is happening with the current user guide content then? Is the rendering wrong?

matthewjohnston4: Yeah - the guide controllers were always deriving a page title from the Guide's filename (in this case for example "Overview: Async") and I think we need to ditch that autogenerated page title, that way having the `<h1>` following them won't be bad for SEO/accessibility purposes.

fredemmott: How about we just go with '#' for top level /in the current file/, ignoring what the website does, then use a filter in md-render to normalize it?

guides/hack/00-overview/03-async.md
fredemmott: went for a shorter hybrid - this seems very verbose

fredemmott:
Given the time and that what we've currently got is incorrect, going to merge this.

 - heading discussion: will file an issue
 - further changes to this: issues or pull requests on top of this

PR #70, opened 2015-11-13
Removed unnecessary backslashes in Collections examples.

psyclaudeZ:
If those backslashes were left there on purpose (like for future codegen etc.), then I'm sorry...


fredemmott:
Why are these unneeded? How has this been tested?

WIthout them, it should be a reference to Hack\UserDocumentation\Collections\Semantics\Examples\Foo\InvalidOperationException

JoelMarcey:
No, these `\` are needed because since we are in a namespace, we need to refer to the root `InvalidOperationException`.

psyclaudeZ:
Sorry for being negligent. I thought it was a basic UI thing since it just didn't look right when code examples were rendered on webpage with an exception preceded by \\.

PR #79, opened 2015-11-15
Make menu click targets full-width

daviddoran:
To me the existing menu styling isn't the user-friendliest for three reasons:

1. Only the text of each item is clickable so "at" and "add" have tiny click targets
2. The grey arrows appear when hovering the h5/h6 and not the click targets (giving the impression that clicking will select that item)
3. The background color of the items doesn't change

Before (note the small click targets and the misleading arrows):

menu - before

After (all click targets are full-width and the arrows only appear when the item is clickable):

menu - after

### DIFF
--- a/public/js/APISideNav.js
+++ b/public/js/APISideNav.js
@@ -76,9 +76,9 @@ var DocNav = React.createClass({displayName: "DocNav",
         React.createElement("h4", {id: currentGroup}, 
           React.createElement("a", {className: 'navItem', href: groupHref}, currentGroup)
         ), 
-        React.createElement("ul", {className: 'subList'}, 
-          navGroupChildren
-        )
+        navGroupChildren.length > 0
+          ? React.createElement("ul", {className: 'subList'}, navGroupChildren)
+          : null
       )
     );
   },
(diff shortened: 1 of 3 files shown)


JoelMarcey:
I am going to leave the review of this to @matthewjohnston4

matthewjohnston4:
@daviddoran this looks good if you can confirm that you've tested this on all the responsive breakpoints too:

+ <480px
+ <1020px
+ <1200px
+ <1400px

JoelMarcey:
@fredemmott Anything you are doing that may affect this?

@daviddoran Are we good with the responsive breakpoint question?

fredemmott:
#80 breaks this. Either should be a trivial (but manual) rebase on top of the other though

daviddoran:
@JoelMarcey Just making a few last tweaks padding/margin tweaks but other than that it's looking good

daviddoran:
Okay, here's > 1400px:
hh - 1440

And <1400px:
hh - 1300

And <1200px:
hh - 1000

And <1000px:

vector 2015-11-16 22-39-49


And <460px:

hh - 450

JoelMarcey:
@daviddoran :+1:  This should be merged.

@fredemmott Can we merge this and then rebase #80? Or would you rather vice-versa?

fredemmott:
merge this - want to get mathew's feedback on #80, and that shouldn't block this.

Issue #78, opened 2015-11-15
Sidebar should have APIs in alphabetical order like the actual content does


JoelMarcey:
<img width="1137" alt="screenshot 2015-11-14 17 43 49" src="https://cloud.githubusercontent.com/assets/3757713/11166729/60c4d154-8af7-11e5-97fc-3459ff4be3fc.png">

PR #80, opened 2015-11-16
Unify API and Guide nav implementations, use more appropriate data

fredemmott:
- removes the 'SideNav.js' entirely
 - JS is no longer aware of the structure of GroupsIndex or APIIndex (giving flexibiltiy to modify them to support search and redirects more easily)

Once this is merged, I'll split out some common stuff into xhp classes - eg `<navbar data={} currentItem={} />`


fredemmott:
I'll also squash this a bit before landing.

JoelMarcey:
@fredemmott Will this by default solve https://github.com/hhvm/user-documentation/issues/78

fredemmott:
Nope, that should probably be sorted out in the index build step. I'll fix that separately and commit direct.

fredemmott:
Here, have a rebase :p

matthewjohnston4:
Perfect, looks good!

PR #82, opened 2015-11-17
Add example output after the code examples

### LINKED ISSUE #47: Show output for all examples
Right now we have `.hhvm.expect[f]` files for all the examples so we can run the test runner on them. But we should also have a `.hhvm.output` or some file (maybe we can get that through the test runner by keeping the raw output) so that we can show output within the guides themselves too.

We could use the `hhvm.expect` files for this, but the `.hhvm.expectf` files have regex wildcards that might not be as useful.

JoelMarcey:
Fixes #47

### DIFF
--- a/md-render/HHVM/UserDocumentation/IncludeExamplesFilter.rb
+++ b/md-render/HHVM/UserDocumentation/IncludeExamplesFilter.rb
@@ -51,7 +51,49 @@ module HHVM
             File.read(full_path),
             language: 'Hack',
           );
-          node.replace(code)
+          code_node = node.replace(code)
+          # If the example has HHVM output associated with it, then print it
+          # after the code example. However, if both a .example.hhvm.out and
+          # .hhvm.expect exist, that is an error
+          output_header = doc.document.create_element(
+            'em',
+            "Output",
+          );
+          if (File.exists? (full_path + ".hhvm.expect")) ^ (File.exists? (full_path + ".example.hhvm.out"))
+            if File.exists? (full_path + ".hhvm.expect")
+              output_path = full_path + ".hhvm.expect"
+            else
+              output_path = full_path + ".example.hhvm.out"
+            end
+            output = doc.document.create_element(
+              'pre',
+              File.read(output_path),
+              language: 'Bash',
+            );
+            code_node.after(output).after(output_header)
+          elsif (File.exists? (full_path + ".hhvm.expect")) && (File.exists? (full_path + ".example.hhvm.out"))
+            text = "Both .hhvm.expect and .example.hhvm.out exist. Remove one: #{full_path}"
+            STDERR.write(
+              text.red.bold + "\n"
+            )
+            warning = doc.document.create_element(
+              'h1',
+              text,
+              style: 'color: red',
+            )
+            code_node.after(warning).after(output_header)
+          elsif File.exists? (full_path + ".hhconfig") # We have some include .php files that we don't want output for. Files that are test run have a .hhconfig
+            text = "Neither .hhvm.expect or .example.hhvm.out exist. Add one: #{full_path}"
+            STDERR.write(
+              text.red.bold + "\n"
+            )
+            warning = doc.document.create_element(
+              'h1',
+              text,
+              style: 'color: red',
+            )
+            code_node.after(warning).after(output_header)
+          end
         end
         doc
       end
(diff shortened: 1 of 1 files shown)


JoelMarcey:
screenshot 2015-11-16 16 46 16

md-render/HHVM/UserDocumentation/IncludeExamplesFilter.rb
```diff
@@ -51,7 +51,21 @@ def call
             File.read(full_path),
             language: 'Hack',
           );
-          node.replace(code)
+          code_node = node.replace(code)
+          # If the example has HHVM output associated with it, then print it
+          # after the code example
+          if File.exists? (full_path + ".example.hhvm.out")
```
fredemmott: Do you want to include .expect too (but not .expectf etc)?

JoelMarcey: Right now, I always expect a `.example.hhvm.out` for output.

I instruct example creators to directly copy the `.expect` file as an `.example.hhvm.out`; 

if there is an `.expectf`, then I should write that you run the the test with `no-clean` and copy the resulting `.out` file into `.example.hhvm.out`.

We could check for `.hhvm.expect` to optionally avoid the copy. Check for `.example.hhvm.out`, if doesn't exist, then check for `.hhvm.expect`.

fredemmott: Yeah, avoiding the copy would be nice. It would also be good if we could check that the .example.hhvm.out match the expectf/expectre

JoelMarcey: > It would also be good if we could check that the .example.hhvm.out match the expectf/expectre

By running the test runner in process here for that particular example? Or some sort of regex check? All of our examples should pass the test runner. So if you produce the `.example.hhvm.out` via a run of the test runner, they would have to match because the test pass.

JoelMarcey:
Maybe I don't understand git well enough :/

I just updated the pull request and pushed the changes. My changes are in https://github.com/hhvm/user-documentation/commit/a70ddfb6fd60cf07287589074eb4bfaea2070b05, but why do I see those other two commits from @fredemmott there?

Are we in sync? Or do I need to do something else?

JoelMarcey:
Heck, I think I know why. I probably rebased against master instead of this branch and pull in the latest commits from master. Ugh. Ok. I will try to fix.

JoelMarcey:
OK. I think we are settled now. Interactive rebase and force push.

md-render/HHVM/UserDocumentation/IncludeExamplesFilter.rb
```diff
@@ -51,7 +51,49 @@ def call
...
+          if (File.exists? (full_path + ".hhvm.expect")) ^ (File.exists? (full_path + ".example.hhvm.out"))
+            if File.exists? (full_path + ".hhvm.expect")
+              output_path = full_path + ".hhvm.expect"
+            else
+              output_path = full_path + ".example.hhvm.out"
+            end
+            output = doc.document.create_element(
+              'pre',
+              File.read(output_path),
+              language: 'Bash',
```
fredemmott: ? Why not text?

JoelMarcey: Oh whoops.

JoelMarcey: Oh wait - I did do that on purpose actually.

This is output like you would see at the shell, from the command line. That's why I chose Bash.

JoelMarcey: I can change it to "text". That's fine too, I guess.

fredemmott: I think bash gives you shell-script highlighting - i.e. it's for files that start with `#!/bin/sh` or `#!/bin/bash`

JoelMarcey:
I committed https://github.com/hhvm/user-documentation/commit/49718bfdd817e14131dc2f820a396d4382acd920 in preparation for this. So the only .example.hhvm.out files that exist now are the ones where we have a `.expectf` file. We can add an option to the test runner called `--keep-explicit-output` that produces a `hhvm.out` file if there is a `.expectf` file. 

But for now we have manually added them in the interim.

Any way to merge this before the push today?

PR #83, opened 2015-11-17
Use backticks for <ul> / <li> to prevent them being interpreted as HTML

daviddoran:
- http://beta.docs.hhvm.com/hack/XHP/guidelines#use-contexts-to-access-higher-level-information
- http://127.0.0.1:8080/hack/XHP/guidelines#use-contexts-to-access-higher-level-information

Before:
xhp guidelines 2015-11-17 14-14-32

After:
xhp guidelines 2015-11-17 14-15-57


JoelMarcey:
nice catch!

PR #98, opened 2015-11-18
Fix links.

manzyuk:
Fix two broken links (one was lacking a closing paren, and one had a redundant extra opening paren).


JoelMarcey:
good eye :)

PR #105, opened 2015-11-18
Declare function as async.

manzyuk:
There is an await in the body, so the function must be declared async.


fredemmott:
This needs a rebase - it would also be good to split out this example (even though it's trivial) so that we automatically test it.

fredemmott:
ping

manzyuk:
Rebased and updated the pull request.

PR #111, opened 2015-11-19
Make the Hack getting started example more approachable

### LINKED ISSUE #37: Make Hack Getting Started example more User Friendly
Instead of `Complex`, maybe do something with `Animals`, etc.

JoelMarcey:
We had `Complex` before. Let's use something like `Animal` and
`Human` to be more relevant to someone new. Use `async` on things
a Human might be able to task switch on (e.g., walk and chew gum).
Those sort of things.

Fixes #37


JoelMarcey:
I am open to suggestions here. I had used `Complex` with a lot of comments to explain things one by one. Here I am doing something possibly more relatable to the masses, with less comment clutter, but links to the various topics that the example uses.

JoelMarcey:
cc: @fredemmott @jamesgpearce

guides/hack/01-getting-started/01-getting-started-examples/first.php
```diff
@@ -1,79 +1,66 @@
...
+  private string $name;
+
+  public function __construct(string $name) {
+    $this->name = $name;
+  }
+
+  public async function eat(Food $type): Awaitable<(string, bool)> {
+    // pretend this is the human eating time; usleep takes microseconds, so
+    // 5000000 is 3 seconds
+    await \HH\Asio\usleep(3000000);
```
JoelMarcey: Is there anything I can do here to make it so that `eat` and `move` are actually really `async`? Right now, they will always finish in 5 seconds, I think. Anyway to show that these two things can finish sooner? But still maintain simplicity. Do we need to? Does it matter?

guides/hack/01-getting-started/01-getting-started-examples/first.php
```diff
@@ -1,79 +1,66 @@
...
+  MEAT = "Meat";
+}
+
+interface Animal {
+  // You can eat and move at the same time so allow them to be async
+  public function eat(Food $type): Awaitable<bool>;
+  public function move(int $distance): Awaitable<void>;
+
+  // You cannot do anything else when you sleep, so it cannot be async
+  public function sleep(): int;
```
fredemmott: ... but another animal could sleep at the same time?

guides/hack/01-getting-started/01-getting-started-examples/first.php
```diff
@@ -1,79 +1,66 @@
...
+  public function __construct(string $name) {
+    $this->name = $name;
+  }
+
+  public async function eat(Food $type): Awaitable<(string, bool)> {
+    // pretend this is the human eating time; usleep takes microseconds, so
+    // 5000000 is 3 seconds
+    await \HH\Asio\usleep(3000000);
+    // I am not going to finish all my vegetables!!
+    $finished = $type === Food::VEGETABLE ? false : true;
```
fredemmott: $finished = $type !== Food::VEGETABLE;

guides/hack/01-getting-started/01-getting-started-examples/first.php
```diff
@@ -1,79 +1,57 @@
...
+  // You can eat and move at the same time so allow them to be async by
+  // returning an Awaitable
+  public function eat(Food $type): Awaitable<bool>;
+  public function move(int $distance): Awaitable<void>;
+}
+
+class Human implements Animal {
+  const ONE_SECOND = 1000000; // usleep takes microseconds
+  private string $name;
+  public function __construct(string $name) {
```
fredemmott: Constructor parameter promotion?

PR #114, opened 2015-11-19
Fix copypasta.

manzyuk:


manzyuk:
> If `Foo<int>` is a subtype of `Foo<num>`, then `Foo` is covariant on `T`.

Isn't it backwards: if `Foo<T>` is covariant on `T`, then `Foo<int>` is a subtype of `Foo<num>`, but for `Foo<T>` to be covariant on `T` the type `Foo<Tu>` must be a subtype of `Foo<Tv>` for any type `Tu` that is a subtype of `Tv`.  The above sentence reads as a definition but it is rather an example.  I realize it is a complex topic that requires careful definition of the subtyping relation, but we still could be less hand-wavy here.

JoelMarcey:
Your copypasta change is correct. I am going to merge it.

> Isn't it backwards: if Foo<T> is covariant on T, then Foo<int> is a subtype of Foo<num>

Hmmm... I think the way it is written is correct for that specific case, and the examples help explain it.

We could be more generic (no pun intended) in our definition to talk about preserving the ordering of types in the case of covariance and reverse for contravariance. 

But we need to define it in some fashion -- doesn't have to be formal.

Your statement  

> for Foo<T> to be covariant on T the type Foo<Tu> must be a subtype of Foo<Tv> for any type Tu that is a subtype of Tv

Could possibly be the words we use instead.

@oyamauchi - any thoughts here?

Issue #73, opened 2015-11-14
Add examples for all the methods in the Collections API


JoelMarcey:

daviddoran:
@JoelMarcey did you have a plan of attack in mind? Other than auto-generating the skeletons and filling out each method example manually?

JoelMarcey:
@daviddoran So, I figured I would start with an overall Collections example that would live in the top level class information (e.g., http://beta.docs.hhvm.com/hack/reference/class/Vector/). Then, I would basically just surgically go through the API and add examples for each method -- starting with the ones that I think are probably more heavily used.

My theory is that once you have one class done, like `Vector`, you can utilize the same examples with some modifications for the other collection classes.

JoelMarcey:
Although, I need to verify we have the machinery for the overall Collection example that could land somewhere like http://beta.docs.hhvm.com/hack/reference/class/Vector/.

JoelMarcey:
@daviddoran Did you want to take a stab at more of the collections examples? I imagine if we can get through `Vector`, `Map` especially should fall into place, and probably the other two as well.

I was planning on starting on #74 anyway. 

No worries if not, I will get to them at some point :)

daviddoran:
@JoelMarcey Yeah I can try and get `Vector` done in one block. I realised after doing a few that going one-by-one (with screenshots in the PR, etc) is too slow. So if I find time I'll sit down and try and knock out a bunch of examples in one PR this week.

**PS: If someone else decides to do this instead just comment here so we're not doubling our effort.**

JoelMarcey:
That's great :+1: 

I won't work on collections -- I will work on some other issues. Just let me know if you need some help or need me to take over due to time, etc.

JoelMarcey:
\o/

:+1: 

Thanks @daviddoran 

This issue is closable I think. You?

daviddoran:
@JoelMarcey Yup, except for the `Imm*` variants. Should I run a script to generate markdown files for the immutable collections, e.g. "See [`Set::blahMethod`](#examples) for example usage"?

JoelMarcey:
@daviddoran Yeah, that would be great. I don't think we need to duplicate examples for the `ImmXXX` classes (at least initially). If there is pushback when we launch, we can just copy the current `XXX` examples for methods that exist on `ImmXXX` and pretty much change the code to `ImmXXX`.

sgolemon:
[...]

PR #126, opened 2015-11-21
[RFC] Add script to autogen API examples for a particular class

daviddoran:
To help mass-create API reference examples (e.g. for the `Vector` or `Map` classes) I created a basic script to auto-generate example files using `APIIndex`. I deliberately added `TODO` text to the `expect` files so that the tests will fail (giving an easy way to know how many examples are left to write).

The script is designed to be idempotent so you can safely run it on a class with a new method, for example, and only the example files for that method will be generated.

To generate examples for all of `Map`'s methods:

    hhvm bin/class-reference-skeletons.php Map

Here's the generated PHP for `Map::toImmSet`:

~~~php
<?hh

namespace Hack\UserDocumentation\API\Examples\Map\ToImmSet;

// TODO: Add example code
~~~

And the generated markdown:

    TODO: Add preamble content

And the `.expect` file:

    TODO: Replace with expected output

And a screenshot of the rendered reference page:

map toimmset 2015-11-21 00-28-28

### DIFF
--- a/bin/class-reference-skeletons.php
+++ b/bin/class-reference-skeletons.php
@@ -1,0 +1,128 @@
+<?hh // strict
+
+/**
+ * Usage: hhvm class-reference-skeletons.php SomeClassName
+ *
+ * Use this script to create skeleton API reference examples
+ * for every method in class SomeClassName.
+ */
+
+namespace HHVM\UserDocumentation;
+
+require __DIR__.'/../vendor/autoload.php';
+
+const string EXAMPLES_DIRECTORY = __DIR__.'/../guides/hack/99-api-examples';
+
+/* HH_FIXME[1002] This is a script that has toplevel statements */
+classReferenceSkeletons($argc, $argv);
+
+function classReferenceSkeletons(int $argc, array<int, string> $argv): void {
+  if ($argc < 2) {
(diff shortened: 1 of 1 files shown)


fredemmott:
The script looks nice, as long as the output is /never/ committed without modification:

 - I'd rather users see no examples than boilerplate
 - It makes it slightly more annoying to find non-existent examples
 - I don't want expected-failure test runs

bin/class-reference-skeletons.php
```diff
@@ -0,0 +1,128 @@
...
+        '',
+        sprintf(
+          'namespace Hack\UserDocumentation\API\Examples\%s\%s;',
+          $this->className,
+          ucfirst($this->methodName),
+        ),
+        '',
+        '// TODO: Add example code',
+        '',
+      },
```
fredemmott: could a plain old heredoc be nicer?

daviddoran:
@fredemmott Yeah I actually wrote this for #73 where I'm planning to write examples for all the methods in a class, then commit. So the purpose of the boilerplate and failing tests is to help writing actually content (not to commit boilerplate or broken tests).

I'm happy to not commit this though and just use it locally.

fredemmott:
nah :)

PR #135, opened 2015-11-24
Remove duplicate `like` from sentence.

sgolemon:
It's like, so like, valley or whatever... ohmigawd...


daviddoran:
@sgolemon should the sentence be "... should **look** like this ..."?

JoelMarcey:
yep, it should -- and @sgolemon's awesome summary comment is still valid too :)

sgolemon:
derp. Got distracted by TV... lemme fix that :)

PR #138, opened 2015-11-30
Remove redundant backquote and fix link

manzyuk:


manzyuk:
A question about the spaceship operator:

> `$A <=> $B // return 0 if equal, 1 if $A is greater, -1 if $B is greater`

This is different from other languages that have this operator (e.g., Ruby) or a `compare` function/method (e.g., OCaml, Java).  I would expect `$A <=> $B` to return -1, 0, 1 if the `$A` is respectively less than, equal to, greater than `$B`. Is Hack going to be different?

fredemmott:
This looks like it's very confusing wording, but strictly accurate, equivalent to your wording, and matches ruby:

If A is 1 and B is 2:

 - B is greater than A, so by the wording here (taken from the PHP7) RFC, the result is -1
 - A is less than B, so by your wording, the result is also -1

```
irb(main):003:0> A = 1
=> 1
irb(main):004:0> B = 2
=> 2
irb(main):005:0> A <=> B
=> -1
```

Full PHP7 behavior (which we should match):

https://3v4l.org/sHe1j

manzyuk:
Whoa, that's embarrassing. Of course they are equivalent. Sorry for the noise.

PR #147, opened 2015-12-01
Fix styling of code blocks in lists

jwatzman:
The old rule here is pretty specific, `p > code` to get the green
background. At first glance, it seems overly specific, but we don't want
to give code inside links green backgrounds. So let's just add a second
rule that `li > code` also gets the green background, which fixes all
the cases I can see.

### DIFF
--- a/sass/innerContent.scss
+++ b/sass/innerContent.scss
@@ -31,6 +31,10 @@
     max-width: 100%;
   }
 
+  li > code {
+    @include codeemph;
+  }
+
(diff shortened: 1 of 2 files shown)


jwatzman:
Sending as a PR instead of directly committing so that @matthewjohnston4 can take a look.

matthewjohnston4:
Meant to do this myself, so thanks for doing it :) And :+1: for using mixins :D

PR #185, opened 2015-12-09
Fix typo in XHP Bootstrap url

jjeising:
This was `boostrap` instead of `bootstrap`, working link is http://github.com/hhvm/xhp-bootstrap


JoelMarcey:
Thanks!

PR #186, opened 2015-12-09
Fix XHP introduction link

jjeising:
The XHP link previously linked to `/hack/unsupported/xhp/introduction` but it should link to `/hack/XHP/introduction`.


JoelMarcey:
Thanks!

PR #207, opened 2015-12-12
Also redirect to php manual for appendix type

SiebelsTim:
Previously we would not redirect on http://docs.hhvm.com/manual/en/stream.constants.php

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -152,8 +152,8 @@
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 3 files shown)


SiebelsTim:
As a PR rather than directly pushing because I guess you would have thought about adding appendix to the list. 
Does this do anything bad, or did you just miss it?

http://docs.hhvm.com/manual/en/stream.constants.php was reported 404 in IRC.

fredemmott:
Add test please :)

PR #228, opened 2015-12-18
Use HH\autoload_set_paths for autoloading

fredemmott:
Uses fredemmott/hhvm-autoload composer plugin

Given this is my plugin, cc @SiebelsTim and @JoelMarcey instead of just merging this directly :p

Planning to split up typedefs.php after this.


JoelMarcey:
shipit

Issue #277, opened 2016-02-07
description for Indexish in Collections.Interfaces section is incorrect


savil:
Currently says:
"Extends KeyedTraversable<Tk, Tv> and is anything that can be an index in an array or collection, which in this case is string and int. It is somewhat common to have a method take an Indexish or return an Indexish."

The "and is anything that can be an index in an array or collection" is describing an `arraykey` not an Indexish

PR #278, opened 2016-02-07
fix description of Indexish

savil:
The previous description was more suitable for explaining what `arraykey` is.
Addresses issue #277, which i just filed.


fredemmott:
This seems incorrect, or at least unclear, given:

```
[fredemmott@devbig349.prn1 ~/test] cat test.php
<?php

class Foo implements ArrayAccess {
  public function offsetGet($offset){}
  public function offsetSet($offset, $value){}
  public function offsetExists($offset){}
  public function offsetUnset($offset){}
}

var_dump((new Foo()) instanceof Indexish);
var_dump((new Foo()) instanceof HH\Indexish);
[fredemmott@devbig349.prn1 ~/test] hhvm test.php
bool(false)
bool(false)
```

SiebelsTim:
@fredemmott Well, in Hack you can't really use ArrayAcces. Classes implementing ArrayAccess cannot be accessed with square brackets.

fredemmott:
PHP interoperability is still a concern - this would imply that a Hack function taking indexish could be called with an arrayaccess implementation from PHP code.

savil:
is the concern with the "square-bracket notation" part? Would amending that to "is anything that can be indexed into, like an array or collection" suffice?

fredemmott:
Still inaccurate for arrayaccess.

savil:
hmm okay, yeah, i haven't fully grokked the implications of needing to not imply this is ArrayAccess. I'm a tad lost on how to proceed. Do you agree the current wording is incorrect? Suggestions on how to fix it?

fredemmott:
Yep, the current wording is currently incorrect. Right now we're leading towards what you have + "except for objects that implements ArrayAccess", though I'm trying to find out if there was a more consistent answer here.

The whole collections hierarchy is somewhat overcomplicated for what it does...

fredemmott:
@savil : is there a plan here?

PR #285, opened 2016-03-01
Pipe operator

alexchow:


guides/hack/45-operators/05-pipe-operator.md
```diff
@@ -0,0 +1,62 @@
...
+
+which conceptually means:
+
+```
+h(5, g(f()))
+```
+
+That is, evaluate the left hand side (LHS) of the pipe, then evaluate the RHS of the pipe with the result of the LHS bound to the placeholder `$$`.
+
+Runtime execution is explicitly left-to-right - a pipe's LHS is always executed before its RHS. More precisely, the LHS of a Pipe is evaluated and its result stored in a temporary, hidden variable. Then evaluate the RHS where `$$` refers to that hidden variable. This means you can use $$ more than once in the RHS but the LHS will still be evaluated only once. For example:
```
mpark: 1. `s/Then evaluate the RHS/Then the RHS is evaulated/`
1. backticks around $$.

guides/hack/45-operators/05-pipe-operator.md
```diff
@@ -0,0 +1,62 @@
...
+
+```
+f() |> g1($$) + g2($$) |> h(5, $$)
+```
+
+roughly means:
+
+```
+h(g1(f()) + g2(f()))
```
mpark: `h(5, g1(f()) + g2(f()))`

guides/hack/45-operators/05-pipe-operator.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+except that `f()`:
+- is evaluated only once
+- is evaluated before either `g1` or `g2`.
```
mpark: before either `g1` or `g2` => before both `g1` and `g2`

fredemmott:
As a more general note, please stick to 80 cols; a lot of the documentation doesn't do this because it used to create wordwrap issues with our rendering pipeline, but not any more.

guides/hack/45-operators/01-introduction.md
```diff
@@ -15,3 +15,7 @@ Allows for method and property calls without explicit checks to ensure that the
 ## `->:` ([XHP attribute access operator](XHP-attribute-access.md))
 
 Allows for accessing XHP attributes without explicitly having to call `getAttribute()` on the XHP object.
+
+## `==>` ([Pipe operator](pipe-operator.md))
```
SiebelsTim: That's the lambda operator.

guides/hack/45-operators/05-pipe-operator.md
SiebelsTim: If I understood correctly, the left hand side will only be evaluated once for multiple usages. This implies that it will be evaluated twice.

guides/hack/45-operators/05-pipe-operator.md
```diff
@@ -0,0 +1,62 @@
...
+
+It may be easier to conceptualize this as a lambda rewrite:
```
SiebelsTim: I think this is more confusing than helping.

guides/hack/45-operators/05-pipe-operator.md
```diff
@@ -0,0 +1,62 @@
...
+
+(Note: there is no lambda construction overhead in the runtime because of the hidden-variable execution strategy described above)
+
+Hack will enforce that you do use the `$$` of each pipe at least once (if you didn't, it's probably a bug or you don't need the pipe anyway).
+
+## Example
+
+Classic style:
+```
+$filtered = $collection->filter($x ==> $x.getExperiment() == nullthrows(idx(Experiments::getAll(), ucfirst(strtolower($input))))) === 0)
```
SiebelsTim: These might be better inside a file to get it tested.

guides/hack/45-operators/05-pipe-operator.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Pipe style:
+```
+$filtered = $input
```
SiebelsTim: same here. Although we require a newer HHVM version for the docs then..

guides/hack/45-operators/01-introduction.md
JoelMarcey: `|>`

guides/hack/45-operators/05-pipe-operator.md
alexchow: Hmmm... I thought a reification of the runtime's execution strategy would make things clearer than the english paragraph above. Is it not?

There's an alternative reification using name replacement:
```
(
  () ==> {
    $tmp1 = f();
    $tmp2 = g1($tmp1) + g2($tmp2);
    return h(5, $tmp2);
  }
)()
```
(The reason for the lambda wrapper is because the pipe expression is an expression not a statement, so this generalizes.)

Is that clearer?

SiebelsTim: I think that's clearer. I don't think it's helping though :P 
This is more of the implementation the user needs to know. That lambdas don't have anything do to with the pipe operator makes this confusing to me.

All I needed to immediately understand how to use the operator and why it exists was the filter/map example from the proposal in https://github.com/facebook/hhvm/issues/6455
Especially the why one wants to use it is important for me. That's why I prefer examples showing real usage over function names with no meaning.

```
// map -> filter -> count
array(...)
  |> array_map($x ==> $x->getNumber(), $$)
  |> array_filter($$, $x ==> is_even($x))
  |> count($$)
```

The user doesn't really need to know about temporaries, scoping, and expression vs statement in my opinion. 

That's just my opinion though. I could be wrong ;-)

SiebelsTim:
> I am super glad that docs are being written by the actual feature writers. I hope this starts a trend. :)

I really like this, too!

Also I'd like to point out that examples showing real world usage (array functions, map, filter, something, ...) are more valueable than using made up functions that have no meaning. :)

guides/hack/45-operators/05-pipe-operator.md
alexchow: I can remove the reification.

A map->filter->count example is a good idea and I can add that. I want an additional example because not everyone is convinced using pipes there is actually easier to read than the standard nested calls, and also pipes are useful beyond simply "the collections library is insufficient".

guides/hack/45-operators/05-pipe-operator.md
alexchow: Let's make files later then.

fredemmott:
We can have the files now, but have the expected output be syntax errors - then we update them when we update to HHVM 3.13

PR #294, opened 2016-03-16
Noticed a typo while stealing examples for PHP RFC

philsturgeon:
It's not very often I get to fix somebody else's typos! :sunglasses:


SiebelsTim:
Looks like it's a word that is hard to type :D

philsturgeon:
Oh I get it wrong all the time. Thanks!

PR #309, opened 2016-04-21
fixed index values in comments

HakuDouga:
Array indexing starts at zero. Minor mistakes there. Orvid King guessed a better labeling for that one so I've changed ´index´ to ´element´.

Bye~


Orvid:
It may be better to label these as elements rather than index.

PR #358, opened 2016-08-15
Implement display __Deprecated attributes

### LINKED ISSUE #129: Handle __Deprecated annotations

SiebelsTim:
fixes #129

### DIFF
--- a/src/api-gen/FunctionMarkdownBuilder.php
+++ b/src/api-gen/FunctionMarkdownBuilder.php
@@ -40,6 +40,7 @@ final class FunctionMarkdownBuilder {
       "\n\n",
       [
         $this->getHeading(),
+        $this->getDeprecation(),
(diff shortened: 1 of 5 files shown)


fredemmott:
Can you add something to the tests for this?

SiebelsTim:
@fredemmott Do we have any tests for similar things? Such as if a method is static.

fredemmott:
No, just that some example pages load; would still be good to add to APIPagesTest though.

SiebelsTim:
Is this enough?

simonwelsh:
This doesn't handle a sample rate set as part of the annotation.

SiebelsTim:
I completely forgot about this one. This makes things easier as it is invalid to pass more than 2 and less than 1 parameter. So we only need the first parameter for the message.

PR #359, opened 2016-08-16
Simplify __Deprecated handling

SiebelsTim:
and also fix when using $sample_rate

This shouldn't change the output we generate as hhvm's codebase does
not use $sample_rate. However it simplifies the code :)

refs #358


SiebelsTim:
As always Simon found a mistake in my code :P 

This shouldn't affect the generated output, so (assuming this gets merged) this doesn't need a push.

Issue #116, opened 2015-11-19
Hack API reference docs don't indicate which parameters and return types are nullable


daviddoran:
For example, these methods all have a nullable return type:

- http://staging.docs.hhvm.com/hack/reference/class/Vector/firstValue/
- http://staging.docs.hhvm.com/hack/reference/class/Vector/get/

This method takes a single nullable argument:

- http://staging.docs.hhvm.com/hack/reference/class/Vector/setAll/

But they're displaying as non-nullable.

JoelMarcey:
Good find. :+1: 

Not sure how far down the tree this goes. We get the information from hhi/hni files. Then create a YAML tree. Then render. Etc.

cc @fredemmott

sorawee:
Can we reopen this issue? Although http://staging.docs.hhvm.com/hack/reference/class/Vector/firstValue/ shows 

> Tv - The first value in the current Vector, or null if the Vector is empty.

But the signature of the function is still 

```
public function firstValue(): Tv
```

This should actually be

```
public function firstValue(): ?Tv
```

JoelMarcey:
It is on the actual live site too:

https://docs.hhvm.com/hack/reference/class/HH.Vector/firstValue/

I noticed in the `.hhi` file, that it does show `?Tv` .. https://github.com/facebook/hhvm/blob/3d171f73f2becfe4976a2920581f53b7dd0350e6/hphp/hack/hhi/collections/Vector.hhi#L362

@fredemmott --- This worked before though, I think, right?

fredemmott:
We still have the param one (https://staging.docs.hhvm.com/hack/reference/class/HH.Vector/setAll/) btu not the return types.

As usual for regressions, please open a new issue.

PR #360, opened 2016-08-24
Fix merging nullable typehint and mixed

SiebelsTim:


SiebelsTim:
refs #116

---

This happens when merging the YAML files. 

Example for `Vector::firstValue()`:

HHI defines `firstValue(): ?Tv`.
Systemlib defines `firstValue()` making it `firstValue(): mixed`.

Rules for merging this are:
- Use HHI if it isn't `mixed` or `object`. If it is `mixed` or `object` use Systemlib.
- Nullable is always `hhi[nullable] && systemlib[nullable]`.
- `mixed` gets `false` for `nullable`.

This results in taking `Tv` from HHI, and taking `nullable` from Systemlib, which is false.

---

I assume changing `nullable` from `hhi[nullable] && systemlib[nullable]` to `hhi[nullable] || systemlib[nullable]` is wrong when systemlib needs to define nullable as true but hhi as false because in practice it doesn't happen or isn't sensible for an hhi.

I could make `mixed` be `nullable`, but I guess that will mess up any correct mixed typehints ending up in html.

The proposed solution sets nullable to true if one of the typehints is mixed and the other nullable (as mixed is never nullable).

fredemmott:
lgtm; as it sounds like there's been a regression here, can you add to the tests too?

fredemmott:
thanks :)

PR #376, opened 2016-10-02
Split APIGenericPageController into API(Class)?PageController

fredemmott:
Truly generic stuff in the superclass, class-specific stuff goes in new APIClassPageController.
This is both neater, and gets us to a point where every WebController is either abstract or final, which is both a good thing in itself, and unblocks some other stuff I want to do.

Added dependcy on type assert as the previous implementation of
`ArgAssert::isClassname()` only worked for direct subclasses, which is no
longer the case here. Will put up a separate diff removing ArgAssert now
that it's not good enough anyway.


JoelMarcey:
This is definitely reasonable to me. Going to let the HHVM team officially review and accept though.

PR #377, opened 2016-10-02
Remove ArgAssert, just use TypeAssert

fredemmott:
Dependency added in #376, might as well just use it directly.

Depends on/includes #376


JoelMarcey:
Looks good. HHVM team should accept #376 first.

PR #378, opened 2016-10-02
Split api list page

fredemmott:
- cleaner
 - one URI pattern per controller (which unblocks futue work to use a common routing library)
 - unblocks future work to turn this into composition instead of inheritance

Depends on (and includes) #376


JoelMarcey:
looks legit. HHVM team, this needs #376 to be merged in first.

PR #379, opened 2016-10-03
Generate uri map from URI patterns defined in the controllers

fredemmott:
- fredemmott/hack-router was a fork of the router in the this site, meant to make it
   re-usable
 - fredemmott/hack-router-codegen builds on fredemmott/definition-finder (already
   needed) and facebook/hack-codegen to build routers/urimaps, uri builders, and
   type-safe request parameter objects.

As well as generating the URI map, it replaces the main routing implementation with a generic type-safe one - no UNSAFE or HH_FIXME any more :)

This diff just makes the site use a generated router; will do uri builders and
type-safe parameter objects in a future pull request if this lands (as it depends
on this).

Depends on (and includes) #376, #377, #388


JoelMarcey:
👍

PR #389, opened 2016-11-15
Add mirrors.tuna.tsinghua.edu.cn

ym:
Add another HHVM mirror in CN
https://github.com/tuna/issues/issues/136


fredemmott:
Sorry, we've not been able to prioritize looking at the documentation until recently :(

Closing as:

this mirror is now stale (see instructions for updating on hhvm/user-documentation@ca087df )
dl.hhvm.com is now fronted by a CDN (CloudFront) so this should be less useful.

Issue #375, opened 2016-10-02
Shapes::idx example indicates it can be used for fields that aren't in the shape type definition


fredemmott:
Please complete the information below:

# Where is the problem?

Bottom of the first example

# What is the problem?

`Shapes::idx` is for use with optional/nullable fields, not ones that aren't defined.

For example:

```Hack
root@d32efb1aa9e9:~/hackdev/test# cat test.php
<?hh

function needs_nullable_string(?string $foo): void {
}

type TFoo = shape('somestr' => ?string, 'someint' => ?int);

function main(TFoo $x) {
  needs_nullable_string(Shapes::idx($x, 'somestr')); // ok
  some_undefined_func(Shapes::idx($x, 'nopenopenope')); // type error
}

main(shape());
root@d32efb1aa9e9:~/hackdev/test# hh_client
test.php:10:39,52: Invalid argument (Typing[4140])
  test.php:8:15,18: The field 'nopenopenope' is missing
```

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: 2016-08-15T18:49:04+0000:e756f9162e96b867a06b2805e03056a63fc17931:2b940b7ef772640b5980b78c645f89cbd216d81f
 - Page requested: /hack/reference/class/HH.Shapes/idx/
 - Page requested at: Sun, 02 Oct 2016 04:02:07 +0000
 - Controller: APIMethodPageController

Issue #374, opened 2016-10-02
Shapes::idx examples are all not typechecked, some invalid


fredemmott:
Please complete the information below:

# Where is the problem?

The entire example.

# What is the problem?
The typechecker doesn't look at pseudomains, and all the code is in pseudomains; a lot of these examples are not legal Hack code.

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: 2016-08-15T18:49:04+0000:e756f9162e96b867a06b2805e03056a63fc17931:2b940b7ef772640b5980b78c645f89cbd216d81f
 - Page requested: /hack/reference/class/HH.Shapes/idx/
 - Page requested at: Sun, 02 Oct 2016 03:47:00 +0000
 - Controller: APIMethodPageController

fredemmott:
They're now in a pseudomain, but this is still invalid and should raise a type error:

```
+  // The key 'z' doesn't exist in $point so our explicit default 0 is returned
 -var_dump(Shapes::idx($point, 'z', 0));		 +  var_dump(Shapes::idx($point, 'z', 0));
```

fredemmott:
Though I guess that's partially a dupe of #375, the change commit you made for this issue should cause test failures

PR #395, opened 2016-12-04
Explain that nullable fields in shapes are also optional

barillari:
Warn shape users that the typechecker conflates optional and nullable.


daviddoran:
You could mention that [`Shapes::idx`](https://docs.hhvm.com/hack/reference/class/HH.Shapes/idx/) can be used to access a possibly missing or null field safely?

fredemmott:
See also #374 #375

`??` is usable instead of `Shapes::idx()` and largely preferred externally (as it's not hack-specific and is just as safe)

fredemmott:
Merging in as-is - agree with @daviddoran, but that can be a separate PR

PR #439, opened 2017-05-31
Two commits: update package dependencies, and fix toVector() examples

Meowse:


Meowse:
I would definitely prefer a rebase-and-merge, as the two commits are entirely distinct (non-overlapping files, and logically unrelated).

PR #479, opened 2017-12-05
Removed mirror.yourwebhoster.eu

MrKoopie:
We love the HHVM project, however since we do not have an official rsync source we can not longer run a HHVM mirror. If you (HHVM) can provide a rsync server, we will enable mirroring again.


fredemmott:
Thanks; this is also probably slightly less useful now anyway given that dl.hhvm.com is backed by cloudfront now.

PR #507, opened 2018-02-19
Clearly illustrate how *type-definitions* work

admdikramr:
The *type-definition* option for enums should be listed under examples for casting, because they directly relate to the utility, necessity, and risks involved thereof. I add an extra example to illustrate all three of these.


admdikramr:
Ugh, nevermind. I forgot how difficult github is to work with.

PR #508, opened 2018-02-19
Include an illustration of the danger of implicit casting for enums

admdikramr:
The current docs do not illustrate the trouble in allowing implicit casting, which is that a string or int with semantic meaning is almost never correct input to a function that wants "normal" meaning. Normal meaning being, counts or quantities for ints, and messages or blobs for strings.

### DIFF
--- a/guides/hack/28-enums/01-introduction-examples/implicit-casting.php
+++ b/guides/hack/28-enums/01-introduction-examples/implicit-casting.php
@@ -21,13 +21,24 @@ function give_shirt(Size $size): void {
 
 function sales(): void {
   // Implict casting at work
-  say_greeting_with_size(Size::SMALL);
+  $s = Size::SMALL;
+  say_greeting_with_size($s);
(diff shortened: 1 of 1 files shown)


fredemmott:
Examples are unit tests; this can't be merged as:
- separate to this PR, we need to fix a bug in HHVM to be able to test the typechecker
- the HSL is not currently available to these; if it were, a `use namespace HH\Lib\Str` would be needed
- you are referencing a completely undefined function

AndrewDiMola:
@admdikramr do you still want to work on this PR to add more clarity to the dangers of implicit casting?

AndrewDiMola:
Closing as it's been almost a year since I reached out and the PR itself is 4 years old, with many changes to the documentation set since then. @admdikramr - Feel free to create a new issue or PR against this page https://docs.hhvm.com/hack/built-in-types/enum if you still have feedback, and I can help you sort it out.

PR #523, opened 2018-04-21
Set a default namespace to improve auto-linkify

### LINKED ISSUE #522: AutoLinkify linking the Wrong `join` Method in Awaitables guide
It's linking to `HH\Lib\Str\join` instead of `HH\Asio\join` in:
https://docs.hhvm.com/hack/async/awaitables

Do we need to fix the priority namespaces in:

https://github.com/hhvm/user-documentation/blob/master/src/markdown-extensions/AutoLinkifyInline.php#L71-L78

kevinastone:
Add support for a metadata field to specify the default namespace for a block to keep short code references but allow the auto-linker to resolve the correct reference.

Fixes #522

### DIFF
--- a/src/typedefs.php
+++ b/src/typedefs.php
@@ -37,6 +37,7 @@ type YAMLMeta = shape(
   ?'fbonly messages' => vec<string>,
   ?'min-versions' => dict<RequirableProduct, string>,
   ?'experimental' => bool,
+  ?'namespace' => string,
 );
 
 
--- a/guides/hack/22-async/02-awaitables.md
+++ b/guides/hack/22-async/02-awaitables.md
@@ -1,3 +1,9 @@
+```yamlmeta
+{
+  "namespace": "HH\\Asio"
+}
+```
+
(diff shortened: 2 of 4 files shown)


src/typedefs.php
```diff
@@ -37,6 +37,7 @@
   ?'fbonly messages' => vec<string>,
   ?'min-versions' => dict<RequirableProduct, string>,
   ?'experimental' => bool,
+  ?'default_namespace' => string,
```
fredemmott: thanks just a nit: please replace underscore with space for consistency with `fbonly messages`; I see that I messed up and `min-versions` is already inconsistent, but let's not add a third way

fredemmott: inherited from when this actually was yaml :p I'd also be happy for all of these to be changed to something else, I don't have any strong preference as long as they're consistent :)

kevinastone: Agreed.  I just renamed it to `namespace` to avoid this :)

PR #540, opened 2018-07-31
Update emacs plugin documentation to point to new hack-mode

jra3:
Summary: The new plugin now lives in its own repo. The documentation
for emacs was copied verbatim from the Vim section.

Reviewers: joelm, fredemmott


Wilfred:
We should consider dropping `hack-mode.el` from the HHVM repo to avoid confusion.

jra3:
@Wilfred indeed https://github.com/facebook/hhvm/pull/8274

I'll be honest though, I'm not sure if i made that diff and pull request correctly

PR #556, opened 2018-10-10
Note that thrown exceptions are not cached

hurrymaplelad:


Wilfred:
Agreed, that's worth documenting. Thanks! 👍

PR #559, opened 2018-11-07
Don't bother showing "This API is part of HSL"

Wilfred:
The breadcrumbs already show where the function is. Removing this text
ensures that the first sentence on the page is more useful to the
user, the overview of the function/class.

cc @fredemmott, let me know if you have any objections.


fredemmott:
I think this is still valuable, but the wording could do with some work: I think it's important to emphasize that they're not available if you've just installed HHVM

Wilfred:
hh_filter_docs

We mention HSL in the breadcrumbs, and the name starting `HH\Lib` also shows that it's in the HSL. Do we have anyone using HHVM without HSL?

What do you think of using a badge instead? Something like this:

hack_badge

Wilfred:
Updated: Improving the wording rather than removing.

PR #564, opened 2018-12-07
Fix typo in class property initialization doc

watashi:


Wilfred:
Well spotted, thanks :)

PR #565, opened 2018-12-18
Remove references to PHPUnit

Wilfred:
We are no longer using this test runner as far as I can see. `vendor/bin/phpunit` does not seem to exist any more.

Note that `bin/test-staging.sh` still has a reference to phpunit, so is that script even used still?

cc @fredemmott


fredemmott:
It's not currently used due to https://github.com/hhvm/hacktest/issues/60

PR #580, opened 2019-02-22
Flesh out tools section, and add 'real projects' getting started page

fredemmott:
Real projects covers installing dependencies, CLI entrypoints, autoloader, linters, and unit tests.

### DIFF
--- a/guides/hack/01-getting-started/02-tools.md
+++ b/guides/hack/01-getting-started/02-tools.md
@@ -1,3 +1,47 @@
-**Under Construction**
+## Tools
 
-A placeholder for tools info.
+The core tools are:
+
+- `hh_client`: this is the command line interface for Hack's static analysis; it
+  is needed to verify that a project is valid Hack, and is used to find errors
+  in your programs
+- `hhvm`: this is used to execute your Hack code, and can either be used for
+  CLI (e.g. `hhvm foo.hack`) or as a server, and has
+  [extensive documentation](/hhvm/)
+
+### Editors and IDEs
+
+You can use any plain-text editor to edit Hack files - however, enhanced
+integration with Hack is available for several editors.
+
+We primarily recommend using [Visual Studio Code] with the [VSCode-Hack]
+extension; this provides standard IDE-like features such as
+syntax highlighting, go-to-definition, and in-line display of Hack errors.
+
+For Vim users, [vim-hack] provides syntax highlighting and language detection,
+and the [ALE] project provides enhanced support for Hack.
+
+[hack-mode] provides a major mode for Emacs users.
+
+### Dependency Management
+
+Hack dependencies are currently managed using [Composer], which must be executed
+with PHP.  Composer can be thought of as an equivalent to `npm` or `yarn`.
+
+### Other Common Tools
+
+- `hackfmt` is a CLI code formatter included with HHVM and Hack, and is also
+  used by the various editor and IDE integrations
+- [HHAST] is added provides code style linting, and the ability to automatically
+  modify code to adapt to some changes in the language or libraries
+- [hacktest] and [fbexpect] are commonly used together for writing unit tests
+
+[ALE]: https://github.com/w0rp/ale
+[Composer]: https://getcomposer.org
+[HHAST]: https://github.com/hhvm/hhast
+[VSCode-Hack]: https://github.com/slackhq/vscode-hack/
+[Visual Studio Code]: https://code.visualstudio.com
+[fbexpect]: https://github.com/hhvm/fbexpect
+[hack-mode]: https://github.com/hhvm/hack-mode
+[hacktest]: https://github.com/hhvm/hacktest
+[vim-hack]: https://github.com/hhvm/vim-hack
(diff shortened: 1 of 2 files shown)


guides/hack/01-getting-started/02-tools.md
```diff
@@ -1,3 +1,47 @@
...
+- [hacktest] and [fbexpect] are commonly used together for writing unit tests
+
+[ALE]: https://github.com/w0rp/ale
+[Composer]: https://getcomposer.org
+[HHAST]: https://github.com/hhvm/hhast
+[VSCode-Hack]: https://github.com/slackhq/vscode-hack/
+[Visual Studio Code]: https://code.visualstudio.com
+[fbexpect]: https://github.com/hhvm/fbexpect
+[hack-mode]: https://github.com/hhvm/hack-mode
+[hack-test]: https://github.com/hhvm/hacktest
```
Wilfred: I suggest you use inline links rather than putting them at the end. It makes longer files easy to edit without scrolling between the link and its usage.

In any case, I think you want `hacktest` here :P

guides/hack/01-getting-started/02-tools.md
```diff
@@ -1,3 +1,47 @@
...
+  is needed to verify that a project is valid Hack, and is used to find errors
+  in your programs
+- `hhvm`: this is used to execute your Hack code, and can either be used for
+  CLI (e.g. `hhvm foo.hack`) or as a server, and has
+  [extensive documentation](/hhvm/)
+
+### Editors and IDEs
+
+While you can edit Hack with any text editor, enhanced support is available
+for several.
```
Wilfred: This sentence reads oddly to me. How about something like this?

> Whist you can edit Hack with any text editor, several editors are available with dedicated Hack functionality.

(Aside: Have we stopped mentioning Nuclide already?)

guides/hack/01-getting-started/03-starting-a-real-project.md
```diff
@@ -0,0 +1,332 @@
...
+
+Real projects generally aren't a single file in isolation; they tend to have
+dependencies such as the [Hack Standard Library], and various optional tools.
+
+A good starting point is to:
+- [install Composer]
+- create an empty `.hhconfig` file
+- create `src/` and `tests/` subdirectories
+- configure autoloading
+- use Composer to install the common dependencies and tools:
```
Wilfred: Is this bullet point finished? I don't understand why it ends with a colon.

guides/hack/01-getting-started/03-starting-a-real-project.md
```diff
@@ -0,0 +1,332 @@
...
+
+As a toy example, we're going to create a function that squares a vector of
+numbers; save the following as `src/square_vec.hack`:
+
+```Hack
+use namespace HH\Lib\Vec;
+
+function square_vec<T as num>(vec<T> $numbers): vec<int> {
+  return Vec\map($numbers, $number ==> $number * $number);
+}
```
Wilfred: Hack with generics and type constraints seems a little intense for newbies. Could we use concrete types?

```php
function square_vec(vec<float> $numbers): vec<int> {
  return Vec\map($numbers, $number ==> $number * $number);
}
```

guides/hack/01-getting-started/03-starting-a-real-project.md
```diff
@@ -0,0 +1,332 @@
...
+If you then run `hh_client`, it will tell you of a mistake:
+
+```
+src/square_vec.hack:4:10,57: Invalid return type (Typing[4110])
+  src/square_vec.hack:3:53,55: This is an int
+  src/square_vec.hack:4:40,56: It is incompatible with a num (int/float) because this is the result of an arithmetic operation with a num as the first argument, and no floats.
+  src/square_vec.hack:3:35,35: Here is why I think the argument is a num: this is a num
+```
+
+To fix this, change the return type of the fucntion from `vec<int>` to `vec<num>`.
```
Wilfred: *function

guides/hack/01-getting-started/03-starting-a-real-project.md
```diff
@@ -0,0 +1,332 @@
...
+```
+
+This program:
+ - requires and initializes the autoloader so that the function we just defined
+   can be found
+ - calls the function
+ - prints the results
+ - exits
+
+ Unlike C, there is nothing special about the name `main`; the
```
Wilfred: Where possible, I think our prose is clearer where we don't assume knowledge of specific other languages.

```suggestion
There is nothing special about the name `main`; the
```

guides/hack/01-getting-started/02-tools.md
fredemmott: it really hurts readability though, especially with generally sticking to 80col. Nice catch though :)

guides/hack/01-getting-started/02-tools.md
fredemmott: Yep, Nuclide is dead as an open source project

fredemmott:
Thanks ; did suggestions (some with slightly different wording) by amending the previous commits.

guides/hack/01-getting-started/02-tools.md
Wilfred: In which case I believe you can put the links immediately after the paragraph, rather than at the end of the file.

PR #582, opened 2019-03-06
[rex-docs] Clean up mentions of static local variables

dneiter:
Static local variables are being removed from Hack and HHVM. Update documentation not to reference them.

I restructured memoize examples to wrap both time_consuming and memoization wrapper into abstract final class, so we get clean and logical syntax for static class property. Other options would be:
- use global variable for memoization (I don't want to promote globals use for this)
- only wrap static property in a class but leave functions free-standing. This would work fine, but code looks less clean.

Test Plan:
Run guides/hack/09-asynchronous-operations/28-guidelines-examples/memoize-awaitable.php and guides/hack/09-asynchronous-operations/28-guidelines-examples/memoize-result.php. Make sure, both run successfully.


guides/hack/09-asynchronous-operations/28-guidelines-examples/memoize-awaitable.php
```diff
@@ -2,26 +2,29 @@
 
 namespace Hack\UserDocumentation\AsyncOps\Guidelines\Examples\MemoizeAwaitable;
 
-async function time_consuming(): Awaitable<string> {
-  \sleep(5);
-  return "Not really time consuming but sleep."; // For type-checking purposes
-}
+abstract final class MemoizeAwaitable {
+  private static async function time_consuming(): Awaitable<string> {
+    \sleep(5);
```
azjezz: `HH\Asio\usleep()` ?

guides/hack/09-asynchronous-operations/28-guidelines-examples/memoize-result.php
```diff
@@ -2,26 +2,28 @@
 
 namespace Hack\UserDocumentation\AsyncOps\Guidelines\Examples\MemoizeResult;
 
-async function time_consuming(): Awaitable<string> {
-  \sleep(5);
-  return "This really is not time consuming, but the sleep fakes it.";
-}
+abstract final class MemoizeResult {
+  private static async function time_consuming(): Awaitable<string> {
+    \sleep(5);
```
azjezz: same here

guides/hack/09-asynchronous-operations/28-guidelines-examples/memoize-awaitable.php
dneiter: Done

fredemmott:
This should also be rebased against master now - rex-docs is merged :)

PR #618, opened 2019-03-23
Use type check vs is_null

tcopeland:


tcopeland:
Hm, I don't understanding why CI is failing, but it's due to this change.  Perhaps `is null` checks aren't valid for these examples?  I'll close this until I understand this better.

fredemmott:
https://github.com/facebook/hhvm/commit/b88f1de255cdbfa39d93d6979786e505654d9399 is not yet in any public release of HHVM

tcopeland:
Ah gotcha, makes sense.  I hadn't thought about the gap there.  Thanks @fredemmott!

PR #627, opened 2019-04-02
Add Reified Generics to docs

oulgen:


guides/hack/08-generic-types-and-functions/20-reified-generics.md
```diff
@@ -0,0 +1,187 @@
...
+
+Generics are currently implemented in HHVM through erasure, in which the runtime drops all information about generics. This means that generics are not available at runtime. Although the typechecker is able to use the generic types for static typechecking, we are unable to enforce generic types at runtime.
+
+The goal of opt-in reified generics to bridge the gap between generics and runtime availability while keeping erasure available to maintain performance when reification is not needed. To mark a generic as reified, simply add the `reify` keyword at the declaration site.
+
+## Parameter and return type verification
+
+Prior to reified generics, if the type hint for a parameter or return value had generics, HHVM would only check whether the input is of the outermost type (whether `$c` is a `C` in the next example). With reified generics, HHVM will also check and enforce the reified type parameters.
+
+```Hack
```
fredemmott: - pseudomain example :'( Please use `<<__EntryPoint>>`
- please split out to .php or .hack files and add .hhvm.expect and .typechecker.expect files (or expectf/expectre files - we use the HHVM test runner)
- instead of commetijng with expected result, please change to output something indicating success (so that the .hhvm.expect file is meaningful)

This should go in a '*examples/' subdirectory, and be included with `@@ foo-examples/foo.hack @@` (or .php) - see the other markdown files here for examples.

This makes sure that our examples actually work, and continue to work on new releases of HHVM and the typechecker

guides/hack/08-generic-types-and-functions/20-reified-generics.md
```diff
@@ -0,0 +1,187 @@
+# Reified Generics
+
+## Introduction
```
fredemmott: Are they enabled by default? If so, since when - if not:
- how are they enabled?
- add a note strongly discouraging use of opt-in features (as it means that any dependents must use the same opt-in)

guides/hack/08-generic-types-and-functions/20-reified-generics.md
```diff
@@ -0,0 +1,187 @@
...
+function f<reify T>(T $x): C<T> {
+  return new C<int>();
+}
+
+f<int>(1); // success
+f<int>("yep"); // parameter type hint violation
+f<string>("yep"); // return type hint violation
+```
+
+## Type testing and assertion with `is` & `as` expressions
```
Wilfred:
```suggestion
## Type testing and assertion with `is` and `as` expressions
```

Wilfred:
As we discussed yesterday, you can build the docs locally by installing hhvm from homebrew. I've built it with:

```
$ hhvm bin/build.php; cd public; hhvm -m server -p 8080 -c ../hhvm.dev.ini
```

and I can confirm the formatting looks good :)

fredemmott:
How close is this to ready to go, other than non-executable examples?

oulgen:
@fredemmott - The content is all correct, all its missing is moving the code parts into different files, adding entrypoints and adding expect files. If you're interested in fixing these up, that'd be awesome. Unfortunately, I'm busy with the internal lockdown.

fredemmott:
When’s the planned launch of this feature?

guides/hack/08-generic-types-and-functions/21-reified-generics-migration.md
```diff
@@ -0,0 +1,66 @@
+# Migration Features for Reified Generics
+
+In order to make migrating to reified generics easier, we have added the following migration features:
+
+* The `<<__Soft>>` annotation on a type parameter implies the intent to switch it to reified state. If generics at call sites are not provided the type checker will raise an error, and the runtime will raise a warning so you can capture these locations through logging. The generic cannot be used in the body of the function.
+* The `<<__Warn>>` annotation on a type parameter implies that if the reified generic is incorrect at parameter or return type hints locations, the runtime will raise a warning instead of type hint violation.
+* The `getType<reify T>()` function, given a reified type, returns the type structure representing the type.
```
Lexidor: Is `getType()` the fully qualified name? Is it an instance method on some class or a function?

guides/hack/08-generic-types-and-functions/20-reified-generics.md
```diff
@@ -0,0 +1,187 @@
...
+
+```Hack
+function f<<<__Enforceable>> reify T>(): void {
+    $x is vec<int>; // Cannot use vec<int>
+    $x is T;
+}
+f<vec<int>>(); // not enforceable
+```
+
+but can be used as inner type for classes
```
Lexidor: Why would `AClass<vec<int>>` be `__Enforceable`, `vec<int>` would not be? Is it because the vec<int> is not being checked, or because the type system ensures that checking it would not be required.

Lexidor: How would this interact with TAny?
```HACK
// I am defined in a partial file. I take any variable and erase its type by not declaring my return type.
function get_tany(mixed $mixed){ return $mixed; }

class Box<reify T>{ public function __construct(public T $v){} }
<<__EntryPoint>>
function main(): noreturn{
  $box = new Box<vec<int>>(vec[get_t_any('not an int')]);
  invariant($box is Box<vec<int>>, "As far as I understand this will be true, even though the vec contains a string.");
  exit(0);
}

File: guides/hack/08-generic-types-and-functions/20-reified-generics.md [hunk 6]
```diff
@@ -0,0 +1,187 @@
...
+function f<<<__Newable>> reify T as A>(): T {
+  return new T();
+}
+
+f<A>(); // not newable since it is abstract class
+f<B>(); // success
+f<C>(); // success
+```
+
+Notice that the reified type parameter has the attribute `<<__Newable>>`. In order for a type to be `<<__Newable>>`, the type must represent a class that's not abstract and has a consistent constructor or be a final class. Creating a new instance using the reified generics also carries across the generics given. For example,
```
Lexidor: `__Newable` implies that `new T(/*__ConsistentConstructs params*/)` will be valid.
Does this also open up `is` and `as` checks? Or do you need a `__Newable __Enforceable T` if you with to use `new` and `is` / `as` ? `__Newable` already prevents HHVM from erasing the `T` at runtime.

guides/hack/08-generic-types-and-functions/20-reified-generics.md
```diff
@@ -0,0 +1,187 @@
+# Reified Generics
```
Lexidor: What are the interactions between not reified generics and reified ones? What happens when a lie is passed? F.e. when the type checker thinks something is a `Box of int`, but at runtime this turns out to be false. Are type checks preformed on arguments to functions with reified generic parameters?

```HACK
class Ref<reify T>{ public function __construct(public T $v) {} }
class Box<T>{public function __construct(public T $v) {} }
function get_tany(mixed $mixed){ return $mixed; }

//Would this be allowed? 
$b = new Box(1);
$r = new Ref<int>($b->v);

//Or this?
function lie(Box<int> $boxed_int): void{
  $_ = new Ref<int>($boxed_int->v);
}
lie(new Box(get_tany('not an int')));
```

PR #631, opened 2019-04-18
update build instructions

kunigami:
* hhvm does not work with composer (php) anymore
* it was not super clear step 7 was to be followed by the webserver


README.md
```diff
@@ -21,7 +21,7 @@ Our strategy to create better documentation begins with a re-thinking of our doc
 * Markdown instead of docbook provides an easier path for documentation source readability and updates.
     - Have extensions to support things like example insertion, etc.
 * Token scan our the HHVM code block documentation (instead of reflection) so that rebuilding HHVM isn't necessary to update the documentation.
-  - Use [phpDocumentor](http://www.phpdoc.org/) for docblock parsing 
+  - Use [phpDocumentor](http://www.phpdoc.org/) for docblock parsing
```
fredemmott: This is also no longer true - we use HHAST and hh-apidoc

kunigami: sent an update

PR #641, opened 2019-05-06
Update 01-getting-started.md

brianduff:
Update getting started guide to use <?hh instead of <?php.

This eliminates the following error following these instructions:
Fatal error: HHVM 4+ does not support PHP: <filename>

### DIFF
--- a/guides/hhvm/00-getting-started/01-getting-started.md
+++ b/guides/hhvm/00-getting-started/01-getting-started.md
@@ -30,7 +30,7 @@ hhvm -m server -p 8080
 Once you have HHVM running, write a simple "Hello World" program named `hello.php`:
 
 ```
-<?php
+<?pp
 echo "Hello World!";
 ```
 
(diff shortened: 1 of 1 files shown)


SiebelsTim:
Do you mean `hh`? What about file extensions that are mentioned in that chapter?

fredemmott:
This change is an improvement in itself (thank you!), but the example should be reworked to:
- be a `.hack` file (no header)
- have an entrypoint function

While it's the very first example so there's a pretty strong argument to made for simplicity, I think that's outweighted by top-level statements not being supported in .hack files or `<?hh // strict` files at all - and other modes are unsafe extremely strongly discouraged.

As this is definitely an improvement in itself, I can either merge this as-is (thanks again!) and file an issue, or it could be fixed in here? Might need corresponding text changes for it to make sense in context

brianduff:
I'll rework into a new pull request.

PR #642, opened 2019-05-06
Update 01-getting-started.md

brianduff:
This eliminates the following error following these instructions:
Fatal error: HHVM 4+ does not support PHP:

### DIFF
--- a/guides/hhvm/00-getting-started/01-getting-started.md
+++ b/guides/hhvm/00-getting-started/01-getting-started.md
@@ -37,7 +37,7 @@ function main(): noreturn{
 }
 ```
 
-Save this `hello.hack` in the same directory that you ran the `hhvm` command from above. Then, load [http://localhost:8080/hello.php](http://localhost:8080/hello.php) in your browser and verify you see "Hello World!" appear.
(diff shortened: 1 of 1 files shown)


guides/hhvm/00-getting-started/01-getting-started.md
```diff
@@ -27,14 +27,17 @@ hhvm -m server -p 8080
...
-echo "Hello World!";
+<<__EntryPoint>>
+function main(): noreturn{
+  echo "Hello World!";
+  exit(0);
+}
 ```
 
-Save this `hello.php` in the same directory that you ran the `hhvm` command from above. Then, load [http://localhost:8080/hello.php](http://localhost:8080/hello.php) in your browser and verify you see "Hello World!" appear.
+Save this `hello.hack` in the same directory that you ran the `hhvm` command from above. Then, load [http://localhost:8080/hello.php](http://localhost:8080/hello.php) in your browser and verify you see "Hello World!" appear.
```
fredemmott: there's another hello.hack here

brianduff: Fixed,

fredemmott:
Thanks again!

PR #643, opened 2019-05-06
Update build.php

brianduff:
Attempt to fix travis ci build error:

Fatal error: Uncaught Error: Call to undefined function HHVM\UserDocumentation\array_key_exists() in /var/source/bin/build.php:80

### DIFF
--- a/bin/build.php
+++ b/bin/build.php
@@ -77,8 +77,8 @@ function build_site(?Traversable<string> $filters = null): void {
   }
 }
 
-if (array_key_exists(1, $argv)) {
-  build_site(array_slice($argv, 1));
+if (\array_key_exists(1, $argv)) {
+  build_site(\array_slice($argv, 1));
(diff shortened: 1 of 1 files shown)


brianduff:
Nope :)

Fatal error: Uncaught exception 'Facebook\HHAST\SchemaVersionError' with message 'In file "/var/source/api-sources/hhvm/hphp/system/php/filter/filter_input_array.php": AST version mismatch: expected '2019-03-01-0001' (4.1.0), but got '2019-04-03-0001'' in /var/source/vendor/hhvm/hhast/src/entrypoints.hack:20

fredemmott:
yeah, needs dependencies updating to target a more recent version of HHVM

unfortuantely, thats' held back by the travis test failing anyway, even when it passes locally in the same docker image - I think it's hitting some resource limit in number of subprocesses.

PR #656, opened 2019-05-31
Switch from explicit in-test paralleization to more standard run of h…

fredemmott:
…h_server

Clearer test output, and simpler.

Hopefully works on Travis :(


fredemmott:
doesn't work.

PR #661, opened 2019-06-04
Replace submodules with a fetch build step.

fredemmott:
Pull in HHVM 3.30's pure-Hack implementation of tar (used for
`PharData`) and use that to extract github release tarballs.

Benefits:
- it's extremely common for new contributors to miss the submodule step
- it's *much* faster - especially given that people/Travis will often
  reflexively include `--recursive` in the git submodule command
- only extracting the parts of the tree we need - the API definitions.


Wilfred:
Definitely nicer, thanks :)

PR #668, opened 2019-06-21
Rework function chapter of docs

Wilfred:
Simplify the examples (we don't need to show tuples here), and
simplify the prose. Leverage subheadings to make it easier for readers
to find the sections relevant to their current need.

Ensure there's a single canonical page for documenting anonymous
functions, and set up redirects from the removed pages.

Separate function type syntax from anonymous function declaration
syntax.

Ensure we have examples of all supported anonymous function syntax
variants. Minimise example size to help newcomers.


Wilfred:
I've deliberately included the anonymous function examples in the markdown files rather than in separate example files.

The problem is that:

```
$f = $x ==> $x + 1;
$two = $f(1);
```

Is a lot easier to read than:

```
<?hh // strict

namespace Hack\UserDocumentation\Functions\Anonymous\Examples\Basic;

<<__EntryPoint>>
function main(): void {
  $f = $x ==> $x + 1;
  $two = $f(1);
  echo "$two\n";
}
```

I hope that's OK. I think we should look at optionally wrapping sample code in a `main`, to make this less verbose.

fredemmott:
Hi - the test failures looks related; are you still working on this?

fredemmott:
also #694

Wilfred:
I'm going to rework this content and land it.

PR #688, opened 2019-07-22
Clean up wording around `__Memoize`

hgoldstein:
It's confusing to use the phrase "future requests" to describe `__Memoize` which explicitly does _not_ work across requests.


lexidor:
What do you think of this?
```
The presence of this attribute causes the method or function to automatically cache each value it returns, so future calls with the same parameters can retrieve the value from the cache. The set of parameters are hashed into a single hash key, so changing the type, number, and/or order of the parameters will run the function again cache the new value.
```

hgoldstein:
That's roughly what I came up with, FWIW. I'll just merge this as-is.

PR #698, opened 2019-08-03
[rfc] new (better?) class & function filtering

jjergus:
This uses a list of classes/functions downloaded from php.net (see included script) to determine if a function/class is HH-specific.

This results in a bunch of function/classes being added. See https://gist.github.com/jjergus/638f90531974f3b205dd5d025f855ffb for the full lists before/after and diff between them.

Most of the additions seem correct but there are some that we probably want to exclude (__hhvm_intrinsics, ImageMagick stuff, any others?).

We still use the original heuristics (in addition to the new list) -- these functions/classes would have been excluded if we only used the PHP list, but are included thanks to the heuristics:

```
> ~~~ class.DOMNamedNodeMap
> ~~~ class.DOMNode
> ~~~ class.DOMNodeList
> ~~~ class.ReflectionClass
> ~~~ class.ReflectionFunction
> ~~~ class.ReflectionMethod
> ~~~ class.ReflectionParameter
> ~~~ class.WeakRef
> ~~~ function.xhprof_disable
> ~~~ function.xhprof_enable
> ~~~ function.xhprof_sample_disable
> ~~~ function.xhprof_sample_enable
```


fredemmott:
looks good; could you extend the tests to cover this?

jjergus:
Final list after adding PHP functions from older documentation and other sources + a few manual additions: https://gist.github.com/jjergus/77563dbd1a1786d4f09ad3c5c75420c1

PR #700, opened 2019-08-05
Use hh-apidoc library

fredemmott:
hh-apidoc was split out from this website - but after that refactoring,
hh-apidoc was never actually removed from here. Switch to using the
library.

This reduces duplication, and the library supports documenting type
aliases while the old copy doesn't.

For now, skip documenting type aliases, and add that thoroughly in a
follow-up diff, so that this change is hopefully a no-op.


src/hh-apidoc-extensions/APIDefinitionTypeBasedPathProvider.php
```diff
@@ -71,4 +71,19 @@ public function getPathForTraitMethod(
...
+		);
+	}
+
+	public function getPathForOpaqueTypeAlias(string $alias): string {
+		invariant_violation(
+			'Opaque type aliases are not implemented: %s',
+			$alias,
+		);
+	}
+
```
jjergus: nit: indentation looks off here (tabs vs spaces?)

fredemmott:
Merged manually into master.

PR #702, opened 2019-08-07
commit HHAPIDocBuildStep build artifacts

jjergus:
Followup to #701

This commits the build artifacts from HHAPIDocBuildStep so that (thanks to the logic from #701) the build step can be skipped, even on completely new checkouts.

One thing that might be worrisome is that there are over 2000 files in api-markdown. As an alternative, I could put them all in a single .tar.gz archive (which would only be a few hundred kB) and only commit that. Do you think that would be preferable?


fredemmott:
Is this an optimization, or requires for ondemand to function?

jjergus:
Not required anymore, so we can revisit later (if/when we want to do fully incremental builds).

PR #709, opened 2019-08-14
ADD: two example for async

AlphaWong:
Hi Community,

As I currently learn hacklang, It seems the sample code in this page `https://docs.hhvm.com/hack/asynchronous-operations/some-basics` is not runnable.

# spec
```console
hhvm --version
HipHop VM 4.18.0 (rel)
Compiler: 1565634828_N
Repo schema: 87cc9f3f8ece4c477c3bcf388b96a2a3031e653
```

# issue
I find that if the run the sample code of 
```hack
<?hh // strict

namespace Hack\UserDocumentation\AsyncOps\Basics\Examples\AsyncCurl;
use namespace HH\Lib\Vec;

async function curl_A(): Awaitable<string> {
  $x = await \HH\Asio\curl_exec("http://example.com/");
  return $x;
}

async function curl_B(): Awaitable<string> {
  $y = await \HH\Asio\curl_exec("http://example.net/");
  return $y;
}

async function async_curl(): Awaitable<void> {
  $start = \microtime(true);
  list($a, $b) = await Vec\from_async(vec[curl_A(), curl_B()]);
  $end = \microtime(true);
  echo "Total time taken: " . \strval($end - $start) . " seconds\n";
}

<<__EntryPoint>>
function main():void {
  \HH\Asio\join(async_curl());
}
```
It will always throw following error
```console
Fatal error: Uncaught Error: Call to undefined function HH\Lib\Vec\from_async() in /Users/alphawong/works/src/github.com/alphawong/hello-hhvm/bin/async_func.hack:20
Stack trace:
#0 /Users/alphawong/works/src/github.com/alphawong/hello-hhvm/bin/async_func.hack(27): Hack\UserDocumentation\AsyncOps\Basics\Examples\AsyncCurl\async_curl()
#1 (): Hack\UserDocumentation\AsyncOps\Basics\Examples\AsyncCurl\main()
#2 {main}
```
When I Google it I find that I need to change `require_once(__DIR__.'/../vendor/autoload.hack');` to `require_once(__DIR__.'/../vendor/hh_autoload.hh);` to make it runnable.

So I modify the currently sample code by split it to two file based on the hhvm version 4 like following added commit in order to let the tutorial working.

Please review my change.

Best,
Alpha


AndrewDiMola:
Closing because this was fixed in https://github.com/hhvm/user-documentation/pull/1184. As stands in the language now, `Call to undefined function HH\Lib\Vec\from_async()` was fixed by importing the `Vec` HSL to the snippet.

PR #711, opened 2019-08-14
Document `??=`

lexidor:


guides/hack/03-expressions-and-operators/97-assignment.md
```diff
@@ -62,4 +62,31 @@ $list[$i++] += 10;
 
 When the left-hand expression contains a side-effect, that side-effect is performed only once.
 
-The complete set of compound-assignment operators is: `**=`, `*=`, `/=`, `%=`, `+=`, `-=`, `.=`, `<<;=`, `>>=`, `&=`, `^=`, and `|=`.
+The complete set of compound-assignment operators is: `??=`, `**=`, `*=`, `/=`, `%=`, `+=`, `-=`, `.=`, `<<;=`, `>>=`, `&=`, `^=`, and `|=`.
+
+## Operator ??=
+
+The `??=` operator can be used for conditionally writing to a variable if it is null or not _set_.
+The rules around when something is _set_ are rather complex.
```
Wilfred: Could we state the rules rather than saying they're complex and giving some examples?

I believe that for a variable, it's just whether it's `null`. For a Hack array or PHP array, it's whether the value is present and `null`, or not present.

If you split up your example and add a sentence, that would be perfect :)

guides/hack/03-expressions-and-operators/97-assignment.md
```diff
@@ -62,4 +62,31 @@ $list[$i++] += 10;
...
+
+```HACK
+$nully = null;
+$nully ??= 'I assign'; // Assigns "I assign" to $nully
+$nully ??= 'I do not assign' // Does not assign, since $nully is set
```
Wilfred:
```suggestion
$nully ??= 'I do not assign'; // Does not assign, since $nully is set
```

guides/hack/03-expressions-and-operators/97-assignment.md
```diff
@@ -62,4 +62,31 @@ $list[$i++] += 10;
...
+```
+
+It might be helpful to think about it in the expanded form:
+ - `$nully = $nully ?? 'value';`
+ - `$dict['not_key'] = $dict['not_key'] ?? 'value';`
+
+For the exact rules of what is _set_, visit [operator ??](/hack/expressions-and-operators/coalesce).
+
+It is important to keep in mind that the expression on the RHS is only evaluated if the RHS is not _set_.
+So if your RHS is a function call, it will be called conditionally.
```
Wilfred: Perhaps give an example here?

```
$foo = 42;
$foo ??= 1 / 0; // no exception here!
```

lexidor:
@Wilfred I would love to make your adjustments and move them to the `Operator ??` file. I'll refer to the rules of what is set to that file instead.

We mostly document strict hack here, so we skip the undeclared property setness on purpose _or_ include it for completeness?

guides/hack/03-expressions-and-operators/97-assignment.md
lexidor: I have expanded `Operator ??`. That is a better place for these definitions.

guides/hack/03-expressions-and-operators/97-assignment.md
lexidor: This example is used in the new `Operator ??`.

lexidor:
fixes #712

jjergus:
I integrated changes from here into #729 to avoid merge conflicts

PR #724, opened 2019-09-19
Document the __Enforceable attribute

vassilmladenov:
I noticed that the examples directory didn't have the `-examples` suffix, so I added it.


jjergus:
Awesome, thanks!

PR #725, opened 2019-09-19
Dynamic coercion documentation

vassilmladenov:
This adds a Coercion section to the documentation for the `dynamic` type and clarifies a few points re: subtyping.

Here is a preview: http://34.220.76.150:8080/hack/built-in-types/dynamic


jjergus:
Thanks! Looks like there is a test failure because of some Hack error in one of the examples. Can you fix that one? The output is at the very bottom of https://travis-ci.org/hhvm/user-documentation/builds/587251626 and you can probably reproduce using `./vendor/bin/hacktest tests/ExamplesTest.hack`

jjergus:
Awesome, thanks!

PR #726, opened 2019-09-20
Update documentation for inout parameters to cover restrictions correctly

dneiter:
Previously the doc explicitly mentioned possiblity of passing a class property via inout. This was never supported for inout parameters and there are no plans to support it.
Update documentation to clearly state restrictions on inout arguments (local variables, possibly with some index access chained on them) and remove the explanation about type invariance - it's not applicable, because we do not support typehints on local variables. It may be worth writing another paragraph on how typing works for inout, but I'm going to punt on wordsmithing that for now.


jjergus:
Awesome, thanks! I'll merge it in as soon as the tests pass.

PR #730, opened 2019-09-27
magically strip boilerplate from examples

jjergus:
See comments in code for explanation of what gets stripped.

There are incantations available to override the magic, which I used in 2 examples where the boilerplate is actually important.

Build diff: https://gist.github.com/jjergus/b7136a277e2a44d6b65556f5bbe5bc90


jjergus:
Good point, I guess that depends on whether the start/end markers are placed "reasonably".

The alternative approach seems a bit hacky so I'd probably keep this for now, at least until someone comes up with a legitimate use of markers placed in some weird place that breaks highlighting.

+apparently we're doing the highlighting in JS on the client right now anyway

PR #742, opened 2019-10-25
rewrite "constructor parameter promotion" docs

### LINKED ISSUE #736: Awkward grammatical flow

# Where is the problem?

https://docs.hhvm.com/hack/classes/constructors

# What is the problem?

In the paragraph beginning as

> Can we use this approach with the `Point` class? ,

it's not clear what "this approach" is referring to.

 - Build ID: HHVM=HHVM-4.25.0:HSL=v4.15.2:2019-10-07T23:29:02+0000:c118b76b199117aab2681ec0b752ae5739343f09
 - Page requested: /hack/classes/constructors
 - Page requested at: Wed, 09 Oct 2019 19:20:35 +0000

### LINKED ISSUE #741: Missing documentation on public and protected constructor parameters

# Where is the problem?

The docs on constructors:
https://docs.hhvm.com/hack/classes/constructors

# What is the problem?

There is no documentation that explicitly describes the behavior where constructor parameters annotated as "public" or "protected" become class instance members:

```
class A { 
  public function __construct(public int $x) {} 
}
(new A(3))->x; // 3
```

The old docs described much more explicitly: https://web.archive.org/web/20170706193234/https://docs.hhvm.com/hack/other-features/constructor-parameter-promotion

 - Build ID: HHVM=HHVM-4.28.0:HSL=v4.25.0:2019-10-24T22:48:17+0000:20b292e78a5083871964d26362f50b9037c536b0
 - Page requested: /hack/classes/constructors
 - Page requested at: Fri, 25 Oct 2019 05:00:39 +0000

jjergus:
- adds a new "constructor parameter promotion" section to the "constructors" page
- copied over most of the original text from https://github.com/hhvm/user-documentation/blob/f74fda1e716ad2fd60cacc870c5174c2f8dc1dce/guides/hack/14-other-features/02-constructor-parameter-promotion.md
- moved the arguments against them to a new, shorter subsection
- short subsection added to the "properties" page that links here

Fixes #736
Fixes #741


guides/hack/06-classes/16-constructors.md
```diff
@@ -11,28 +11,49 @@ being incremented.  Note that a constructor may call any *private* method in its
...
+we've designed it. Our private properties have type `float`, so any arithmetic
+coordinate value can be represented, yet we've declared the parameters to the
+constructor to have type `num`, so either integer or floating-point values can
+be passed in.
+
+As this example illustrates, it is often valuable to keep the internal data
+representation separate from the public-facing constructor, which is impossible
+with promoted constructor parameters.
+
+Using promoted parameters also makes it harder to change the internal
```
fredemmott: Is there an example of this? I don't see how it's at all true - the "That said" part below requires absolutely no changes to callers.

This is true for `public` promoted parameters, however that's just because it's true for all `public` properties - using parameter promotion does no difference.

guides/hack/06-classes/16-constructors.md
```diff
@@ -11,28 +11,49 @@ being incremented.  Note that a constructor may call any *private* method in its
...
+
+Can we use constructor parameter promotion with the `Point` class? No, not as
+we've designed it. Our private properties have type `float`, so any arithmetic
+coordinate value can be represented, yet we've declared the parameters to the
+constructor to have type `num`, so either integer or floating-point values can
+be passed in.
+
+As this example illustrates, it is often valuable to keep the internal data
+representation separate from the public-facing constructor, which is impossible
+with promoted constructor parameters.
```
fredemmott: I know this is copy-paste from the previous version, but I'd like this removed: it does not affect callers at all.

This is as true as saying that "it's impossible to directly return an awaitable without awaiting it if you have `async` on your function declaration header"

Both this and the previous paragraph should probalby be replaced with a generic warning about public/protected properties in general, but there is no additional harm from using paramter promotion.

guides/hack/06-classes/16-constructors.md
jjergus: If you need to change the internal representation, you have to first un-promote the params, so that technically makes it "harder", but yeah you're probably right, that's pretty trivial so probably not worth mentioning.

guides/hack/06-classes/16-constructors.md
jjergus: Yeah you're probably right, I should make that section even shorter or maybe just remove it.

I guess the main argument is not as much that promoted params are bad, but that they could lead people towards bad design decisions (making your constructor take arguments equal to your private properties, instead of some more reasonable public API).

jjergus:
- now only a single paragraph about when not to use it (with less negative language)
- another paragraph with a positive spin on using promotion + un-promoting something later in a backwards-compatible way


Ffeel free to suggest more changes even after I merge this, I can apply them later in a separate commit (or feel free to just commit any changes).

PR #743, opened 2019-10-26
Add example with protected and public annotations

robz:
Addresses #741

### DIFF
--- a/guides/hack/06-classes/16-constructors-examples/visibility-on-all-parameters.php
+++ b/guides/hack/06-classes/16-constructors-examples/visibility-on-all-parameters.php
@@ -1,0 +1,12 @@
+<?hh // strict
+
+namespace Hack\UserDocumentation\Classes\Constructors\Examples\ParmVisibiity;
+
+class C2 {
+  public function __construct(
+    protected int $pr1,
+    private int $p2,
+    public int $pr2,
+  ) {
+  }
+}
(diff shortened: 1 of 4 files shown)


robz:
Oh shoot I just saw https://github.com/hhvm/user-documentation/pull/742
That one looks better, abandoning

jjergus:
Actually that one doesn't include any public or protected ones (only private) so this might still be useful to add, but...can you think of any example where public instance variables (and to a lesser extent protected too) are not an anti-pattern?

robz:
I don’t have a strong opinion, but I don’t know why they would ever be an anti patterns. I also don’t particularly understand the argument for discouraging private members.

But in any case, having documentation would be great since this pattern is in my codebase a lot so I’d like to share it with folks who are new.

If you think they should be discouraged, this would also be a great place to explain it.

jjergus:
New documentation is live at https://docs.hhvm.com/hack/classes/constructors#constructor-parameter-promotion, let me know if anything is missing.

> I don’t have a strong opinion, but I don’t know why they would ever be an anti patterns.

In general, it's better to use private variables and make them accessible from outside the class (if needed) using getter/setter methods. This has various advantages, for example:

- you can have variables that only have a getter and no setter, effectively making them "read only"
- setters can be updated to do additional validation (e.g. that a string is non-empty, or that an integer value for something like "count" is non-negative, etc.)
- if you decide to change any private variables, you can keep your class backwards-compatible by keeping the old getters/setters (just the body of each getter/setter would need to be updated to handle the change)
- more complicated/interesting use cases: for example, for a private array variable, you could decide to expose a method that adds to the list, but not expose any methods that change/remove elements
- ...

PR #744, opened 2019-10-29
delete dead code

jjergus:
I think most of this has been replaced by hh-apidoc, but there's a bunch of other stuff too. Let me know if anything seems suspicious, sometimes I get carried away :)


src/typedefs.php
```diff
@@ -11,16 +11,6 @@
 
 namespace HHVM\UserDocumentation;
 
-use type Facebook\DefinitionFinder\ScannedDefinition;
-
-type PHPDotNetAPIIndexEntry = shape(
```
fredemmott: not HHAPIDoc, but yes, dead - search used to intentionally include links to php.net

fredemmott: rest is mostly hhapidoc

PR #747, opened 2019-11-08
Remove Tuple\from_async and Asio\va

fredemmott:
- special-casing removed from typechecker in next release
- Tuple\from_async removed from hsl master
- Asio\va removed from HHI in next release

### DIFF
--- a/src/build/HHAPIDocBuildStep.php
+++ b/src/build/HHAPIDocBuildStep.php
@@ -24,7 +24,7 @@ use type Facebook\DefinitionFinder\{
 use namespace Facebook\{HHAPIDoc, TypeAssert};
 use namespace Facebook\HHAPIDoc\Documentables;
 use type Facebook\HHAPIDoc\Documentable;
-use namespace HH\Lib\{C, Dict, Str, Tuple, Vec};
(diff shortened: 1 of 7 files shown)


src/build/HHAPIDocBuildStep.php
```diff
@@ -47,10 +47,12 @@ public function buildAll(): void {
...
     Log::i("\nParsing builtins");
-    list($runtime_defs, $hhi_defs) = \HH\Asio\join(Tuple\from_async(
-      self::parseAsync($runtime_sources),
-      self::parseAsync($hhi_sources),
-    ));
+    list($runtime_defs, $hhi_defs) = \HH\Asio\join(async {
+      return tuple(
+        await self::parseAsync($runtime_sources),
+        await self::parseAsync($hhi_sources),
+      );
```
jjergus: This makes it not concurrent anymore, doesn't it? Is wrapping this in `concurrent` legal? If not, we may need something like

```
concurrent {
  $a = ...
  $b = ...
}
return tuple($a, $b);
```

fredemmott: multiple awaits in the same expression are concurrent

PR #752, opened 2019-11-19
Clarify the purpose of `dynamic` more

fredemmott:
Received feedback that "This special type is used to help capture dynamism in
the existing codebase in typed code, in a more manageable manner than `mixed`"
implied that `dynamic` was usually preferable to `mixed`; I don't think
this is the case as `mixed` is a real supertype-of-all.

### DIFF
--- a/guides/hack/11-built-in-types/71-dynamic.md
+++ b/guides/hack/11-built-in-types/71-dynamic.md
@@ -1,16 +1,26 @@
-**This type is intended to help with code being transitioned from untyped mode to strict
-mode.**
-
 Although `dynamic` can be used as the type of a class constant or property, or a function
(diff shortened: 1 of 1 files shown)


fredemmott:
David has some feedback about this - in particular, `dynamic` can be safer than `($mixed as foo)->bar()` if you are not absolutely certain that `mixed` is a `foo`

fredemmott:
Sorry, I still need to incorporate David's feedback

fredemmott:
whoops, did not mean to close it; still needs feedback addressing though

Wilfred:
Perhaps worth mentioning soft types? That gives you all the checking of hard types, but without the runtime risks.

Issue #673, opened 2019-06-28
[Missing docs] `nothing` is not described


lexidor:
Please complete the information below:

# Where is the problem?

This is a summation of built-in type. `Nothing` is not mentioned. 

# What is the problem?

`Nothing` would be a quite interesting type to have documentation on, since you are starting to see it when hovering over incomplete code in HHVM versions that use the new inference.

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: HHVM=HHVM-4.10.0:HSL=v4.7.0:2019-06-24T22:51:21+0000:24194533a68f50062246986e35b8afb154180909
 - Page requested: /hack/built-in-types/introduction
 - Page requested at: Fri, 28 Jun 2019 13:01:44 +0000
 - Controller: GuidePageController

lexidor:
```HACK
$vec = vec[];
\HH\Lib\Vec\map($vec, $e ==> $e);
```
Hovering over `$e` tells you that the type is `nothing`.

PR #757, opened 2019-12-03
Document `nothing` type

allewun:
Addresses #673. 

This (https://github.com/facebook/hhvm/commit/5543eb2f765eb3fd988a899cc3797d30a8057b48 and https://github.com/facebook/hhvm/commit/5543eb2f765eb3fd988a899cc3797d30a8057b48) was the only official information I could find on `nothing`, please feel free to correct or amend with more details!

### DIFF
--- a/guides/hack/11-built-in-types/74-nothing.md
+++ b/guides/hack/11-built-in-types/74-nothing.md
@@ -1,0 +1,2 @@
+The `nothing` type is a bottom type and has the exact same semantics as the [`noreturn`](/hack/built-in-types/noreturn) keyword. 
+However, the two types behave differently with respect to type-checking.


guides/hack/11-built-in-types/74-nothing.md
```diff
@@ -0,0 +1,2 @@
+The `nothing` type is a bottom type and has the exact same semantics as the [`noreturn`](/hack/built-in-types/noreturn) keyword.
```
fredemmott: This isn't really accurate; if it has different typechecker behavior, it has different semantics.

allewun: True, though it's what I pulled out of the source. I don't really have any more context than those comments. Open to suggestions here - should I just remove that misleading clarification?

lexidor: I have only got my own experiences to rely on, but I can write something.

Wilfred: We plan to make `noreturn` the same as `nothing`. The nothing type is more general and is integrated into type inference. `noreturn` is an older implementation, isn't integrated into the type system, so has some limitations.

`noreturn` isn't a subtype of other types, so the following code only works for `nothing`:

```
function throw_error(): nothing {
  throw new Exception('oh dear');
}

function bool_to_int_or_error(bool $b): int {
  if ($b) {
    return 1;
  } else {
    return throw_error(); // this is an error for noreturn
  }
}
```

However, we have some dead code analysis for `noreturn` that isn't yet available for `nothing`, so the following code only works for `noreturn`:

```
function throw_error(): noreturn {
  throw new Exception('oh dear');
}

function bool_to_int_or_error(bool $b): int {
  if ($b) {
    return 1;
  } else {
    throw_error();
  }
}
```

Summary: users should be using `nothing` wherever possible, it's a proper bottom type.

jjergus:
Closing this one since discussion has moved to #765

PR #760, opened 2019-12-13
Add automatically generated documentation for HSL IO

fredemmott:
Including `Debug\` in this because var_dump becomes unusable with async stdout


src/APIIndex.php
```diff
@@ -25,6 +25,9 @@ private function __construct(private APIProduct $product) {
       case APIProduct::HSL:
         $this->index = $idx[APIProduct::HSL];
         break;
+      case APIProduct::HSL_EXPERIMENTAL:
+        $this->index = $idx[APIProduct::HSL_EXPERIMENTAL];
```
fredemmott: doing this instead of a lookup because there's a shape here.

src/build/HHAPIDocBuildStep.php
```diff
@@ -71,14 +71,26 @@ public function buildAll(): void {
     );
 
     Log::i("\nFinding HSL sources");
+    // prefixe whitelisted are handled when extracting the tarballs
```
Wilfred:
```suggestion
    // prefix whitelisted are handled when extracting the tarballs
```

fredemmott:
rebased, fully rewrote that comment to be much clearer

PR #765, opened 2019-12-14
Document the type `nothing`

### LINKED ISSUE #673: [Missing docs] `nothing` is not described

lexidor:
Fixes #673 
Extends #757

The `$person->getOutOfMyCar()` error is intentional. Should I HH_FIXME it? That loses the point of the test typechecker though.


lexidor:
> _not_ impossible

We both realized that mistake at the same time. This was an unintended double negation.

Thanks for clarifying the `return <<nothing>> ~=~ [[terminal]]` thing. I never used `nothing` as a return, so I never figured out that this was allowed.

I'll gladly add the things I learned today to the docs.

Wilfred:
To clarify the `CallbackBase` example: functions are contravariant in their return type, so you couldn't write this code without `nothing`. I'm trying to show that sometimes you need `mixed` for a base class, and sometimes you need `nothing`.

`vec` is also contravariant, so perhaps this is a simpler example:

```hack
abstract class BaseProcessor {
  abstract public function processItems(vec<nothing> $_): void;
}

class NumberProcessor extends BaseProcessor {
  <<__Override>>
  public function processItems(vec<int> $items): void {
    foreach ($items as $item) {
      echo Str\format("Processing number: %d", $item);
    }
  }
}

class StringProcessor extends BaseProcessor {
  <<__Override>>
  public function processItems(vec<string> $items): void {
    foreach ($items as $item) {
      echo Str\format("Processing string: %s", $item);
    }
  }
}
```

jjergus:
I like Wilfred's contravariance examples -- this is probably the most useful thing you can do with the `nothing` type, so I would use that as the main example instead of the current one.

The current example, which demonstrates how you can call a non-existent function on a `nothing` value, seems a bit convoluted to me, and also seems like an antipattern that we probably shouldn't encourage.

We could mention somewhere in the text that you can do anything with a `nothing` value (call non-existent functions, assign to variables of any type, ...) but I'm not sure how useful that is, I wouldn't spend too much time on it.

> vec is also contravariant, so perhaps this is a simpler example

vec is not contravariant, I think you meant method arguments in general? Your example should work without the vec (`processItems(nothing $_)`), which makes it even simpler.

jjergus:
Although even those contravariance examples seem a bit unrealistic -- why would anyone declare a function that takes `nothing` and is therefore impossible to call?

I'm trying to come up with a more realistic example, so far I came up with this:

```hack
interface Animal<-TFood> {
  public function getName(): string;
  public function eat(TFood $food): void;
}

function get_first_name(Animal<nothing> $animal): string {
  // We don't care what $animal eats (because we only need its name)
  // so we can use Animal<nothing>.
  return C\firstx(Str\split($animal->getName(), ' '));
}
```

jjergus:
Also this is not very realistic but I think demonstrates the concept nicely:

```hack
interface Example<+TCovariant, -TContravariant> { ... }

function takes_any_example(Example<mixed, nothing> $any_example): void { ... }
```

Wilfred:
> vec is not contravariant, I think you meant method arguments in general?

Woops, you're right.

I agree it's really hard to find good examples for `nothing`. I think that's why it's such a new feature: whilst it's useful, it's not useful very often, and usually inside larger frameworks with lots of generics.

Your examples are correct, but I'm hoping we can avoid talking about `-Foo` and `+Foo` syntax in this section. It's already got some new concepts on it, and there are built-in familiar types that have both types of variance.

Wilfred:
> why would anyone declare a function that takes nothing and is therefore impossible to call?

I suppose it's a base class just defining the overall structure of child classes. The method itself is abstract at least.

> We could mention somewhere in the text that you can do anything with a `nothing` value

This is true, but I'm not sure this is useful to talk about. I'm really struggling to think of an example where it's useful that you can do anything inside dead code.

It's similar to the `int_to_bool_or_error` example really.

lexidor:
I am incredibly sorry that this is taking so long. It just got very busy around the holiday season. I have not forgotten about this, it will just have to wait until early January.

jjergus:
> I'm hoping we can avoid talking about `-Foo` and `+Foo` syntax in this section

You're right, that's probably better to avoid. So your `BaseProcessor` example (maybe simplified, without the `vec`) is probably the best option.

> there are built-in familiar types that have both types of variance.

I can only think of covariant ones (most containers); what's a familiar built-in type with a contravariant generic param? It would probably be worth including an example using that type if we can think of one -- the example could demonstrate that `SomeType<nothing>` is the parent of all `SomeType`s and maybe contrast it with `SomeOtherType<mixed>`.

jjergus:
This looks good to me. I personally like the more informal/personal style but it's a bit inconsistent with the rest of the documentation so others may have different preferences. I'll leave it open for another day or so, then merge it in if no one complains in the meantime.

lexidor:
I think that mostly stems from the fact that I feel like it is a very difficult topic to wrap your head around. I guess I just started explaining it to the reader in the way I would like someone to explain it to me. I have also reread a lot of the [Hack & HHVM](https://www.goodreads.com/book/show/26401259-hack-and-hhvm) book recently. Which has these more informal sections everywhere. I didn't plan it to turn out this way, but I am still quite happy with the result. It feels incomplete, however, I still need to add the `return thisisverybad()` thing and how it can be used in the place of `noreturn`. However, this doesn't work inside of void functions or `__construct`.
```HACK
function bad(): nothing {
  invariant_violation('bad');
}

function void(): void {
  return bad(); // Typing[4084] You cannot return a value. This is a void function.
}

class C {
  // Hack models constructors as return `void`, instead of `this`,
  // which would be the perfect fit.
  public function __construct() {
    return bad(); // Typing[4084] You cannot return a value. This is a void function.
  }
}
```
Is this a case of "This should be okay, but we're not allowing it right now" or is there a real reason why `nothing` is not a subtype of (doesn't "extend") `void`? Is `void` even a type or just a piece of syntax type you only find where you would expect a type? 

 I've learned more about `nothing` writing this than I expected. It is indeed out of place amongst the rest of the pages.

fredemmott:
> I personally like the more informal/personal style but it's a bit inconsistent with the rest of the documentation

I find the informal/personal style much more readable; in some ways, a lot of the documentation can read like a specification or technical reference rather than tutorial text, and personally I appreciate changes that move us away from that :)

jjergus:
I think the distinction between `noreturn` and `nothing` (and possibly `void`) is largely historical and there's a plan to clean it up at some point, but it hasn't been the highest priority thing so far...

fredemmott:
Currently, the important difference is that `nothing` supports subtyping, but `noreturn` does not

Wilfred:
> I think the distinction between noreturn and nothing (and possibly void) is largely historical and there's a plan to clean it up at some point

Yep, so we do want to eventually remove `noreturn` and just have `nothing`.

Note that `void` is different. A `void` function is allowed to return without errors. From a type system perspective, it's a singleton type: there's only one value it can return. `nothing` has zero values it can return.

> However, this doesn't work inside of void functions or __construct.

```
function bad(): nothing {
  invariant_violation('bad');
}

function void(): void {
  return bad(); // Typing[4084] You cannot return a value. This is a void function.
}
```

We take the view that you should never have `return some_expression;` in a `void` function. You're absolutely right that this works from a type system/subtyping perspective, but we choose to ban it because we think it makes the code clearer.

You can make the same argument with `mixed`. From a type system/subtyping perspective, this is fine, but we choose to ban it too:

```
function my_example(): mixed {
  return;
}
```

I like to think of the requirement of `return;` only in `void` functions as a lint on top of the type system.

> I still need to add the `return thisisverybad()`

Yeah, for `noreturn` functions (e.g. `invariant_violation`) Hack understands that execution doesn't go beyond them. The control flow analysis isn't doesn't know this about `nothing` yet. As a result, the following doesn't work with `nothing`:

```
// Changing this to noreturn fixes the type error.
function give_up(): nothing {
  throw new Exception('giving up!');
}

function int_to_bool_or_error(int $i): bool {
  if ($i === 0) {
    return false;
  } else if ($i === 1) {
    return true;
  } else {
    give_up();
  }
}
```

We want to teach control flow analysis about `nothing`, but it's a non-trivial change to roll out (prototyped in D14420275).

jjergus:
To further complicate things, at runtime, empty or missing `return` statement actually means `return null;` (a behavior inherited from PHP)

PR #766, opened 2019-12-16
Add fb-specific documentation link to 'Getting Started' page

fredemmott:
This guide covers how to set up various parts of the environment that
are automatically configured for FB engineers; it's not useful for them.

Test plan:

- opened page on localhost, saw nothing
- deleted `isNotFbViewer` class on body, saw text and link
- clicked link, got internal docs

### DIFF
--- a/guides/hack/01-getting-started/01-getting-started.md
+++ b/guides/hack/01-getting-started/01-getting-started.md
@@ -1,3 +1,10 @@
+```yamlmeta
+{
+  "fbonly messages": [
+    "Unless you are specifically working on open source Hack code, you want [Facebook's internal documentation](https://our.internmc.facebook.com/intern/wiki/First-app/) instead."
+  ]
+}
+```
 ## Overview
 
 The prerequisites you need to write and execute Hack code are pretty straightforward:


fredemmott:
Screen Shot 2019-12-16 at 10 16 21 AM

azjezz:
Unless FACEBOOK is not planning on opening hack/hhvm for adoption outside FACEBOOK, I don't think this should go in public documentation.

I believe it should be the other way around. in FACEBOOK's private documentation : 
```
If you are working on an open source hack project, you want the [public](https://docs.hhvm.com) documentation.
```

fredemmott:
These sections are only visible by default (without editing CSS) if you are connected from a Facebook corporate IP address (based on the ASN)

fredemmott:
The primary reason for doing this is that if FB employees look at internal docs first instead, they will also *improve* the internal docs instead of improving the public docs.

azjezz:
> The primary reason for doing this is that if FB employees look at internal docs first instead, they will also improve the internal docs instead of improving the public docs.

but here, this is pointing from public to internal, not internal to public.

+ why have 2 different docs ? won't merging the two make more sense ?

fredemmott:
Language documentation is shared; the link documents:
- how to get a Facebook dev server using Facebook internal tools
- how to connect to it with Facebook’s IDE extensions
- how to add a page to facebook.com

It does not cover installation of Hack or HHVM. The FB content is not relevant externally, and the external content is not relevant internally.

azjezz:
It doesn't make sense that someone would go to hhvm/hack documentation looking for something facebook.com specific, such as how to add a new route/page ( which, i agree, should have its own internal documentation )

but, if the internal documentation is the same as hhvm/user-documentation + facebook internal things, it should be split into its own.

so whether you do work at Facebook, or you are part of the community, there's only 1 place to go for hack/hhvm documentation, and learn/contribute.

fredemmott:
> it should be split into its own.

This would require that the public docs be re-written in a facebook-specific markup language; if a new system internal system were made to avoid this, the content would be forked to the standard system. It would also *definitely* require moving the development workflow to an FB-internal source of truth (similar to the HSL); we do not sync entire repos from github on a regular basis, and the language is moving too fast for a manual process to be workable.

> so whether you do work at Facebook, or you are part of the community, there's only 1 place to go for hack/hhvm documentation, and learn/contribute.

This is the case for anything not fb-specific at present.

fredemmott:
4a92e1a5

jjergus:
We want FB people to consider docs.hhvm.com their primary source of documentation for Hack, because the more they use the site, the more likely they are to contribute improvements or at least report issues.

Historically this has been a big problem -- most FB people currently only look at internal documentation, and a lot of useful stuff ends up being only documented internally as a result.

I think it's reasonable to do changes like this if it helps us achieve that.

fredemmott:
> Historically this has been a big problem -- most FB people currently only look at internal documentation, and a lot of useful stuff ends up being only documented internally as a result.

While docs.hhvm.com is meant to be the primary docs, there's some valuable internal docs as even with things as they are now, people still prefer to write things internally; I'll hopefully be publishing some of that soon - and  in those cases, I will be editing the internal docs to link here instead of duplicating

PR #767, opened 2019-12-18
Update versions

fredemmott:
The big thing here is hsl-experimental: there's some interface changes,
and much more documentation.


fredemmott:
Automated change; will merge without review if test run is clean.

PR #768, opened 2019-12-18
Add intro page for the HSL (whoops...), and container cheat sheets

fredemmott:
- doesn't look like we actually had a page saying what the HSL is...
  whoops.
- replace hardcoded lists of functions with:
  - link to generated API references
  - less in-depth but more usable 'cheat sheets' from internal docs

`fbOnly` spans are visible if either:
- you are on FB corpnet
- you delete the `isNotFBViewer` class from `<body>`

### DIFF
--- a/guides/hack/01-getting-started/04-the-standard-library.md
+++ b/guides/hack/01-getting-started/04-the-standard-library.md
@@ -1,0 +1,30 @@
+## The Hack Standard Library (HSL)
+
+The Hack Standard Library is rapidly evolving, and currently distributed
+separately to HHVM; there are two GitHub projects/composer packages:
+
+- [hhvm/hsl](https://github.com/hhvm/hsl/): The Hack Standard Library
+- [hhvm/hsl-experimental](https://github.com/hhvm/hsl-experimental/):
+  Experimental features, which may be added to the Hack Standard Library in the
+  future
+
+Design goals include:
+- fits well with Hack's type system and overall design
+- internal consistency and predicatability
+- provide building blocks for clear composition instead of many high-level
+  functions
+
+HSL features are grouped by namespace; the namespace indicates the problem area,
+or - if more fine-grained separation is needed - by return type. For example:
+
+- `HH\Lib\Str` contains functions for interacting with strings, such as
+  `Str\contains()`
+- `HH\Lib\Dict` contains functions that return `dict`s, such as `Dict\map()`
+- `HH\Lib\Vec` contains functions that return `vec`s, such as `Vec\map()`
+- `HH\Lib\C` contains functions that operate on `C`ontainers, but do not
+  return or require a specific kind of container, such as `C\contains()`
+
+For a full list, see [the API reference](/hsl/reference/); there is also
+a table containing common functions for [container manipulation].
+
+[container manipulation]: /hack/built-in-types/arrays#using-dicts-keysets-and-vecs
--- a/guides/hack/11-built-in-types/22-arrays.md
+++ b/guides/hack/11-built-in-types/22-arrays.md
@@ -1,4 +1,4 @@
-The preferred object types for providing array-like storage and operations are
+The preferred object types for providing container-like storage and operations are
 `vec`, `dict`, and `keyset`.  These supersede the earlier Hack types
 [Vector, Map, and Set](#legacy-vector-map-and-set), and the
 [legacy array](#legacy-array) type inherited from PHP. Additional special types,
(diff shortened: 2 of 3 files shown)


fredemmott:
Probably best reviewed in markdown preview mode rather than in source view :)

guides/hack/01-getting-started/04-the-standard-library.md
```diff
@@ -0,0 +1,30 @@
...
+  future
+
+Design goals include:
+- fits well with Hack's type system and overall design
+- internal consistency and predicatability
+- provide building blocks for clear composition instead of many high-level
+  functions
+
+HSL features are grouped by namespace; the namespace indicates the problem area,
+or - if more fine-grained separation is needed - by return type. For example:
```
fredemmott: @kmeht anything to add here, or to the previous paragraph?

fredemmott:
Screen Shot 2019-12-18 at 12 47 36 PM

It's hard to see from the diff - the best part of the dif is things like this (with the `[f]` bits not being visible to most viewers)

guides/hack/11-built-in-types/22-arrays.md
```diff
@@ -193,208 +117,77 @@ To check membership in a keyset, use `C\contains` or `C\contains_key`. To remove
...
+
+| Operation                 | `vec`                        | `dict`                         | `keyset`
+| ---------                 | -----                        | ------                         | --------
+| Initialize empty          | `$v = vec[];`                | `$d = dict[];`                 | `$k = keyset[];`
+| Literal                   | `$v = vec[1, 2, 3];`         | `$d = dict['foo' => 1];`       | `$k = keyset['foo', 'bar'];`
+| From Another Container*   | `$v = vec($container);`      | `$d = dict($keyed_container);` | `$k = keyset($container);`
+| Keys from Container*      | `$v = Vec\keys($container);` | N/A                            | `$k = Keyset\keys($container);`
+| Add Elements              | `$v[] = 4;`                  | `$d['baz'] = 2;`               | `$k[] = 'baz';`
+| Bulk Add Elements         | `$v = Vec\concat($t1, $t2)`  | `$d = Dict\merge($kt1, $kt2)`  | `$k = Keyset\union($t1, $t2)`
+| Remove Elements           | `$v = Vec\concat(Vec\take($t1, $n), Vec\drop($t1, $n)` | `unset($d['baz']);` | `unset($k['baz']);`
```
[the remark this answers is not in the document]
fredemmott: yep, thanks

fredemmott:
> Why remove all the information about legacy arrays and the migration types?

Legacy arrays are being completely removed, we just need to flip the parser option; already done in www.

As for the migration types, the old content encourages migrating from varray/darray to vec/dict; we're not recommending this now.

For both, it's also out of date, and rapidly changing - e.g. the comment about key conversions is simply wrong:

```
hphpd> =array_keys(darray['123' => 'foo'])
=array_keys(darray['123' => 'foo'])
Array
(
    [0] => "123"
)
```

The only stuff that seems stable is that varrays basically act like vecs and darrays act like dicts.

fredemmott:
> Why remove all the information about legacy arrays and the migration types? I think we should keep


It also seems like the bulk of the information really belongs in documentation for Container/KeyedContainer - e.g. we don't need separate documentation on `foreach` for each type, or separate append documentation/specification for vec/vector/varray

guides/hack/11-built-in-types/22-arrays.md
```diff
@@ -193,208 +117,97 @@ To check membership in a keyset, use `C\contains` or `C\contains_key`. To remove
...
-Within this section, the term *array* means *legacy array*.
-
-Note: Legacy arrays are quite different to arrays in numerous mainstream languages. Specifically, legacy array elements need not have the
-same type, the subscript index need not be an integer (so there is no concept of a base index of zero or one), and there is no concept of
-consecutive elements occupying physically adjacent memory locations.
+`Vector`s, `Map`s and `Set`s are generally interacted with via instance methods on the classes.
 
-An array is a data structure that contains a collection of zero or more elements whose values are each accessed through a corresponding key.
-As the number of elements in an array can change at runtime, an array type does *not* include an element count.  Consider the following class
-property, which has an array type:
```
fredemmott: These are a mix of not-really-true, or just as true for hack arrays

fredemmott:
Sorry, I was too fast on saying it's inaccurate; while some is inaccurate, I think there's a big problem with it being repetitive, and also just too much content that hopefully doesn't affect most users; I feel that '`varray` means "I think it should be a `vec` and basically pretend it is one"' is the appropriate level of detail, though I 've added a bit more about the downsides.

If you'd still rather have more of the original content, I'll put it in verbatim in a 'legacy containers' page or something.

fredemmott:
squashed the trivial fixes, kept the various meaningful changes as separate commits

PR #773, opened 2019-12-27
XOR is not a valid operator | Remove XOR from spec

lexidor:
Remove the use of the PHP XOR operator in 37-await-as-an-expression-spec.md.


jjergus:
Good catch.

PR #776, opened 2020-01-02
document missing special comments

### LINKED ISSUE #774: Program structure partial not reflected as special recognized comment

# Where is the problem?
https://docs.hhvm.com/hack/source-code-fundamentals/comments

# What is the problem?

I may be wrong, but under [special recognized comments](https://docs.hhvm.com/hack/source-code-fundamentals/comments) `// partial` is not mentioned, despite it being listed on the  [program structure](https://docs.hhvm.com/hack/source-code-fundamentals/program-structure) page alongside `// strict`. Thanks!

 - Build ID: HHVM=HHVM-4.36.0:HSL=v4.36.0:2019-12-30T17:14:57+0000:2fbf0284beeb3f9eaf249c1cefcabcd98d115e74
 - Page requested: /hack/source-code-fundamentals/comments
 - Page requested at: Tue, 31 Dec 2019 21:51:49 +0000

jjergus:
Fixes #774

Any other I missed?

### DIFF
--- a/guides/hack/02-source-code-fundamentals/10-comments.md
+++ b/guides/hack/02-source-code-fundamentals/10-comments.md
@@ -31,5 +31,6 @@ A *single-line comment* starts with `//` or `#`, and ends with a new line, which
 
 A number of special comments are recognized; they are:
 * [`// FALLTHROUGH`](../statements/switch.md)
-* [`// strict`](program-structure.md)
-* `/* HH_IGNORE_ERROR[`*nnn*`] */`, where *nnn* is the compiler error number that is to be ignored.
+* [`// strict` or `// partial`](program-structure.md)
+* `/* HH_FIXME[`*nnnn*`] */` or `/* HH_IGNORE_ERROR[`*nnnn*`] */`, where *nnnn*
+  is the type checker error number that is to be ignored.


lexidor:
I can't think of any that you would have missed.

lexidor:
Things like `//@oss-enable` and friends don't count, right?

fredemmott:
> Things like //@oss-enable and friends don't count, right?

That's not special to Hack, it's just respected by one program written in Hack - the one we use for github <-> fb import/export:

- https://github.com/facebook/fbshipit/blob/577ad2017c54dbf8bc217ebd699b47502c8f285b/fb-examples/lib/config/FBShipItConfig.php-example#L732-L753
- https://github.com/facebook/fbshipit/blob/2fce6abb0870cd575fc2aca8503ddf3249c0c749/src/shipit/filter/ShipItConditionalLinesFilter.php

guides/hack/02-source-code-fundamentals/10-comments.md
```diff
@@ -31,5 +31,6 @@ A *single-line comment* starts with `//` or `#`, and ends with a new line, which
 
 A number of special comments are recognized; they are:
 * [`// FALLTHROUGH`](../statements/switch.md)
-* [`// strict`](program-structure.md)
-* `/* HH_IGNORE_ERROR[`*nnn*`] */`, where *nnn* is the compiler error number that is to be ignored.
+* [`// strict` or `// partial`](program-structure.md)
```
fredemmott: Add "(only in the header line of legacy file formats)" or similar?

fredemmott:
also 'just respected by one program', but /maybe/ HHAST_IGNORE_ERROR things should go in here given how prevalent it is - though it's probably less necessary given it has HHAST in the text.

Wilfred:
Looks correct to me.

Arguably `// strict` does nothing these days, since it's the default anyway.

PR #777, opened 2020-01-03
Fix some lval_as_expressions for hhvm 4.38

lexidor:
P.S. A note about `class Stack<T>`.
A vec<T> does not allow for implicit append.
push would throw an OutOfBoundsException
when the first value is pushed.
This does not remove the value for explaining the type system however.


guides/hack/03-expressions-and-operators/99-yield-examples/process-file.php
```diff
@@ -9,7 +9,11 @@ function getTextFileLines(string $filename): \Generator<int, string, void> {
   }
 
   try {
-    while ($textLine = \fgets($infile)) { // while not EOF
+    while (true) {
+      $textLine = \fgets($infile);
+      if (!$textLine) {
```
fredemmott: this is equivalent to the old code, but could you replace it with an explicit check for `=== false`, given falsey strings are reasonable here

guides/hack/15-asynchronous-operations/31-examples-examples/polling.php
```diff
@@ -10,7 +10,8 @@
 class Polling {
   private int $count = 0;
   public function isReady(): bool {
-    if ($this->count++ === 10) {
+    $this->count++;
+    if ($this->count === 11) {
```
fredemmott: nit, but `> 10` might be clearer about the intent here

fredemmott:
lgtm now, but leaving for @jjergus in combination with #778

guides/hack/03-expressions-and-operators/27-function-call-examples/array-of-closures.php
```diff
@@ -17,10 +17,27 @@ function(int $p) {
...
+
+  $v = $table[$incr->postIncr()]($incr->value()); // eval is L-to-R
   echo "\$v = $v\n";
 
-  $v = $table[$i](++$i); // eval is L-to-R
+  $v = $table[$incr->value()]($incr->preIncr()); // eval is L-to-R
   echo "\$v = $v\n";
 }
+
+class Incrementer {
```
jjergus: I don't have a strong opinion but given that this is meant to be a relatively minor example in the middle of a larger guide (https://docs.hhvm.com/hack/expressions-and-operators/function-call), I think it might be preferable not to make it even longer than it already is...

In my attempt (#778), I basically removed that part of the example completely which I think is fine, but it might also be fine to keep it if it doesn't make it much longer.

Maybe something like this?

```hack
$v = $table[next_number()](next_number());
echo "\$v = $v\n";

$v = $table[next_number()](next_number());
echo "\$v = $v\n";

...

function next_number(): Generator<int, int, nothing> {
  for ($i = 0; true; $i++) {
    yield $i;
  }
}
```

The pre/post-increment difference doesn't seem to be relevant for demonstrating the L-to-R behavior.

guides/hack/03-expressions-and-operators/27-function-call-examples/array-of-closures.php
```diff
@@ -17,10 +17,27 @@ function(int $p) {
     ), // times 5
   ];
```
jjergus: not your fault but while we're at it, replacing these with lambdas would significantly shorter the example without making it any less readable

fredemmott:
Traversable<T> could be more readable than the full generator type

lexidor:
I must agree here. If this is code included in a place where we're explaining function calls, people might not understand classes just yet. Let's keep it as simple as can be.

I also saw that Jan already packaged up some of my fixes and some of his own in his own PR. I think we'd better merge his fixes instead, since most of them were either equivalent or better.

lexidor:
Oh, I guess we both found our own changes to be inferior (I can't think of the proper word) to those made by the other. What's the plan now? We are very much stepping on the same code.

lexidor:
Oh, I see you withdrew your PR. In that case, I'll implement the suggestions and we merge this one.

guides/hack/03-expressions-and-operators/27-function-call-examples/array-of-closures.php
lexidor: I might be misunderstanding how this program works. Vecs are not indexable with generators. I see what you want here (a function that returns n + 1 where n is the amount of times the function has been called before - 1).

Doing that requires either:
 - Calling `next()` and `current()` on the generator, which can't be done.
 - A lambda capturing a `HH\Lib\Ref<int>` and using that as storage.
 - A class with a private static member to keep the current value in.
 - A static variable in the function (but that died in an hhvm version a long time ago).

I think I'll go with:
```HACK
/**
 * Returns a function which will return the next
 * number in a sequence, starting at 0, 1, 2, 3, etc.
 */
function sequence(): (function(): int) {
  $ticket = new Ref(-1);
  return () ==> {
    $ticket->value++;
    return $ticket->value;
  };
}
```

jjergus: Oh of course, I don't know what I was thinking. I guess a class (like you originally had) is really the simplest working example then, maybe just a simpler class, something like:

```hack
class Incrementer {
  private int $last = -1;
  public function next(): int {
    $this->last++;
    return $this->last;
  }
}
```

jjergus: (I'd guess that most readers probably have better understanding of classes than of higher order functions.)

lexidor: Yeah, now I don't retain the `$i++`, `$i`, and `++$i` semantics, this would do best.

jjergus:
Thanks a lot!

PR #778, opened 2020-01-08
target HHVM 4.39, remove assignments-as-expressions

jjergus:
The ban on assignments-as-expressions required changing some examples, in one case significantly reducing the example's usefulness (see the removed text, which used to explain the parts of the example that are no longer applicable), so requesting review to see if anyone has better suggestions.


jjergus:
Yep that one is better. I'll bump the version to 4.39 after that is merged.

PR #780, opened 2020-01-15
HHVM doesn't run PHP

lexidor:
Learning PHP, I am still leaving it in there, since a PHP dev still knows 75% of Hack, but this will become less relevant over time.


fredemmott:
ilikeit

PR #781, opened 2020-01-15
Remove mysql(i) and phar from the extensions list

lexidor:
Mysqli is 100% dead.
Phar is 100% dead.
Mysql is technically not dead, but it was deprecated in PHP
for security reasons. So... should we be supporting it still?


fredemmott:
I think this section should be entirely removed, or have a biiiiiiiiig warning at the top: I doubt we implement /any/ PHP extensions correctly now, e.g. ref -> inout, ban of optional ref params, etc. The PHP docs can definitely end up misleading now

fredemmott:
@jjergus what do you think ?

jjergus:
I don't have a strong opinion, I'm OK with removing any extensions.

Wilfred:
Agreed, I think we should remove things that are broken and unsupported.

PR #782, opened 2020-01-15
Remove throwaway mentions of instanceof

lexidor:
It was removed in hhvm 4.15.
It is still a keyword (in the runtime and in keywords.md).
The article instanceof.md still exists.
It has a big deprecation warning on top.
This might be worth removing too @reviewer.
_Edit: Sorry, I did not expect `@reviewer` to be an actual user._


Wilfred:
Nice catch, thanks!

PR #784, opened 2020-01-25
Remove is_null

### LINKED ISSUE #599: Usage of \is_null in example

# Where is the problem?

namespace Hack\UserDocumentation\Types\TypeRefinement\Examples\Test2;

# What is the problem?

The documentation shows the use of a type query function. Afaik, that was being phased out. Suggested to use null identity comparison or is-refinement using the int or nonnull type.

 - Build ID: 2019-03-08T20:07:04+0000:dc3e7ed5d57714a7e83c2aedbb04c60b26bab415
 - Page requested: /hack/types/type-refinement
 - Page requested at: Sat, 09 Mar 2019 16:01:05 +0000

lexidor:
It doesn't refine and it is bad taste in Hack to use the is_type functions
Fixes #599

### DIFF
--- a/guides/hack/03-expressions-and-operators/12-invariant.md
+++ b/guides/hack/03-expressions-and-operators/12-invariant.md
@@ -3,9 +3,9 @@ should never occur.  For example:
 
 ```Hack
 invariant($obj is B, "Object must have type B");
-invariant(!\is_null($p), "Value can't be null");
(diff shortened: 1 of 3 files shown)


Wilfred:
Thanks. I think we should still acknowledge the existence of `is_null`, but I totally agree that the examples you've changed are better.

PR #792, opened 2020-01-28
better Composer instructions

### LINKED ISSUE #790: Real Project setup is not easy to follow

# Where is the problem?

[Install Composer]()

# What is the problem?

It should link to
https://getcomposer.org/doc/00-intro.md#installation-linux-unix-macos
which has the full instructions (windows is below, could be a separate link on this page)

instead of the page it links to, which doesn't include how to actually run composer

Also while we're at it, we should explicitly add to this page that hhvm-autload should be added, instead of just linking to it.

 - Build ID: HHVM=HHVM-4.40.0:HSL=v4.40.0:2020-01-17T14:02:04+0000:9f1d392f2f951dfdad225c21329bf894f6577f36
 - Page requested: /hack/getting-started/starting-a-real-project
 - Page requested at: Sun, 26 Jan 2020 08:12:52 +0000

jjergus:
Fixes #790

The GitHub hhvm-autoload page has the installation instructions, but it seems reasonable to also have a short version here like #790 suggested.

I'm not sure what the best `composer` command syntax is since it depends on how you installed it, I used `php /path/to/composer.phar` since it seemed like the most descriptive and generic option. The other examples below just use `composer` which I think is fine, since by then the reader will have figured out how to invoke it correctly.

### DIFF
--- a/guides/hack/01-getting-started/03-starting-a-real-project.md
+++ b/guides/hack/01-getting-started/03-starting-a-real-project.md
@@ -4,7 +4,7 @@ Real projects generally aren't a single file in isolation; they tend to have
 dependencies such as the [Hack Standard Library], and various optional tools.
 
 A good starting point is to:
-- [install Composer]
+- [install Composer](https://getcomposer.org/doc/00-intro.md#installation-linux-unix-macos)
 - create an empty `.hhconfig` file
 - create `src/` and `tests/` subdirectories
 - configure autoloading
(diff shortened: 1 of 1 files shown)


guides/hack/01-getting-started/03-starting-a-real-project.md
```diff
@@ -4,7 +4,7 @@ Real projects generally aren't a single file in isolation; they tend to have
 dependencies such as the [Hack Standard Library], and various optional tools.
 
 A good starting point is to:
-- [install Composer]
+- [install Composer](https://getcomposer.org/doc/00-intro.md#installation-linux-unix-macos)
```
fredemmott: This was/is a link reference; either update the reference definition at the bottom of the file and remove the inline URI here, or delete the reference definition at the bottom of the file

fredemmott:
Not for this PR specifically, but in general it would probably be better for readability if we moved the link reference definitions to below the paragraph they're used instead of the bottom of the file.

jjergus:
> Not for this PR specifically, but in general it would probably be better for readability if we moved the link reference definitions to below the paragraph they're used instead of the bottom of the file.

I don't really have an opinion but isn't it kind of the point of references that they're collected all in one place? If it's better to have the URLs closer to the links we might as well just put them inline, or not?

fredemmott:
IMO inline links make the actual text harder to read, especially when there's a bunch of them in one list/paragraph

PR #803, opened 2020-02-10
remove unnecessary (I think?) CSS flexbox

### LINKED ISSUE #797: Content doesn't fit on screen

# Where is the problem?

Every page (almost)

# What is the problem?

The right pane doesn't fit on screen:

Screenshot 2020-02-05 21 09 57

Some CSS issue with the docs. I do not have the browser zoomed. I'm using Chrome.

 - Build ID: HHVM=HHVM-4.43.0:HSL=v4.40.0:2020-02-05T23:29:58+0000:f852ef8b5561843e9311cb502ae9baf788d6d841
 - Page requested: /hack/getting-started/starting-a-real-project
 - Page requested at: Thu, 06 Feb 2020 05:09:42 +0000

jjergus:
Fixes #797

I don't see any reason for this flexbox, it only wraps 1 element (maybe there were more in the past...?)

Removing it avoids the issues from #797, I tested with all the different browser widths which we support, in Firefox and Chromium.

### DIFF
--- a/sass/layout.scss
+++ b/sass/layout.scss
@@ -5,9 +5,6 @@
 }
 
 .flexWrapper {
-  align-items: flex-start;
-  display: flex;
-  flex-flow: row nowrap;
   padding-bottom: 50px;
   padding-top: 50px;
 }
@@ -25,8 +22,6 @@
 }
 
 .mainWrapper {
-  flex: 1;
-  order: 2;
   max-width: $default-max-width;
 }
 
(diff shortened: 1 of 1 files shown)


fredemmott:
> Removing it avoids the issues from #797, I tested with all the different browser widths which we support, in Firefox and Chromium.

Did you test with the mobile device emulation features?

jjergus:
Good idea, I also tested on my phone, looks all good.

Wilfred:
Looks good to me! Thanks for fixing this :)

PR #804, opened 2020-02-17
doc for Newable attribute + cleanup of Enforceable

francesco-zappa-nardelli:


guides/hack/20-attributes/07-predefined-attributes.md
```diff
@@ -206,6 +208,17 @@ mock class can exist.
...
 
+## __Newable
+
+This attribute is used to annotate reified type parameters to ensure that they are only instantiated with classes on which `new` can be safely called.  A common pattern is:
+```Hack
+function f<<<__Newable>> reify T as A>(): T {
+  return new T();
+}
+```
+where the class `A` is either final or annotated with `__ConstantConstructor`.  As a result the function `f` can only be applied to a _non-abstract_ class `C`.  The `as A` constraint ensures that the interface of the constructor of `C` is uniquely determined by the interface of the constructor of class `A`.
```
lexidor: I believe you meant `__ConsistentConstruct` and not `__ContantConstructor`.

francesco-zappa-nardelli: Yes, thanks.  Will fix in a second.

fredemmott: `Construct`, not `Constructor`

francesco-zappa-nardelli: Silly me.  Should have double checked the exact keyword, thanks.

guides/hack/20-attributes/07-predefined-attributes.md
```diff
@@ -206,6 +208,17 @@ mock class can exist.
...
 Mock classes *cannot* extend types `vec`, `dict`, and `keyset`, or the Hack legacy types `Vector`, `Map`, and `Set`.
 
+## __Newable
+
+This attribute is used to annotate reified type parameters to ensure that they are only instantiated with classes on which `new` can be safely called.  A common pattern is:
+```Hack
+function f<<<__Newable>> reify T as A>(): T {
+  return new T();
+}
+```
```
fredemmott: Can you:
- split this example out to a separate file like the others? This will alert us if/when this documentation becomes outdated
- add an example with parameters
- add an example where the attribute is not present (with error)
- add an example where the attribtue is present, but the class is not final or marked `__ConsistentConstruct`

Given the structure of the text, it might be best for the counter-examples to be in a separate file below the next paragraph

guides/hack/20-attributes/07-predefined-attributes.md
```diff
@@ -206,6 +208,17 @@ mock class can exist.
...
+where the class `A` is either final or annotated with `__ConsistentConstruct`.  As a result the function `f` can only be applied to a _non-abstract_ class `C`.  The `as A` constraint ensures that the interface of the constructor of `C` is uniquely determined by the interface of the constructor of class `A`. 
+See [Generics: Reified Generics](../generics/reified-generics.md) (Creating an instance of a class with `new`) for a complete example of its use.
```
fredemmott: While avoiding duplication is generally good, I feel it's bad for usability here.

guides/hack/20-attributes/07-predefined-attributes.md
francesco-zappa-nardelli: Hi Fred, I have split the example out to a separate file, added a more realistic example, added a sentence about errors, and revised the overall text.  I believe that adding examples about errors would not add much to the text.  Let me know what you think!

fredemmott:
Thanks so much!

PR #805, opened 2020-02-18
Refactor guide pagination to not use PHP arrays

fredemmott:
First commit: PHP array to darray with hhast
Second commit: manual refactor


fredemmott:
Also, this was the only usage of PHP arrays in the project

fredemmott:
Yep, that was needed for last weeks' updates in c5c62ecb0f5862d371b16333a60e473dce00f5fb

PR #810, opened 2020-02-21
Update Server Documentation To Include More Information About server.ini

mike2151:
I removed a broken link and added the location of server.ini in MacOS.


mike2151:
I just signed the Contributor License Agreement. So the above message from facebook-github-bot  is no longer valid.

Wilfred:
Thanks for the PR!

PR #817, opened 2020-02-25
MacOS Path Added For Getting Started: Configure HHVM

mike2151:
Added the location for php.ini in MacOS for the Getting Started Page


Wilfred:
👍

PR #821, opened 2020-03-02
Move darray-varray-runtime-options

lexidor:
I placed this article in `build-in-types/` instead of `types/`.
This way the file is in the same folder as the `arrays.md` article.

I also added a link from the `arrays.md` article to the runtime options guide.
Previously, the link was one way (from the runtime options to the arrays article).

Sorry for having to make this change after the previous PR had already been accepted.
I genuinely didn't notice this before.


Wilfred:
LGTM. Thanks for adding a link to the options page too, that definitely helps :)

PR #823, opened 2020-03-04
Document common nomenclature for containers

lexidor:
I wrote this to aid in the communication while migrating code.


jjergus:
Looks good to me, thanks!

PR #827, opened 2020-03-07
Update list() docs

### LINKED ISSUE #812: improve list(...) documentation
https://docs.hhvm.com/hack/expressions-and-operators/list

- The first examples is using `vec`, I think this is no longer allowed (should be a Hack error)? Even if not, it's definitely not recommended and shouldn't be the first example. The tuple examples are good and should be emphasized.
- Document `foreach ($arr as list(...))`, both here and in https://docs.hhvm.com/hack/statements/foreach (one can link to the other).

lexidor:
closes #812


guides/hack/04-statements/29-foreach.md
```diff
@@ -1,6 +1,6 @@
 The `foreach` statement iterates over the set of elements in a given collection, starting at the beginning, executing a single statement
-each iteration. On each iteration, the value of the current element is assigned to the corresponding variable, as specified.  The loop body
-is executed zero or more times.  For example:
+each iteration. On each iteration, the value of the current element is assigned to the corresponding variable, as specified. The loop body
```
lexidor: Sorry for formatting changes like this.
vscode does that for me when saving the file.

guides/hack/03-expressions-and-operators/13-list.md
```diff
@@ -1,27 +1,30 @@
-This intrinsic function assigns zero or more elements of its source to the corresponding target variable(s). It returns a copy of the source.  For example:
+`list()` is special syntax for unpacking tuples. It looks like a function, but it isn't one. It can be used in positions where you would assign into.
```
fredemmott: s/where/that/

PR #832, opened 2020-03-09
Give a fitting error message on ratelimit

lexidor:
Previously, you would have gotten this error instead:
`GitHub project facebook/hhvm has too many tags to process, giving up.`


lexidor:
### How to test

bin/update-versions (should update to hhvm 4.48 (ignore typechecker errors and single test failure)).

Re run until you see the ratelimit exception.

PR #834, opened 2020-03-18
Configuring TravisCI: Add src/ argument to hhast-lint 

mike2151:
Just running `vendor/bin/hhast-lint` generates the following output: 

> You must either specify PATH arguments, or provide a configurationfile.

and fails the TravisCI build. By adding `src/` as an argument, this message goes away and the build can then succeed.

### DIFF
--- a/guides/hack/01-getting-started/03-starting-a-real-project.md
+++ b/guides/hack/01-getting-started/03-starting-a-real-project.md
@@ -366,7 +366,7 @@ composer install
 hh_client
 vendor/bin/hacktest tests/
 if !(hhvm --version | grep -q -- -dev); then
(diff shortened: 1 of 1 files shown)


lexidor:
Hi, we'd generally recommend using a configuration file instead.
All oss hhvm projects have standardized on that method.
You can add a `hhast-lint.json` to the root of your project and not provide any arguments to `hhast-lint`.
This allows each project to choose what directories to lint and what linters to exclude.
The `hhast-lint` _binary_ will traverse back out of your vendor folder until it finds a configuration file.

mike2151:
Makes sense. Thanks!

PR #836, opened 2020-03-20
support @guide

### LINKED ISSUE #830: AsyncGenerator documentation claims a guide is referenced, it isn't.
" PLEASE READ THE GUIDE REFERENCED IN THIS API DOCUMENTATION FOR MORE INFORMATION"

No guide is referenced.

 - Build ID: HHVM=HHVM-4.44.0:HSL=v4.40.0:2020-03-02T21:51:17+0000:46c873cd1406ff85f61a48da737237395c587cd0
 - Page requested: /hack/reference/class/HH.AsyncGenerator/
 - Page requested at: Mon, 09 Mar 2020 16:50:24 +0000
 - Controller: APIClassPageController

jjergus:
Fixes #830

### DIFF
--- a/src/hh-apidoc-extensions/PageSections/Guides.php
+++ b/src/hh-apidoc-extensions/PageSections/Guides.php
@@ -1,0 +1,41 @@
+<?hh // strict
+/*
+ *  Copyright (c) 2004-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+namespace HHVM\UserDocumentation\HHAPIDocExt\PageSections;
+
+use type Facebook\HHAPIDoc\PageSections\PageSection;
+use type HHVM\UserDocumentation\GuidesNavData;
+use namespace HH\Lib\{C, Str};
+
+final class Guides extends PageSection {
+
+  <<__Override>>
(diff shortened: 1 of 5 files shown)


src/hh-apidoc-extensions/PageSections/Guides.php
```diff
@@ -0,0 +1,40 @@
...
+      return null;
+    }
+
+    $markdown = '## Guide';
+    if (C\count($guides) > 1) {
+      $markdown .= 's';
+    }
+    $markdown .= "\n\n";
+
+    foreach ($guides as $path) {
```
fredemmott: Can you add tests for this, covering:
- known-bad `@guide` tags
- `@guide tags` that need a redirect

jjergus: The current version leaves all the responsibility on whoever wrote the docblock, we literally just use the @guide string as the URL, so there's not much to test (probably the only meaningful test is that the @guide tags are not ignored like before, I'll add that).

The advantage is that we get redirects etc. for free, but the downside is we can't detect invalid guide URLs or other issues (I noticed on some pages like Vector there are multiple @guide links that redirect to the same place).

I wouldn't be opposed to doing something more fancy, but it gets a little tricky because the API documentation is generated fairly early during the build, before the routing map etc.

fredemmott: Ah - but the existing "bad internal link" detection should pick those up after the build?

fredemmott: s/detection/unit test/

PR #837, opened 2020-03-20
Document check_array_key_cast

lexidor:
Documenting another option from the list.

### DIFF
--- a/guides/hack/11-built-in-types/56-darray-varray-runtime-options-examples/hack_arr_compat_check_array_key_cast.php
+++ b/guides/hack/11-built-in-types/56-darray-varray-runtime-options-examples/hack_arr_compat_check_array_key_cast.php
@@ -1,0 +1,29 @@
+<?hh // strict
+
+namespace Hack\UserDocumentation\Types\VarrayDarrayRuntimeOptions\HackArrCompatCheckVarrayPromote;
+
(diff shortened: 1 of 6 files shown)


guides/hack/11-built-in-types/56-darray-varray-runtime-options-examples/hack_arr_compat_check_array_key_cast.php
```diff
@@ -0,0 +1,29 @@
...
+  /*HH_IGNORE_ERROR[4324]*/
+  $varray[true] = 'A bool?!?';
+  /*HH_IGNORE_ERROR[4324]*/
+  $varray[null] = 'null?!?';
+
+  $darray = darray[];
+
+  $darray[1.1] = 'A float?!?';
+  $darray[true] = 'A bool?!?';
+  $darray[null] = 'null?!?';
```
lexidor, line 27: I find it strange that it is not a typechecker error to `$darray[1.1]` or `$dict[1.1]`.
Varrays, Vectors and vecs **do** have an typechecker error.

Wilfred: There's a typechecker flag `disallow_invalid_arraykey` that does this.

PR #839, opened 2020-03-23
Update Docker example to work in hhvm 4.2+

lexidor:
The default mode is strict, which does not allow toplevel code.
We could also fix this by adding a `// partial` comment, but those are slated to be removed.


Wilfred:
Thanks :)

PR #840, opened 2020-03-23
redirect on extra trailing slash

jjergus:
We support automatic redirect if there is a missing trailing slash, but not if there's an extra one, e.g. `/hack/generics/introduction/`

We have a few `@guide` links that are invalid because of this.

### DIFF
--- a/src/site/HHVMDocumentationSite.php
+++ b/src/site/HHVMDocumentationSite.php
@@ -14,7 +14,7 @@ use type Facebook\Experimental\Http\Message\{
   ResponseInterface,
   ServerRequestInterface,
 };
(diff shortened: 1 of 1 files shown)


jjergus:
Ooh this is problematic because the `/{product}/{guide}/{page}` pattern now incorrectly matches `/{product}/reference/{type}` and therefore prevents the correct redirect to `/{product}/reference/{type}/`

We could add a trailing slash to the `/{product}/{guide}/{page}` pattern but that would probably cause a lot of existing links to become redirects, which we probably don't want.

I guess I'll just work around this for the `@guide` links for now.

PR #843, opened 2020-03-27
When building on mac, tell cmake to use homebrew's preferred SDK

fredemmott:
- Homebrew always uses XCode command line tools if installed
- CMake always uses the system/user's configured SDK, which if both are
  installed, is usually the non-CLI XCode
- Right now they're not compatible :/


fredemmott:
Merging this as it's now required, and tested on azure

PR #846, opened 2020-04-08
Fixed Push Function in Stack Class Example 

mike2151:
The current code does not compile when you call push


Wilfred:
Thanks.

Strictly speaking it throws an exception, it does compile correctly. You're absolutely right that the code didn't work though :)

PR #847, opened 2020-04-08
update to 4.52

fredemmott:


fredemmott:
I set the hackttp requirement to * as it will still need to be compatible with the hack-router and hack-router-codegen versions we’re using - and the version of the interfaces they depend on.

fredemmott:
https://github.com/hhvm/user-documentation/commit/d96130cc77b2ef2fc08fd66697c635a9d1aa5519

PR #853, opened 2020-04-16
__Memoize memoizes awaitables, this affects exceptions

fredemmott:
Screen Shot 2020-04-16 at 9 23 39 AM

### DIFF
--- a/guides/hack/20-attributes/07-predefined-attributes.md
+++ b/guides/hack/20-attributes/07-predefined-attributes.md
@@ -193,7 +193,7 @@ instead of a static property with `__LateInit`.
 
 The presence of this attribute causes the designated method to automatically cache each value it looks up and returns, so future calls with
 the same parameters can be retrieved more efficiently. The set of parameters is hashed into a single hash key, so changing the type, number,
-and/or order of the parameters can change that key.
+and/or order of the parameters can change that key. Functions with variadic parameters can not be memoized.
 
 This attribute can be applied to functions and static or instance methods; it has no attribute values.  Consider the following example:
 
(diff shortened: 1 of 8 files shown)


guides/hack/20-attributes/07-predefined-attributes.md
```diff
@@ -191,6 +191,26 @@ The interface type `IMemoizeParam` assists with memorizing objects passed to asy
 - If an exception is thrown, this is not memoized.
```
jjergus, line 191: Probably remove this, since it's been replaced by the new section.

This leaves the "Limitations" section with only 1 item, so maybe integrate that into the text above (there's a paragraph about what kinds of argumentsd are allowed where it would probably fit)

guides/hack/20-attributes/07-predefined-attributes-examples/memoize-async-throw.hack
```diff
@@ -0,0 +1,22 @@
...
+  public async function doStuff(): Awaitable<void> {
+    $this->count += 1;
+    throw new \Exception('Hello '.$this->count);
+  }
+}
+
+<<__EntryPoint>>
+async function main(): Awaitable<void> {
+  $x = new CountThrows();
+  for($i = 0; $i < 2; ++$i) {
```
Wilfred, line 15: Perhaps run these code samples through `hackfmt`?

```suggestion
  for ($i = 0; $i < 2; ++$i) {
```

guides/hack/20-attributes/07-predefined-attributes.md
```diff
@@ -191,6 +191,26 @@ The interface type `IMemoizeParam` assists with memorizing objects passed to asy
...
 - Functions with variadic parameters can not be memoized
```
Wilfred, line 192: Huh, this a typehole. I've filed T65637246.

guides/hack/20-attributes/07-predefined-attributes.md
```diff
@@ -191,6 +191,26 @@ The interface type `IMemoizeParam` assists with memorizing objects passed to asy
...
+example:
+
+@@ predefined-attributes-examples/memoize-throw.hack @@
+
+However, if the the memoized function is `async`, the function does not
+technically throw - it returns an Awaitable that throws when resolved. This
+means that if the Awaitable is resolved multiple times, the same exception
+will be thrown every time; as the Awaitable is cached for `__Memoize`'d
+async functions, this results in the same exception being thrown every time,
+shown by the counter remaining 0 in this example:
```
Wilfred, line 210: Could you break up this sentence?

> If the Awaitable is resolved multiple times, the same exception
will be thrown every time. For `__Memoize`'d async function, this results
> in the same exception being thrown each time.
>
> The counter remains 0 in this example:

fredemmott:
> When I read this the first time I was a bit confused that the section is called "Awaitables and Exceptions", but the 1st example has nothing to do with Awaitables. I'd suggest splitting it into 2 sections, something like:

Hmm, I'll try to think of a better way to structure this: the point I'm trying to clarify is how these two behaviors interact, which can be surprising

jjergus:
> Hmm, I'll try to think of a better way to structure this: the point I'm trying to clarify is how these two behaviors interact, which can be surprising

I guess my point was that we should probably make sure each behavior is clearly explained on its own first, then move on to how they interact.

It seems to be effectively 3 separate concepts:

1. __Memoize + exceptions
2. __Memoize + async
3. __Memoize + async + exceptions

I'm not sure if we need a separate section with its own title for each of those, but at least 2 separate sections probably make sense?

fredemmott:
Resurrecting this... will rebase + address feedback shortly

fredemmott:
- suggested restructurings (latest commit)
- made the inline-example extractor create async main functions - needed here, but I don't think it should hurt anywhere, and I'd rather not add an `if(/^await /)` or similar

guides/hack/20-attributes/07-predefined-attributes.md
```diff
@@ -221,10 +221,77 @@ The types of the parameters are restricted to the following: `null`, `bool`, `in
...
+    \var_dump($e->getTrace()[2] as shape('function' => string, ...)['function']);
+  }
+}
+```
+
 ### Limitations
 
-- If an exception is thrown, this is not memoized.
-- Functions with variadic parameters can not be memoized
+Functions with variadic parameters can not be memoized.
```
jjergus, line 294: We probably don't need a whole section for this one item -- I'd just move this at the end of the paragraph that describes what parameter types are accepted (line 220).

Alternatively, if you feel a separate section is better, maybe create a "Parameters" section and move all of that there.

jjergus (approved):

jjergus:
You might have to rebase & run build.php to update any of my newly migrated examples to async main functions.

jjergus:
I did the async main conversion in https://github.com/hhvm/user-documentation/commit/d07191ec1c7cace726a2ecc57c5f4739dcc21621 (also needed to update some expectf files that contained stack traces) so you just need to rebase and rebuild your stuff now.

PR #855, opened 2020-04-17
Separate out test commands in travis.yml

Wilfred:
This makes it easier to see which command is producing which output,
and which is failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,10 @@ install:
 - docker pull hhvm/hhvm-proxygen:$HHVM_VERSION
 script:
 - docker build -t hhvm/user-documentation:scratch -f .deploy/built-site.Dockerfile .
-- docker run --rm -w /var/www hhvm/user-documentation:scratch ./.travis-test.sh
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -15,7 +15,10 @@ install:
 - docker pull hhvm/hhvm-proxygen:$HHVM_VERSION
 script:
 - docker build -t hhvm/user-documentation:scratch -f .deploy/built-site.Dockerfile .
-- docker run --rm -w /var/www hhvm/user-documentation:scratch ./.travis-test.sh
+- docker run --rm -w /var/www hhvm/user-documentation:scratch hh_server --check .
```
fredemmott, line 18: Could you make the first be `docker start`, then replace these all with `docker exec`, to avoid the creation/teardown each time?

Wilfred: Creation/teardown seems to be very quick (the fastest step is 1.2 seconds including doing the work), and this prevents filesystem state affecting the different test steps.

PR #856, opened 2020-04-22
Fix typo: $colorVect -> $colorVec

TysonAndre:
Elsewhere in this document, the variable name is $colorVec: `$colorsVec has an implicit key type of int`


Wilfred:
Thanks.

PR #857, opened 2020-04-22
Fix miscellaneous typos detected by codespell

TysonAndre:
"inbetween" should either be "in-between" or "between"

https://en.wiktionary.org/wiki/targetting is no longer used in British


lexidor:
Whoops, some of these typos are mine.
👍 I should probably install something similar to codespell.

TysonAndre:
https://github.com/tysonandre/phantypocheck is what I'm using, along with https://github.com/codespell-project/codespell. The only benefit is for camelCase or snake_case, until codespell supports that.

> (using heuristics to guess individual words of camelCase or snake_case identifiers).

There's also LanguageTool, which various IDEs are integrating in recent releases

Wilfred:
Wow, that's caught a ton of typos! Thanks again :)

PR #858, opened 2020-04-23
Rewrite of 27-using-a-trait

francesco-zappa-nardelli:
This is an attempt at rewriting 27-using-a-trait.  Goal is to describe what a trait is via simple examples, and to explain naming conflict resolution.  I believe this is an improvement over the previous text.  Trait requirements will be discussed in a different page.


francesco-zappa-nardelli:
Travis is right, there is a "public" missing and will fix.  However I followed the testing instruction of the on-demand server, that is:
```
$ hhvm bin/build.php
$ ./vendor/bin/hh-autoload
$ ./vendor/bin/hacktest tests/
``` 
and the type error was undetected.  Is there one more command that should be executed to test the examples of the doc?

jjergus:
> Is there one more command that should be executed to test the examples of the doc?

That was probably a Hack (typechecker) error, so probably `hh_client` (or just `hh`).

Adding @Wilfred as a reviewer since he's been working on the class section most recently...

PR #860, opened 2020-05-09
Run update-versions and fix two tests

lexidor:
Relies on a single change in usox/hackttp, readRawBlocking() -> read()
Once that update is in, a simple `composer update` will fix the composer.lock file.
https://github.com/usox/hackttp/pull/28
is closed by #861


fredemmott:
Switching to https://github.com/nuxed/http-message + https://github.com/nuxed/http-emitter might be a viable alternative way to unblock this.

PR #861, opened 2020-05-12
Migrate to using Nuxed HTTP

azjezz:
this is an experiment I did to see how can I improve Nuxed HTTP-message in order to make it easier to use.

results:

- added `Response\{text|html|xml}_file` response factories
- added `Response\{has|get|with|without}_cache_control_directive` helpers

benefits: 

- less code to maintain for maintainers of HHVM documentation.

---

In the end. it's up to the HHVM on whether to merge this PR or not.

but *NOTE* Nuxed components require HHVM ^4.56, I am not going to release a new version with support for older HHVM versions, so #860 needs to be merged first :)


public/index.php
```diff
@@ -2,40 +2,18 @@
...
-  );
+  $app = new HHVMDocumentationSite();
 
-  await HHVMDocumentationSite::respondToAsync($request);
-  exit(0);
+  $request = await Http\Message\ServerRequest::capture();
 
+  $response = await $app->handle($request);
+
+  await Http\Emitter\emit($response);
```
azjezz, line 18: it is also possible to do:

```hack
$emitter = new Http\Emitter\Emitter();

await $emitter->emit($response);
```

but I don't think we really need an emitter instance in this case.

src/site/Router.php
```diff
@@ -9,5 +9,18 @@
...
+  final public function match(
+    Message\IServerRequest $request,
+  ): (classname<\RoutableController>, ImmMap<string, string>) {
+    $method = HackRouter\HttpMethod::coerce($request->getMethod());
+    if ($method === null) {
+      throw new HackRouter\MethodNotAllowedException(keyset[]);
+    }
+
+    return $this->routeMethodAndPath($method, $request->getUri()->getPath());
+  }
```
azjezz, line 25: hack router offers a similar method for facebook request/response interfaces, but it doesn't support nuxed.

src/site/controllers/StaticResourcesController.php
```diff
@@ -54,24 +55,23 @@ protected static function getExtraParametersSpec(
...
+        |> Response\with_cache_control_directive($$, 'no-cache')
+        |> Response\with_cache_control_directive($$, 'no-store');
     } else {
-      $response = $response->withAddedHeader(
-        'Cache-Control',
-        vec['max-age=31556926'], // 1 year
-      )
-        ->withAddedHeader('ETag', vec['"'.$checksum.'"']);
+      $response = Response\with_max_age($response, 31556926) // 1 year
+        |> Response\with_etag($$, $checksum);
```
azjezz, line 72: no need to do `'"'.$checksum.'"'`, it will be done automatically if the checksum is not wrapped within `"`

src/site/controllers/WebController.php
```diff
@@ -34,14 +32,11 @@ abstract class WebController {
...
+    private Message\IServerRequest $request,
   ) {
     $combined_params = new Map($parameters);
     foreach ($request->getQueryParams() as $key => $value) {
-      if (is_array($value)) {
-        continue;
-      }
-      $combined_params[(string)$key] = (string)$value;
+      $combined_params[$key] = $value;
     }
```
azjezz, line 40: `$request->getQueryParams()` in Nuxed returns `KeyedContainer<string, string>`, while previously it returned `KeyedContainer<arraykey, mixed>`

Nuxed normalizes all inputs using : github.com/azjezz/http-normalizer

this is done in order to avoid the use of mixed and you can be sure that you are either getting `string` or nothing, while previously it can be `string|array<arraykey, string|array<arraykey, string.......` ( `string|tree<arraykey, string>` )

src/site/controllers/WebController.php
```diff
@@ -89,15 +84,13 @@ protected static function getExtraParametersSpec(
     return shape('required' => ImmVector {}, 'optional' => ImmVector {});
   }
 
-  abstract public function getResponseAsync(
-    ResponseInterface $response,
-  ): Awaitable<ResponseInterface>;
+  abstract public function getResponseAsync(): Awaitable<Message\IResponse>;
```
azjezz, line 87: we don't pass a response to the controller anymore as we don't actually need to, previously it was required as the response body is only writable, in nuxed case, its both readable and writeable, meaning the controller can write anything, and we can read from it without knowing where the body is located ( in case its a file ). we know its readable, we can read from it at the end ( done int the emitter )

src/site/controllers/WebController.php
```diff
@@ -89,15 +84,13 @@ protected static function getExtraParametersSpec(
...
 
   final protected function getRequestedPath(): string {
     return $this->request->getUri()->getPath();
   }
 
-  final protected function getRequestedHost(): string {
+  final protected function getRequestedHost(): ?string {
```
azjezz, line 93: Nuxed's `Uri::getHost()` returns `?string` as the host can be missing sometimes ( when its only a path )

fredemmott:
> so #860 needs to be merged first :)

Heh, merging this would unblock #860 

Probably best for us to get these two as good as possible, then you + @lexidor  look at a combined PR, with one commit for each current PR so we get a combined test run

src/site/HHVMDocumentationSite.php
```diff
@@ -10,84 +10,54 @@
  */
 
 use namespace Facebook\HackRouter;
-use type Facebook\Experimental\Http\Message\{
-  ResponseInterface,
-  ServerRequestInterface,
-};
-use namespace HH\Lib\{File, IO, Math};
+use namespace Nuxed\Contract\Http\{Message, Server};
```
fredemmott, line 13: Is this required? Would extremely strongly prefer to stick to the existing interfaces.

azjezz: @nuxed interfaces don't extend current facebook interfaces, as i mentioned in slack, there's quite few issues with facebook interfaces, that's why @Nuxed doesn't use them.

azjezz: P.S `Nuxed\Contract` only contains abstractions ( interfaces, traits, enums ..etc), not the implementation itself.

facebook interfaces:

- has only 1 current stable implementation
- uses `KeyedContainer<arraykey, mixed>` for inputs
- uses write only handles for responses
- not standard way to emit responses

nuxed contracts:

- has only 1 current stable implementation
- uses `KeyedContainer<string, string>` for inputs
- supports cookies
- supports session/falsh
- the available implementation offers tons of helper methods to deal with cache, file responses ... etc

fredemmott: Yeah ,that mostly feels appropriate for a framework :)

> read/write responses

I'd expect most implementations to provide a read/write (or a pipe) by default, however forcing read/write in the type system bans 'streaming' responses

Almost at the point in HSL IO where updating the interfaces is next - will discuss further there, but I'm leaning towards 'responses should return generic `<T>`', and it's up to the framework - or emitter - to decide what's acceptable

For example, at FB, a 'response body' is an XHP element - for pages, this boils down to the doctype one, for other stuff, it might be a `<myframework:file-response />` which deals with streaming responses - i.e. instead of passing around the place to write the response, make it so the response actually gets returned

azjezz: > forcing read/write in the type system bans 'streaming' responses

how so? currently nuxed/http-emitter supports streaming response :)

azjezz: > For example, at FB, a 'response body' is an XHP element - for pages, this boils down to the doctype one, for other stuff, it might be a <myframework:file-response />

this somehow doesn't feel right to me, XHP is always associated with XML documents or templating in general for me, returning responses that contain only data ( e.g `json` ), by using XHP, seems wrong, even if it works.

fredemmott: If it's generic in the interfaces, up to the framework/site - e.g. `XHPResponse implements MyFrameworkResponse`, `JSONResponse implements MyFrameworkResponse`

lexidor:
> > so #860 needs to be merged first :)
> 
> Heh, merging this would unblock #860
> 
> Probably best for us to get these two as good as possible, then you + @lexidor look at a combined PR, with one commit for each current PR so we get a combined test run

Can @azjezz checkout `lexidor:support-hhvm-4.56`?
If yes, he could merge my changes to this branch, since we don't touch the same files.

azjezz:
@lexidor if you could squash your commits, I can do the same and then cherry-pick your commits into my fork and leave 1 PR open.

lexidor:
> @lexidor if you could squash your commits, I can do the same and then cherry-pick your commits into my fork and leave 1 PR open.

https://github.com/hhvm/user-documentation/pull/860/commits
There is only one commit (luckily), so is there a need to squash?

azjezz:
Ah no :) thanks

lexidor:
By the way, this is a fun PR.
However, a lot of changes go over my head.
I am interested to see the final diff.

src/site/controllers/NonRoutableWebPageController.php
```diff
@@ -148,13 +144,11 @@ protected function getGithubIssueBody(): string {
...
     $html = await $xhp->asyncToString();
 
-    await $response->getBody()->writeAsync($html);
-    $response = $response
-      ->withStatus($this->getStatusCode())
-      ->withHeader(
-        'Cache-Control',
-        vec['max-age=60'],
-      ); // enough for pre-fetching :)
+    $response = Response\html($html)
```
lexidor, line 147: `Response\xhp($xhp)` would be nice, since `$xhp->asyncToString()` returns an `HTMLString` and not a "normal" string.

src/site/HHVMDocumentationSite.php
azjezz: in that case, we would need to send the response from the response implementation, as it might differ, that's not something I want to do.

as I said, read+write doesn't bring any issues, its also possible to pass a handle that implements read+write but throws on write, writing to the response outside the handler ( called controller here ) is possible in a middleware, to modify the response.

src/site/controllers/WebController.php
```diff
@@ -34,14 +32,11 @@ abstract class WebController {
...
-    private ServerRequestInterface $request,
+    private Message\IServerRequest $request,
   ) {
     $combined_params = new Map($parameters);
     foreach ($request->getQueryParams() as $key => $value) {
-      if (is_array($value)) {
-        continue;
-      }
-      $combined_params[(string)$key] = (string)$value;
+      $combined_params[$key] = $value;
```
lexidor, line 39: Would this now include the normalized array strings in the arguments?
Previously, they were ignored.

src/site/http_exceptions.php
```diff
@@ -9,38 +9,35 @@
...
   <<__Override>>
   public async function getResponseAsync(
-    ServerRequestInterface $_,
-    ResponseInterface $response,
-  ): Awaitable<ResponseInterface> {
-    return $response->withStatus(405);
+    Http\Message\IServerRequest $_request,
+  ): Awaitable<Http\Message\IResponse> {
+    return Response\empty()
+      ->withStatus(Http\Message\StatusCode::MethodNotAllowed);
```
lexidor, line 40: Nice, no more magic ints! :stuck_out_tongue_closed_eyes:

src/site/controllers/WebController.php
azjezz: yes. but that's not an issue in this case.

src/site/controllers/NonRoutableWebPageController.php
azjezz: `Response\xhp` then would return an Awaitable, which seems inconsistent with other response factories such as `html`, `xml`, and `text`.

also, the point of `html` and other is that they add content type http header, with `xhp` we would be unable to do so, since we don't know the content type, it might be html, xml, or something else.

src/site/HHVMDocumentationSite.php
fredemmott: Read/Write/Seekable: bans directly writing to request_output without buffering.

Always-throw implementations of part of it are removing the safety that the types are meant to provide.

Also, messaged you some more context on why I strongly want to continue using the interfaces.

azjezz:
:heavy_check_mark: green.

fredemmott:
@azjezz thanks again for this - very open to this in the future, as long as it does not couple us so tightly, and is primarily based on the interfaces.

I'm aware the interfaces need work; I'm expecting the last relevant breaking changes to HSL IO next week, then the interfaces will be next up in priority, and I'm hoping nuxed will be able to use them again after that.

PR #866, opened 2020-05-20
Update to 4.58

fredemmott:


fredemmott:
Not yet ready: fixes are needed in xhp-lib for new varray/darray notices

guides/hhvm/02-basic-usage/02-command-line.md
```diff
@@ -1,13 +1,13 @@
 In command-line (cli) mode, you run the `hhvm` binary from the command-line, execute the script and then exit HHVM immediately when the script completes.
 
-Here is an example of how to run a script in HHVM cli mode. Take the following PHP script:
+Here is an example of how to run a script in HHVM cli mode. Take the following Hack script:
```
Wilfred, line 3: 👍

fredemmott:
getting rid of varray/darray notices require type-assert v4, but projects require v3 (removing vec_like_array/dict_like_array is a breaking change).

I'll try to get through those tomorrow.

As the dependencies are circular, pull requests can't be tested in isolation so aren't useful - I'll test them as a whole and publish 'failing' releases simultaneously, which will pass once the set is consistent.

fredemmott:
I think I got all the FB libraries; hackttp needs an update too.

fredemmott:
Now blocked by https://github.com/hhvm/hack-codegen/issues/130 - the amount of log spew makes TravisCI terminate the job

fredemmott:
Force-pushed rebase bringing this to 4.58; will land ,then look at bringing it up to 4.61

PR #871, opened 2020-05-28
Explain how exceptions behave in concurrent blocks

hurrymaplelad:
exceptions

### DIFF
--- a/guides/hack/15-asynchronous-operations/14-concurrent.md
+++ b/guides/hack/15-asynchronous-operations/14-concurrent.md
@@ -34,3 +34,39 @@ concurrent {
   $result[] = await z_async();
 }
 ```
+
+## Exceptions
+
+If any statement in a `concurrent` block throws, there are no guarantees about which (if any) awaited values were assigned or which exception will be propagated if more than one of them threw. For example if you have:
+
+```
+$x = 0;
+try {
+  concurrent {
+    $x = await async { return 1; };
+    await async { throw new Exception('foo'); }
+    await async { throw new Exception('bar'); }
+  }
+} catch (Exception $e) {
+  var_dump($x, $e->getMessage());
+}
+```
+Then it is explicitly undefined whether `var_dump` will see `$x === 0` or `$x === 1` and whether the message will be 'foo' or 'bar'.
+
+If you need granular exception handling, consider using nested try-cath blocks inside the concurrent block:
+
+```
+concurrent {
+  $x = await async {
+    try {
+      ...
+      return <success path>;
+    } catch (...) {
+      return <fallback>;
+    }
+  };
+  $y = await async {
+    // same here
+  };
+}
+```


lexidor:
I had no idea that concurrent blocks were so liberal about the state it leaves your awaitables in.

So, if I understand it correctly, if any one of my awaitables throws, while the another awaitable is waiting on some io (curl), the exception will fly out before my curl has finished. What happens to my curling awaitable? If my script catches the exception, echos some error info, and returns from the entrypoint, my curling awaitable never finished.

On a related note, do `Asio\va()` and friends offer the guarantee that all awaitables must finish or throw, before an exception can bubble out? If not, I've been writing my code all wrong and I should audit a lot of it.

guides/hack/15-asynchronous-operations/14-concurrent.md
```diff
@@ -34,3 +34,25 @@ concurrent {
   $result[] = await z_async();
 }
 ```
+
+## Exceptions
+
+If any statement in a `concurrent` block throws, evaluation of the other statements is undefined. They may have already run, they may never run, or they may still be executing. The concurrent block will throw an arbitrary single exception.
```
alexeyt, line 40: This is only partly correct. When a concurrent block has completed, all the await operations in it will have completed in some way (either by throwing or returning a value). If any of them threw, then there are no guarantees about which (if any) returned values were assigned and which exception will be propagated if more than one of them threw. For example if you have:
```
$x = 0;
try {
  concurrent {
    $x = await async { return 1; };
    await async { throw new Exception('foo'); }
    await async { throw new Exception('bar'); }
  }
} catch (Exception $e) {
  var_dump($x, $e->getMessage());
}
```
Then it is explicitly undefined whether var_dump will see `$x === 0` or `$x === 1` and whether the message will be 'foo' or 'bar'.

alexeyt:
> I had no idea that concurrent blocks were so liberal about the state it leaves your awaitables in.
> 
> So, if I understand it correctly, if any one of my awaitables throws, while the another awaitable is waiting on some io (curl), the exception will fly out before my curl has finished. What happens to my curling awaitable? If my script catches the exception, echos some error info, and returns from the entrypoint, my curling awaitable never finished.
> 
> On a related note, do `Asio\v()`, `AwaitAllHandle::fromVec()`, and friends offer the guarantee that all awaitables must finish or throw, before an exception can bubble out? If not, I've been writing my code all wrong and I should audit a lot of it.

concurrent (and other similar operations) require Awaitables to finish by either returning or throwing; I think the PR author misunderstood the extent of what is undefined. I commented with an example on the diff.

guides/hack/15-asynchronous-operations/14-concurrent.md
lexidor: So if the runtime decides to order the awaitables in the order 3,2,1, all awaitables will execute (so all side effects of all the awaitables happen). Even though exception foo happened after exception bar, the runtime could decide to propagate foo instead. The assignment to x may or may not happen.

Am I on the right track here?
Going with the following example:
```HACK
$result = vec[];
concurrent {
  $result[] = await x_async();
  $result[] = await y_async();
}
```

If no exception is thrown, result will be [x, y], if x throws, result may be [] or [y], if y throws result maybe [] or [x].

If this is the case, the heading above could use a *if no exceptions occur*. If this is well defined, because the mutated the same variable, then it can stay as is.

alexeyt: That's almost right. As it says above the code snippet in the doc we currently have, assignment of awaited results is guaranteed to happen in order, so we can't get a result that's [y] (because we would have had to check the result of x first and that would have propagated the exception from x), but $result may be any of [] (possible if either x or y or both throw) or [x] (possible if y throws), and will be [x, y] if no exception is thrown.

In any case, all the side-effects that all the Awaitables produce before they return or throw will happen, and they will all happen before any results are assigned by the concurrent block, but the order they happen relative to each other is undefined.

It is recommended that programmers not rely on the values of any variables assigned to in the concurrent block in the event that an exception is thrown out of it. The order of assignments is guaranteed, so it's possible to reason about them, but code will be more readable and maintainable if you don't.

hurrymaplelad:
Yep, I wan't sure how much was guaranteed, so I wrote up the most pessimistic version. 

@alexeyt the example is super helpful. I'll update my description and include your example.

PR #878, opened 2020-06-24
Update varray and darray documentation describing upcoming default behavior

viratyosin:
Update varray and darray documentation to indicate that versions after 4.62 will turn off and remove the implicit interoperability of the two
types.


Wilfred:
(y)

PR #888, opened 2020-07-15
update to HHVM 4.66

jjergus:
- Review commits individually (whichever commits you're interested in -- most are straightforward).
- Needs to wait for https://github.com/facebook/hhvm/pull/8728 or the deployment will fail again.


guides/hack/01-getting-started/05-input-and-output-examples/hsl-io-basics.hack
```diff
@@ -14,10 +14,12 @@ async function main_async(): Awaitable<void> {
   await $out->writeAsync($message);
 
   // copy to a temporary file, automatically closed at scope exit
-  await using ($f = File\temporary_file()) {
+  using ($tf = File\temporary_file()) {
+    $f = $tf->getHandle();
+
     await $f->writeAsync($message);
```
fredemmott, line 20: actually, these should generally be writeAllAsync nowadays; this can partially succeed

PR #892, opened 2020-07-27
test deployment in Travis cron runs

jjergus:
Any reason not to do this? (e.g. is deploying to EB expensive?)

I skipped the Docker push because I didn't want to deal with tags that may or may not be duplicate (also that step has never failed, and if it does it's likely a network error)...we still build the Docker image though, to test repo-mode build (or is that tested elsewhere already?).

### DIFF
--- a/.travis-deploy.sh
+++ b/.travis-deploy.sh
@@ -1,13 +1,16 @@
 #!/bin/bash
 set -e
 
-if [ "${TRAVIS_EVENT_TYPE}" == "cron" ]; then
-  exit 0
-fi
-
 DEPLOY_REV=$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


fredemmott:
What processes you aiming to test?

This shouldn't push to dockerhub - but without doing that, any EB actions will either fail, or be operating on an old version

jjergus:
> What processes you aiming to test?

- repo-mode has failed in the past due to HHVM regressions (the symlink stuff)
- the `eb` script has also failed for at least 2 different reasons recently

> This shouldn't push to dockerhub - but without doing that, any EB actions will either fail, or be operating on an old version

It skips the Docker push and uses `hhvm/user-documentation:latest` as the image name for the Docker build and the partial deployment. I'm not sure if the deployment uses the local one that was just built or pulls the official one but it shouldn't really matter, we only care about whether anything fails.

PR #896, opened 2020-08-01
target HHVM 4.68, fix a lot of top-level code

jjergus:
Review commits separately (whichever commits look interesting). Possibly interesting ones:

> fix examples for renamed methods toArray --> toDArray/toVArray

`Map::toArray` etc. all got renamed to `toVArray` or `toDArray`, this pretty much just renames the respective directories in api-examples/

I did a few fixes for top-level code etc. but these are irrelevant because we don't actually test anything in api-examples :/ I'll fix that later.

> auto_prepend_file=src/utils/_private/__init_autoload.php
> filter out __init_autoload() when rendering

We can't auto_prepend hh_autoload.php anymore because it has top-level code, so now we auto_prepend this instead, which unfortunately requires explicitly calling `\__init_autoload();` in every example that needs it, but we filter that out during rendering.

> add guides/ to hh_autoload.json

Any reason not to do this? Almost all examples are properly namespaced (we even have that `\main()` function declaration just to prevent any examples from redeclaring it).

This allows us to avoid a bunch of `require_once 'specific_file.php';` which at this point barely works and should probably be discouraged. I think `require_once` will be removed from the language completely soon-ish...?

> manually fix Car interface examples

As far as I can tell, in latest HHVM it is impossible to do anything like

```hack
require_once 'Bar.php';
class Foo extends Bar {}
// fails because of top-level code

<<__EntryPoint>>
function main(): void {
  require_once 'Bar.php';
}
class Foo extends Bar {}
// fails because the require_once is processed too late
```

so here I hacked around it.

> semi-automatically migrate the rest of guides/

I wrote an HHAST migration that moves all top-level statements into an EntryPoint function (existing or creates a new one), but I did a lot of manual fix-ups.

> split out XHP definitions

Same problem as the Car interface, but these could be solved a bit more elegantly (if you look at the first 3 files in the commit, the rest is all the same).


fredemmott:
> add guides/ to hh_autoload.json

This is fine, and generally killing require_once is also good :)

> manually fix Car interface examples

Related to the pseudomain stuff, HHVM's also recently got a little more liberal here:

```
class Baz extends Bar {}
class Bar extends Foo {}
class Foo {}

<<__EntryPoint>>
function main(): void {
  \var_dump(new Baz());
}
```

This would previously be an error on line 2 ('Unknown class Bar'), but is now fine.

PR #898, opened 2020-08-04
convert to xhp-lib v4 [work in progress]

jjergus:
Not finished yet (I haven't migrated any examples yet, not to mention the XHP guides' text, also there might be more issues similar to the last commit that I haven't found yet) but I thought you might have suggestions that would be worth addressing before I convert more stuff. The interesting commits are probably

- `migrate src/site/xhp` -- mostly straightforward, but you might have suggestions about better namespace names or better ways to `use` them (e.g. should we `use type` all HTML elements or just `use namespace Facebook\XHP\HTML`?)
- `fix APCCachedRenderable double rendering` -- mainly, is it expected that we'd run into issues like this when migrating to xhp-lib v4?


jjergus:
Things that might be worth automating:

- migrate all references (extends, param/return types, ...) from `:x:whatever`, `XHPWhatever` to `x\whatever` (must also add `use namespace Facebook\XHP\Core as x;`)
- `render` -> `renderAsync` (maybe include as part of the above, since it's the new base class that requires it)
- migrate `xhp class foo:bar:baz` to `namespace foo\bar; xhp class baz` (this may be suboptimal namespace name and/or cause conflicts)
- fully qualify names (prepend `\`) when moving an xhp class from global into a namespace
- add `use type Facebook\XHP\HTML\{div, a, img, ...};` for all HTML elements used in the file
- unCamelCase `XHPChild\atLeastOneOf` --> `XHPChild\at_least_one_of` etc.

src/site/xhp/APCCachedRenderable.php
```diff
@@ -9,36 +9,40 @@
...
+    if (\is_string($str)) {
       $ret = <x:frag>{new APCCachedRenderable($str)}</x:frag>;
       return $ret;
     }
     return null;
   }
 
-  public static function store(string $key, \XHPRoot $content): void {
-    $str = $content->toString();
+  public static function store(string $key, x\node $content): void {
```
fredemmott: Doesn't need to be in this diff, but this should be refactored to be an async function

jjergus: Yeah I started with that, but it's used from multiple functions and often in loops, so I left it for later.

jjergus:
OK this now passes all tests (the remaining failures are HHAST bug, I'll fix it there), so the only thing that remains is actually updating the guide text to be consistent with the updated examples.

Example changes are mostly the same as the previous src/ changes, possibly notable:

- I changed the `addClass` example to use a standard HTML element instead of a custom XHP class, to avoid relying on the deprecated attribute clobbering. (This is the only one so far that also has updated text.)
- There's an example specifically about attribute clobbering, so I preserved it there even though it's deprecated. (I'll make sure to mention it in the text.)
- It is no longer possible to call `getID()` from inside `renderAsync()` ("Can't addAttribute after render") so I had to modify 2 examples that did that.

fredemmott:
> `It is no longer possible to call getID() from inside renderAsync() ("Can't addAttribute after render") so I had to modify 2 examples that did that.`

Perhaps this constraint should be moved to be *after* render, rather than set in the middle of it? Either that, or perhaps we should consider the `id` attribute more special, like a lateinit property instead of an attribute? I think this pattern is valuable - @lexidor , @swahvay ?

lexidor:
> > `It is no longer possible to call getID() from inside renderAsync() ("Can't addAttribute after render") so I had to modify 2 examples that did that.`
> 
> Perhaps this constraint should be moved to be _after_ render, rather than set in the middle of it? Either that, or perhaps we should consider the `id` attribute more special, like a lateinit property instead of an attribute? I think this pattern is valuable - @lexidor , @Swahvay ?

https://github.com/hhvm/xhp-lib/issues/279#issuecomment-668826892 Linking to the issue in xhp-lib.

lexidor:
You should be allowed to mess with your state during render. It can create a headache, since you could `await` in your `renderAsync()` function though. Not allowing your attributes to change during render would break a lot of XHP already out there.

guides/hack/21-XHP/01-introduction.md
```diff
@@ -1,8 +1,71 @@
...
+```Hack
+$user_name = 'Fred';
+echo "<tt>Hello <strong>$user_name</strong></tt>";
+```
+
+However, with XHP, it looks like this:
+
+```
+$user_name = 'Fred';
+echo <tt>Hello <strong>{$user_name}</strong></tt>;
```
lexidor, line 16: `echo` won't work here, you need to call `->toStringAsync()`

fredemmott: we should generally try to remove all of these inline snippets and have them in external files so they're typechecked

jjergus: Good catch, I'll fix it.

> we should generally try to remove all of these inline snippets and have them in external files so they're typechecked

Agreed but the way it currently works, it would be very annoying for all these short examples. Maybe we could add a build step that extracts examples from the text (I think you suggested something like that before)?

fredemmott: Yeah, I think the best way to do this is a github action that:

1. extracts
2. generates output
3. automatically creates a pull request against the pull request origin branch

... I'm not sure where the source repo would be for 3 though - don't really want it to be this repo.

jjergus: Why not do it in bin/build.php? Ideally I'd want my examples tested before I commit them.

fredemmott: To make things easier for people who approach it as "I'm just editing markdown, no need for the full dev environment" - i.e. they use the github web editor, or just edit the markdown files without running the test.

Unknown

Do it in build.php, but make the CI check that build.php has no committable changes, and submit a PR-against-the-PR if not?

if done in build.php, it should prompt (unless some env var/flag is provided), rather than silently modifying source files

jjergus: > if done in build.php, it should prompt (unless some env var/flag is provided), rather than silently modifying source files

I think I'd rather keep the auto-generated output out of source files completely (no part of the source .md file should be generated), we should keep it in .out/.expect or similar files that get pulled into the final HTML during build.

- build.php would generate any out/expect files that are missing, leave existing ones alone.
- hacktest would verify that out/expect file exists for every example and that the output matches.

jjergus:
I'll continue this in #901 (already addressed most feedback there)

PR #899, opened 2020-08-04
[easy] Fix variable name typo in documentation

dschafer:
This variable is $v in the signature, but $c in the body. Since it's a Car, I consolidated on $c.


Wilfred:
Woops, thanks for spotting and fixing!

PR #901, opened 2020-08-05
rewrite XHP guide for namespace support + xhp-lib v4

jjergus:
**only review the last commit** (https://github.com/hhvm/user-documentation/commit/16b4459a1bf8a1424fe489710626f0a113fd52dd), the rest is from #898

notes:

- removed "introduction", it was a copy of the left nav menu, just with some items wrong
- renamed "some basics" to "introduction"
- moved 2 sections from there to a new "setup" guide, as they became much longer
- everything uses namespaces + xhp-lib v4 by default, but there are "Historical note" sections that document xhp-lib v3 behavior for all significant differences
- I don't really like the last 3 guides (migration, guidelines, further resources) but I didn't want to do too much in a single PR so I mostly left them alone (only fixed anything that was incorrect or misleading)

preview: http://35.162.222.227:8080/hack/XHP/introduction
(will try to keep it up for a few days; sorry no https, you might get MITMed and get evil documentation instead)


lexidor:
_It seems I'll have to redo the review, since the comments are not bound to the line of code I made the comment for._

guides/hack/21-XHP/19-migration.md
```diff
@@ -31,11 +31,12 @@ strings around in the end.
...
 
 ```Hack
-// Assume class Uri
-class :ui:component-link extends :x:element {
-  attribute Uri $uri @required;
+namespace ui;
+
+class link extends x\element {
+  attribute Uri $uri @required;  // Assume class Uri
   attribute string $text @required;
```
lexidor, line 38: TIL that you can have more than one attributes declaration in a class.
I thought that you had to put all attributes in one declaration.

lexidor: I have found the source of my confusion. In the [hhvm 4.3.0 blog post](https://hhvm.com/blog/2019/04/30/hhvm-4.3.0.html) the rules for multiple `children` and `category` declarations changed. Multiple `attribute` declarations are still allowed.

guides/hack/21-XHP/04-setup.md
```diff
@@ -0,0 +1,65 @@
+## Namespace Support
+
+Support for namespaced XHP classes (elements) has only recently been added to
+HHVM (as of HHVM 4.68). Depending on the exact HHVM version, it may not be
```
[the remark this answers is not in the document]
fredemmott: IMO we should strongly discourage use of v4 with whatever version of HHVM have the current defaults, and strongly discourage v3 after the default is flipped

guides/hack/21-XHP/16-extending-examples/attribute-transfer.inc.php
```diff
@@ -1,10 +1,14 @@
...
-  attribute :div; // inherit from attributes from <div>
+use namespace Facebook\XHP\Core as x;
+use type Facebook\XHP\HTML\{div, XHPAttributeClobbering_DEPRECATED};
+
+final xhp class ui_my_good_box extends x\element {
+  attribute :Facebook:XHP:HTML:div; // inherit attributes from <div>
   // Make sure that attributes are transferred automatically when rendering.
-  use XHPHelpers;
-  protected function render(): \XHPRoot {
+  use XHPAttributeClobbering_DEPRECATED;
```
fredemmott, line 9: This example should probably be replaced with an example of attribute splat

guides/hack/21-XHP/16-extending-examples/required-attributes.php.hhconfig
```diff
@@ -0,0 +1,4 @@
+disable_xhp_element_mangling=true
+enable_xhp_class_modifier=true
+disable_xhp_children_declarations=true
+check_xhp_attribute = true
```
fredemmott, line 4: perhaps we should add support for a prepended/default hhconfig for examples in the test

guides/hack/21-XHP/16-extending.md
```diff
@@ -3,16 +3,29 @@ items that are not in the standard HTML specification.
...
+  [base class](/hack/XHP/interfaces)
+* implement the method `renderAsync` to return an XHP object (`x\node`) or the
+  respective method of the chosen base class
 
 @@ extending-examples/basic.inc.php @@
 @@ extending-examples/basic.php @@
 
+**Historical note:** Before XHP namespace support (in XHP-Lib v3), use
+`class :intro_plain_str` instead of `xhp class intro_plain_str` (no `xhp`
+keyword, but requires a `:` prefix in the class name).
```
fredemmott, line 27: we probably want an fb span here saying that (for now) we're also using the old syntax, and no namespace support

guides/hack/21-XHP/04-setup.md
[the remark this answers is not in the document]
jjergus: Heh I intentionally made it vague ("recently as of HHVM 4.68" -- though I guess ~4 months might be stretching the definition of "recently" a bit) because I wasn't sure what the exact version was, and it's probably impossible to pinpoint an exact version anyway (IIRC there have been some bugfixes and improvements after the support was first added...) but I agree, it would be better to clarify this.

That said, I also agree that we should discourage people from using XHP namespaces with HHVM versions that technically support them, but are not thoroughly tested (e.g. not used in production by any project), so I'd highly discourage using any HHVM version from before we converted at least the user-documentation site, and probably (like Fred) said, from before these flags are on by default.

So probably something like: "...has been added around HHVM 4.46, but hasn't been enabled by default until HHVM 4.xx, so we recommend using HHVM 4.xx [same xx] or newer. In older HHVM versions, modern XHP features including namespace support can be enabled by adding the following flags to your `.hhconfig`:" (also clarifies that some flags are technically unrelated to namespaces -- but given that we support/test either with none of them (xhp-lib v3) or with all of them (xhp-lib v4), I'd discourage people from using other subsets -- I also wrote that in the guide later)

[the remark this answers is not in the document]
fredemmott: I'm also going to chagne xhp-lib's minimum HHVM version once the default flips, before 4.0.0 final - this is so that when people do `composer require facebook/xhp-lib`, they get a version that works with the default settings. This isn't currently an issue as composer will not install an rc version unless you explicitly request it.

guides/hack/21-XHP/07-basic-usage.md
```diff
@@ -51,9 +101,11 @@ class defines what attributes are available to objects of that class:
 echo <input type="button" name="submit" value="OK" />;
 ```
 
-Here the `:input` class has the attributes `type`, `name` and `value` as part of its class properties.
+Here the `input` class has the attributes `type`, `name` and `value` as part of its class properties.
```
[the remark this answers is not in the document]
fredemmott: there's a bit of nuance here: they are not properties, but the typechecker considers them to be properties

Runtime: `->:foo` is `->getAttribute('foo')`
Typechecker: `->:foo` is retrieving the instance property `:foo`

guides/hack/21-XHP/16-extending-examples/attribute-transfer.inc.php
jjergus: How do you use attribute splat to get similar behavior? `<div {...$this}>`? Does it ignore attributes that exist on $this but don't exist on <div>?

guides/hack/21-XHP/16-extending-examples/required-attributes.php.hhconfig
jjergus: Doesn't really matter if we wait until these are on by default, I can just remove that commit then.

guides/hack/21-XHP/16-extending-examples/attribute-transfer.inc.php
lexidor: The spread operator expands to all attributes which are valid on the tag to which the attributes are being transfered to. If the attribute doesn't have a value `null` or `unset` in the transfered from tag, the attribute is not transfered. The relative order of the attribute locations in the transfered to tag does matter. Later occurrences of an attribute override previous ones. For an example see [3.24.0 blog post](https://hhvm.com/blog/2018/01/16/hhvm-3.24.html).

jjergus: Should I recommend using attribute inheritance (`attribute :some:other:element;`) + the splat operator?

I assumed we'd want to move away from attribute inheritance along with attribute clobbering (I removed the inheritance section and said we recommend explicitly listing attributes in the clobbering section), but maybe that's incorrect...?

fredemmott: AIUI attribute inheritance is likely staying for as long as attributes remaining a distinct thing to properties (then interfaces can be used) - for now, inheritance + splat seems like the right mix.

Requiring an explicit list will get very verbose - for example, `class`, and `style` are likely to be supported pretty much everywhere

guides/hack/21-XHP/10-interfaces.md
```diff
@@ -1,22 +1,40 @@
...
+one of its two subclasses:
+
+- `\Facebook\XHP\Core\element` (`x\element`): most common; subclasses implement a
+  `renderAsync()` method that returns another `node`, and XHP-Lib automatically
+  takes care of recursively rendering nested XHP objects
+- `\Facebook\XHP\Core\primitive` (`x\primitive`): for very low-level nodes that
+  need exact control of how the object is rendered to a string, or when the
+  automatic handling of nested XHP objects is insufficient; subclasses implement
+  a `stringifyAsync()` method that returns a `string` and must manually deal with
+  any children
```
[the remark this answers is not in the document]
fredemmott: > implementing tags which are actual SGML, like <li> and <body>

It's a bit more generic than that, but that's a handy rule for HTML. I think it's accurate to say that it's any element that can not be replaced by or defined by other elements. This usually means that they have a direct representation in the rendered output.

The fact that they can have children makes it hard to talk about this clearly and unambiguously - if we ignored children, 'atomic' might be a good way to describe them.

guides/hack/21-XHP/16-extending-examples/attribute-transfer.inc.php
lexidor: I don't quite understand the comment about how interfaces could be used when attributes are backed by properties instead of an associative container. Would you mind clarifying that bit?

fredemmott: ... TIL that interfaces can't declare properties. I was thinking of this, but it's invalid:

```Hack
interface Foo {
  public string $bar;
}

class HerpDerp implements Foo {}
```

lexidor: Would traits fit the use case? I know they are sometimes frowned upon, but they seems to be up for the task.

I am still a bit unsure about how this could recreate the splat operator.

fredemmott: This wouldn't affect the splat operator, it would remove the need for attribute inheritance as a distinct concept though.

jjergus:
From the new commits, these are probably worth reviewing:

- https://github.com/hhvm/user-documentation/pull/901/commits/6e193f8cd392974072af42e4f6a033caa8f71a18 rewrite attribute transfer section to use ...splat
- https://github.com/hhvm/user-documentation/pull/901/commits/29dfbabc09d0a210f5d5b043df92d78feeb7d338 update setup instructions

@lexidor @fredemmott

PR #909, opened 2020-08-28
[ Help wanted for // partial migration ] Hhvm version 4 72

lexidor:
I have one failing test locally, something about the file pattern being wrong...

Also, XHP v3 needs an upgrade for an is_array, and usox/hackttp needs one too.

~There are `// partial` files with top level code in them. I'll make them `// strict`.~
Making these files strict changes / breaks the api examples by cluttering them with boilerplate.
Many partial files are skipif, which need a unique name for each.
I don't have the automation required to "upgrade" these files.
Leaving as is for someone else to pick up.


jjergus:
Thanks for the fixes! I pulled them into #901 so as to avoid merge conflicts when both of us try to bump the HHVM version to a different number...

PR #911, opened 2020-08-28
Update AsyncMysql docs fixes #907

### LINKED ISSUE #907: Update Async_mysql docs
/assign @lexidor

# Where is the problem?

Async_mysql

# What is the problem?

Out of date
Docs mention `\mysqli` which was removed in hhvm 4.9.

Incorrect use of the api
Error in example, the string `Fred Emmott` is being passed through `->escapeString()` before being handed to the `%s` modifier, which will escape the string again.

"Best" practice
The example query literal `WHERE name = %s` should use `%=s` to be nullsafe (not needed in this example, but something I've found value in to just do in all cases.)

Modern methods
Most dynamic query problems can now be solved in a safe way using `SQL\Query`, so promote `->queryAsync()` over `->escapeString()`.

 - Build ID: HHVM=HHVM-4.68.0:HSL=v4.41.0:2020-08-17T20:16:46+0000:7af74eee111304c3c25319df9d95501d92a64697
 - Page requested: /hack/asynchronous-operations/extensions
 - Page requested at: Thu, 27 Aug 2020 23:16:08 +0000

### LINKED ISSUE #910: AsyncMysql tests have been skipped since hhvm 4.9 (removal of mysqli)
This statement dies with the message `Fail` since hhvm 4.9.0, when mysqli was removed.

api-examples/__includes/async_mysql_require_env.inc.php
```HACK
if (!extension_loaded('mysql') || !function_exists('mysqli_connect')) {
  die('Skip');
}
```

lexidor:
fixes #907

The things I said here are based on my personal limited experience of using AsyncMysql.

queryAsync is not in the references, so my link to it is invalid. We may need to fix #910 before we can merge this.

### DIFF
--- a/guides/hack/15-asynchronous-operations/19-extensions.md
+++ b/guides/hack/15-asynchronous-operations/19-extensions.md
@@ -31,7 +31,7 @@ This should be your preferred method. It is powerful enough to express all queri
 For example:
  - selecting a post by ID
  - selecting the post with the most likes made within the last _n_ days by user _x_
- - updating all profiles from users who have been active in the last _n_ days with by giving them a badge
+ - updating all profiles from users who have been active in the last _n_ days by giving them a badge
 
 ### [`queryAsync`](../references/class/AsyncMysqlConnection/queryAsync)
 This method is a lot more powerful than `queryf` and should be used when the structure of your query is not known in advance. The `%Q` modifier can be used to insert query fragments into other query fragments.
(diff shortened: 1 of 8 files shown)


lexidor:
image

jjergus:
Veery nice, thanks!

guides/hack/15-asynchronous-operations/19-extensions.md
```diff
@@ -3,24 +3,64 @@ infrastructure. Async is especially useful with database access and caching, web
...
+There is also a function to ensure that queries to be executed are safe called
 [`escapeString`](../reference/class/AsyncMysqlConnection/escapeString/).
 
+## Choosing a query method
+
+Choosing the right tool for the job is very important.
+Each of the three methods has its strengths and weaknesses.
+
+### [`queryf`](../reference/class/AsyncMysqlConnection/queryf/)
+This should be your preferred method. It is powerful enough to express all queries with a known structure, but with unknown data.
```
fredemmott, line 29: queryf should be deprecrated; `queryAsync()` should pretty much always be preferred:
- it uses hack arrays instead of legacy hack collections
- it matches current naming conventions

fredemmott: It also makes it possible to pass queries and data independent of connections, which in turn allows building much more powerful abstractions on top of it

guides/hack/15-asynchronous-operations/19-extensions.md
```diff
@@ -3,24 +3,63 @@ infrastructure. Async is especially useful with database access and caching, web
...
+[`query`](../reference/class/AsyncMysqlConnection/query/),
+[`queryf`](../reference/class/AsyncMysqlConnection/queryf/), and
+[`queryAsync`](../references/class/AsyncMysqlConnection/queryAsync)
+all of which are `async`. The naming conventions for async method have not been applied consistently.
+There is also a function to ensure that queries to be executed are safe called
 [`escapeString`](../reference/class/AsyncMysqlConnection/escapeString/).
 
+## Choosing a query method
+
+### [`queryAsync`](../references/class/AsyncMysqlConnection/queryAsync)
```
fredemmott, line 25: From test:

```
1) HHVM\UserDocumentation\Tests\InternalLinksTest::testInternalLink with data set #2037
1084>>> 404: /hack/references/class/AsyncMysqlConnection/queryAsync
1085>>> Linked from:
1086>>>  - /hack/asynchronous-operations/extensions
```

- need a trailing /
- need `reference`, singular
- prefer `/hack/reference` rather than relative path throughout
- ... but prefer just `AsyncMysqlConnection::queryAsync` in backticks over that - will automatically be converted to a link. This would be a good change for most of this file, but doesn't need to all be fixed here

fredemmott (changes requested):
Content is great, thank you :) some bad links

lexidor:
image
Edited to CSS in dev tools to make the main content wider to make everything fit in the screenshot.
The actual page width has not been altered.

lexidor:
@jjergus Just a quick ping because I had abandoned this PR for so long. I fixed the links Fred asked for.

jjergus:
Thanks!

PR #912, opened 2020-08-28
Remove extensions which are not supported

lexidor:
redis, sessions, and tokenizer were removed in 4.56, 4.4, 4.0
MongoDB is no longer maintained since mid 2017
The other dynamically loaded extensions
may have broken since their last commits


Wilfred:
Thanks :)

PR #913, opened 2020-08-30
Compare nothing to noreturn

lexidor:


guides/hack/11-built-in-types/73-noreturn.md
```diff
@@ -8,4 +8,19 @@ function something_went_wrong(): noreturn {
 }
 ```
 
-`exit` is an example of a library function with a `noreturn` type.
+`invariant_violation` is an example of a library function with a `noreturn` type.
+
+`noreturn` informs the typesystem that code execution can not continue past a certain line.
+In combination with a conditional, you can refine variables, since the typesystem will take note.
+This is actually how [invariant](../expressions-and-operators/invariant) is [implemented](/hack/reference/function/HH.invariant).
```
jjergus, line 15: Did you mean `invariant_violation`? (`invariant` can return.) I'm also not sure if this is strictly true, IIRC `invariant_violation` existed long before the `noreturn` type so might still have custom implementation in the typechecker...

guides/hack/11-built-in-types/74-nothing.md
```diff
@@ -15,14 +15,38 @@ Let's work out the hierarchy of scalar types. Forget about nullable types and `d
...
+
+`nothing` is an amazing solution which allows you to keep working what you were working on, without having the nagging feeling of the typeerror that needs to be fixed. The trick is defining a function with the return type of `nothing`. This function may not return anything (just like a `noreturn` function). You can now use this function as a placeholder for the value which you could not create. I have called this function (and an opaque type alias) `undefined`, as an homage to Haskell [undefined](https://wiki.haskell.org/Undefined).
```
jjergus, line 30: I don't have a strong opinion but I feel like these two paragraphs could probably be shortened to one sentence, something like "This allows you to implement a placeholder function that can be used in place of any type without causing type errors." and let people decide for themselves whether that's something they want or what they'd use it for (the example illustrates your use case pretty well I think, probably doesn't need a long text description on top of that).

Personally, I'm not sure whether using `nothing` to silence a type error like this (possibly turning it into a runtime error unless I have a staging environment like you described) is something I'd want to do, so that's another reason why it might be better to present the example more factually, not as a recommendation.

guides/hack/11-built-in-types/73-noreturn.md
lexidor: Rereading the paragraph, I understand the confusion. The `This` in "This is actually how [invariant](../expressions-and-operators/invariant) is [implemented](/hack/reference/function/HH.invariant) refers to the code sample below.
`invariant(X, Y)` is actually shorthand for
```HACK
if (!(X)) {
  invariant_violation(Y);
}
```

I wrote this with the intention of demonstrating that the concept of checking a condition before a noreturn is a very common idiom. I will either need to rephrase this or turn the `invariant_violation` name into a link to the invariant docs and remove the Sentence `This ... implemented`. I have not yet decided.

guides/hack/11-built-in-types/74-nothing.md
lexidor: I guess I couldn't contain my excitement about this "little hack". It is indeed very out of place when compared to the rest of the docs. I'll rephrase to be more in line with the rest of the docs and trim the excess.

This trick has brought back the ability to take some micro short term tech debt during development. Finishing my thought / thread of work and coming back to it later. I run into the given example use case a lot.

jjergus:
Another use case for `nothing` is if you have a generic class with a *contra*variant generic type param.

```hack
class Foo<-T> {}
function takes_any_foo(Foo<nothing> $foo): void {}
```

Similarly to how with a *co*variant generic type param you'd have:


```hack
class Foo<+T> {}
function takes_any_foo(Foo<mixed> $foo): void {}
```

jjergus:
Looks good to me, thanks!

Issue #727, opened 2019-09-23
String summary=subject1.concat(grade1)+subject2.concat(grade2)+subjec3.concat(grade3)+subject4.concat(grade4)+subjec5.concat(grade5);


GulmireYusuf:
Please complete the information below:

# Where is the problem?

- - - eg "The section describing how widgets work" - - -

# What is the problem?

- - - eg "It doesn't explain that widgets are singletons" - - -

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: HHVM=HHVM-4.23.0:HSL=v4.15.2:2019-09-20T21:41:50+0000:138dd8d9d3c423f9864bf5952632efca10d6696c
 - Page requested: /hack/expressions-and-operators/string-concatenation
 - Page requested at: Mon, 23 Sep 2019 00:49:52 +0000
 - Controller: GuidePageController

Issue #819, opened 2020-02-25
Simplify sample files required


Wilfred:
Currently, adding a new sample requires all of:

```
the_file.php
the-file.php.hhconfig // empty
the-file.php.hhvm.expect // usually empty
the-file.php.no.auto.output // empty
the-file.php.typechecker.expect // usually "No errors!"
```

I propose that `the_file.php` should be sufficient. The others should only be necessary if they differ from the default. Most sample files don't/shouldn't need to show their output.

fredemmott:
> the-file.php.hhvm.expect // usually empty

This is, in general, an antipattern - but a common one.

Agreed on the rest. `no.auto.output` is not required, but is usually desired, but that should probably be the default behavior.

fredemmott:
FYI changes to this set will require both changes to the 'fast' runner embedded here, and to hphp/test/run.php

Wilfred:
> This is, in general, an antipattern - but a common one.

Just to clarify: you're proposing that most files should output something, so we can check the output?

I was thinking a valid type check was usually sufficient. Otherwise many of our examples require a distracting `<<__EntryPoint>>` shim.

The alternative is we develop some kind of `// BEGIN_INCLUDED_SAMPLE` `// END_INCLUDED_SAMPLE` convention for code files. I'd really like this for simple cases like assignments or if statements, so I can only show the if statement but still have some checking.

fredemmott:
> Just to clarify: you're proposing that most files should output something, so we can check the output?

Yes, or at least it should clearly be an intentional decision to have a script with no output

> The alternative is we develop some kind of // BEGIN_INCLUDED_SAMPLE // END_INCLUDED_SAMPLE convention for code files.

@jjergus recently added support for @example-start and @example-end :)

jjergus:
> Otherwise many of our examples require a distracting <<__EntryPoint>> shim.

I added some magic to automatically strip the common boilerplate (@example-start and @example-end can be used to override it but usually it does the right thing even without those; by "usually" I mean I'm parsing the code with regular expressions, what could go wrong).

https://github.com/hhvm/user-documentation/blob/c5c62ecb0f5862d371b16333a60e473dce00f5fb/src/markdown-extensions/ExampleIncludeBlock.php#L101

lexidor:
[...]

Issue #745, opened 2019-11-06
Examples usability


fredemmott:
It would be good if we could:

- allow examples in code blocks in `.md` files
- automatically generate the `*expect*` files

One potential flow would be:
- I create a pull request - fredemmott/fix-foo-branch -> hhvm/master
- Travis-CI creates a hhvm-travis-bot/pr1234-examples -> fredemmott/fix-foo-branch pull request with the expect files

This would also require auto-generation of standard boilerplate, like the entrypoint function name.

It could either remove the example and add `@@ extracted_inline123.hack @@`, or it could add `guides/foo/bar/.../extracted_inline123.{hack,.hack.hhvm.expect}` etc, and we add `extracted_inline*.hack` to .gitignore

fredemmott:
fbmarkdown provides an AST, so we don't need regexps etc to find the code blocks :)

jjergus:
I'm looking into this now; I like the idea of extracting the examples from the .md files -- ideally, this wouldn't just be the example code, but everything else related to each example too (whichever of expectf, expectre, hhconfig is needed -- not including regular .expect files as those can be generated automatically), solving #819 at the same time. I'm thinking an example format like this:

## Common case

```
# some-name.hack
echo "Hello, world!\n";
```

This would be all you need to write for a basic example and it will be **run and tested**.

During bin/build.php, this extracts `some-name.hack`, runs it through HHVM and puts the output in `some-name.hack.hhvm.expect` (it can also generate an empty hhconfig, typechecker-expect or whatever else the test runner requires).

I'm not sure whether `some-name.hack` should be committed and pushed, but `some-name.hack.hhvm.expect` definitely should because

- it would be very expensive to generate the output for each example during bin/build.php
- it allows the author and reviewers to verify the output is as expected
- if the output unexpectedly changes, we will catch it, instead of silently generating a new, bad output file

In the rendered output:

- the `#` comment is removed
- example output is added
- if the example doesn't contain `<<__EntryPoint>>`, we add the boilerplate automatically
- whether or not it contains an `<<__EntryPoint>>`, in either case we can also automatically add a namespace line (unique for each example, probably?), `use namespace` for any HSL functions used, `\init_docs_autoloader()`

Specifying a name for each example is important because:

- it allows us to identify the relevant .hhvm.expect file even if the example code changes (an alternative would be to extract each example into a file named e.g. hash of the example code, but that would disassociate the example with its previous .expect file whenever the code changes)
[...]

PR #915, opened 2020-09-16
automatic example code extraction (initial version)

jjergus:
Initial version of automatic example code extraction, as discussed in #727. I ended up using e.g. <code>```.hhconfig</code> as the format to specify extra files to be written where needed, but these should be very rare so it doesn't really matter that much, I can change it to whatever you like more (there are only a few .hhconfigs and most probably only exist because the defaults were different when the example was added; for .expectf files I want to handle common cases automatically -- current version already handles file paths in the output, I'll add more common cases as I run into them).

I expect this to have some bugs and missing features -- I plan to convert all examples from `@@ filename @@` to this, fix any bugs and add missing features as I do that. There are 2 example code blocks manually converted here (last commit).

You can look at commits separately for easier review.

More detailed description (copied from the code):

```
/**
 * Processes code blocks starting with one of:
 *   ```hack filename.ext[.type-errors]
 *   ```filename.ext[.type-errors]
 * where ext is one of self::ALLOWED_EXTENSIONS.
 *
 * Depending on the subclass, either writes all related files (manual builds),
 * or complains if any file is missing or outdated (CI/automated builds).
 *
 * Optional extra files can be specified using one of:
 *    ```suffix
 *    ```.suffix
 *    ```filename.ext.suffix
 * inside the code block.
 *
 * If not explicitly specified, these files are generated automatically:
 * - filename.ext.typechecker.expectf + filename.ext.example.typechecker.out
 *     by running the typechecker if the main file is *.type-errors
 *     (file paths replaced with %s)
 * - filename.ext.typechecker.expect
 *     with "No errors!" otherwise
 * - filename.ext.hhvm.expect
 *     by running HHVM, only if the main file is not *.type-errors
 * - filename.ext.hhvm.expectf + filename.ext.example.hhvm.out
 *     instead of .expect if the output contains file paths (e.g. in errors)
 * - filename.ext.hhconfig (empty)
 */
```

One intentional change I made is that for `*.type-errors` examples, I include the typechecker output in the rendered HTML instead of HHVM output -- I think this should generally be more helpful/user-friendly (most likely it outputs the same error but nicer), but if I run into examples where that isn't the case, I'll change it back or make it configurable.


guides/hack/11-built-in-types/71-dynamic-examples/coercion_from_union.php.type-errors.example.typechecker.out
```diff
@@ -0,0 +1,7 @@
+File "/home/example/coercion_from_union.php", line 16, characters 17-18:
```
Wilfred, line 1: Could we use a different extension? In the Hack test suite itself, we have foo.php source code, foo.php.exp expected output, and we regenerated foo.php.out on every run.

My assumption was that .out files shouldn't be committed.

Wilfred (approved):
picard_win

This is a massive improvement and so much easier to work with. It'll make casual contributions easier, and help regular contributors too. Thanks :)

guides/hack/11-built-in-types/71-dynamic-examples/coercion_from_union.php.type-errors.example.typechecker.out
jjergus: To be precise, this is not an "out" file, this is an "example out" file. These are needed when the output is non-deterministic (like this one because it contains randomly generated/system specific file paths).

- When the output is deterministic, we just have an .expect file which is used for testing as well as in the rendered HTML guides.
- When the output is non-deterministic, we have an .expect**f** file for testing, and need another file to be placed in the rendered HTML output since we don't want `%s` etc. visible in the guides (hence the "example out" file, which is a new concept not part of the normal Hack test suite).

I used the same pattern we've been using before, see e.g. the example out files here: https://github.com/hhvm/user-documentation/tree/master/guides/hack/15-asynchronous-operations/22-generators-examples (but now it's possible to have both "example.hhvm.out" and "example.typechecker.out" depending on whether we want to show HHVM or Hack output in the guide).

Wilfred:
@jjergus did you mean to link #727? I don't see the relevance.

jjergus:
Oops, sorry. I meant to link #745 which has the most recent dicussion, but #819 is also relevant.

PR #920, opened 2020-09-28
support for patching hackttp

jjergus:
Updating the docs site to the latest HHVM version often gets temporarily blocked because the [hackttp](https://github.com/usox/hackttp) dependency causes Hack errors with the respective HHVM version (see https://github.com/usox/hackttp/commits/master -- pretty much every recent commit is due to that). The owner has been pretty responsive about merging our PRs and tagging releases, but it does occasionally take a while so it would be good if we didn't have to rely on it.

This adds the option to automatically patch the hackttp package after composer install/update, with an arbitrary patch (the included patch is from the latest hackttp commit, which didn't make it into a tagged release yet). We can revert this when a new hackttp release is tagged, then re-introduce it whenever we need it again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "require": {
         "facebook/xhp-lib": "^4.0",
         "hhvm": "4.74.*",
-        "usox/hackttp": "dev-master",
+        "usox/hackttp": "^0.5.2",
         "hhvm/hhvm-autoload": "^3.0",
         "hhvm/type-assert": "^4.0",
         "facebook/hack-codegen": "^4.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -17,7 +17,10 @@
         "hhvm/hhast": "^4.0",
         "facebook/definition-finder": "^2.0.0",
         "hhvm/hacktest": "^2.0",
-        "usox/hackmock": "dev-master",
```
jjergus, line 20: Not needed if we're using a tagged release of hackttp, because those don't contain tests (dev-master contains tests and I couldn't find a way to remove them).

Having this here is undesirable because then we also have to ensure that hackmock is compatible with the latest HHVM, in addition to hackttp.

composer.json
```diff
@@ -17,7 +17,10 @@
...
         "facebook/fbexpect": "^2.6.1"
+    },
+    "scripts": {
+        "post-install-cmd": "patch --dry-run -R -p1 -i hackttp.patch || patch -p1 -i hackttp.patch",
+        "post-update-cmd": "patch --dry-run -R -p1 -i hackttp.patch || patch -p1 -i hackttp.patch"
```
jjergus, line 24: first part succeeds if the patch is already applied and causes the second part to be skipped

jjergus:
I'll merge this now so I can update the docs site to the latest HHVM version, if there are any requested changes later I'll address them in a separate commit (or revert this if the whole approach ends up being terrible or no longer needed).

PR #923, opened 2020-10-01
Add intructions for whitelisting suppression codes

### LINKED ISSUE #905: Document allowed_fixme_codes_strict for new projects
refs #904

Anyone starting a new project will have to add these hhconfig settings.
The superset of all opensource hh_fixmes should be added in the example hhconfig files.

lexidor:
closes #905


lexidor:
I have chosen for curl over jq + error suppression script from Fred. The default jq version on ubuntu is 1.5 and the jq script requires 1.6.

jjergus:
Thanks!

PR #926, opened 2020-10-09
Update FB-only header for greater clarity

mikeweilgart:
Make it clear that the open source docs *are* the right place to learn the language itself, just not the place for dev environment setup within FB, per conversation with Fred Emmot.


Wilfred:
Nice, this has caused confusion several times.

PR #927, opened 2020-10-09
Fix formatting error in escape sequence table

mikeweilgart:
Previously the dollar sign and double quote were rendered without a preceding backslash.


mikeweilgart:
CLA should be good now.  For reviewers: I didn't make an issue for this as it's just a typo fix; if an issue is still desired for fixes this small please let me know.

Wilfred:
@mikeweilgart a standalone PR is perfect, thank you :)

PR #930, opened 2020-10-09
Update to HHVM 4.78

fredemmott:


fredemmott:
Converted to draft: definition-finder has issues which I've fixed in master, but need to do a release for.

PR #931, opened 2020-10-13
Function references + function type

jjergus:
- document function references (aka first-class function pointers; I tried to call them "references" consistently here, but if you prefer "pointers" I can change that), which are enabled by default as of this week's HHVM release
- add a short page about the `(function(...): ...)` built-in type (separate page because this is not specific to FCFP, also applies to lambdas)

The text is largely copypasta from @periodic1236's internal announcement, reorganized slightly (e.g. more prominence for namespaces since they are used more commonly outside FB; removed comparisons with fun/class_meth except for a short section for older HHVM versions).

### DIFF
--- a/guides/hack/05-functions/30-function-references.md
+++ b/guides/hack/05-functions/30-function-references.md
@@ -1,0 +1,183 @@
+It is often useful to refer to a function
+(or a [method](https://docs.hhvm.com/hack/classes/methods))
+without actually calling it&mdash;for example,
+as an argument for functions like `Vec\map`.
+
+**Note:** The following syntax is only supported since HHVM 4.79. For older HHVM
+version, see [old syntax](#old-syntax) below.
+
+To refer to a top-level (global) function named `foo`, you can write:
+
+```
+foo<>; // a reference to global function 'foo'
+```
+
+You can think of it like it’s a function call with empty generics, but the list
+of arguments has been omitted (the missing parentheses).
+
+The following example stores a function reference in a variable and later calls
+the function. Note that the type checker keeps track of the
+[function's type](/hack/built-in-types/function)
+and correctly checks all calls.
+
+```
+function foo(string $x): bool { ... }
+
+$x = foo<>;     // $x:(function(string): bool)
+$y = $x('bar'); // $y:bool
+$_ = $x(-1);    // error!
+```
+
+This syntax supports namespaced names the same way you would refer to them as
+part of a function call, so the following function references are all
+equivalent:
+
+```
+$fun = \Foo\Bar\Baz\derp<>;
+```
+
+```
+namespace Foo;
+$fun = Bar\Baz\derp<>;
+```
+
+```
+use namespace Foo\Bar\Baz;
+$fun = Baz\derp<>;
+```
+
+```
+use function Foo\Bar\Baz\derp;
+$fun = derp<>;
+```
+
+## Static methods
+
+Similarly, you can refer to a static method `bar` on a class `MyClass` by using
+the familiar method call syntax, without providing the call arguments. Just
+append type arguments (like `<>`) to the function call receiver
+(like `MyClass::bar`).
+
+```
+MyClass::bar<>; // a reference to static method 'MyClass::bar'
+```
+
+- **Private/protected** methods can be referenced using this syntax as long as
+  they are accessible to you in your local scope.
+- **Abstract** static methods cannot be referenced. Such methods cannot be
+  called, for they have no implementation. Invoking a hypothetical reference to
+  one would also be an error, so we simply don’t allow a reference to be
+  created.
+- In traits and other classes that are not marked `final`, you cannot use
+  `self::` in a reference. This is to avoid ambiguity and confusion around what
+  `self` actually refers to when the method is called (it depends on the static
+  context of a call, whereas function references can never receive the context
+  information because the target is only resolved once).
+- Also, `parent::` is never supported.
+
+## Generics
+
+If you wish to pass along explicit generic arguments, either as a hint to the
+type checker, or in the case of a function with reified type parameters when
+they are required, that is also supported:
+
+```
+i_am_erased<int, _>; // erased generics (note wildcard)
+i_am_reified<C<string>>; // a reified generic
+```
+
+- Keep in mind that generics, if any, still must be provided at the location
+  where the function reference is created, rather than where it is used/invoked.
+- The arity of the type argument list, if it is non-empty (i.e. not `<>`), must
+  match the declaration, just like for an ordinary function call.
+- The special wildcard specifier `_` may be provided in place of any or all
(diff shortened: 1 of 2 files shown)


guides/hack/05-functions/30-function-references.md
```diff
@@ -0,0 +1,183 @@
+It is often useful to refer to a function
+(or a [method](https://docs.hhvm.com/hack/classes/methods))
+without actually calling it&mdash;for example,
+as an argument for functions like `Vec\map`.
+
+**Note:** The following syntax is only supported since HHVM 4.79. For older HHVM
+version, see [old syntax](#old-syntax) below.
```
fredemmott, line 7: s/version/versions/

guides/hack/05-functions/30-function-references.md
```diff
@@ -0,0 +1,183 @@
...
+
+You can think of it like it’s a function call with empty generics, but the list
+of arguments has been omitted (the missing parentheses).
+
+The following example stores a function reference in a variable and later calls
+the function. Note that the type checker keeps track of the
+[function's type](/hack/built-in-types/function)
+and correctly checks all calls.
+
+```
```
fredemmott, line 23: At least this example should be split out to a separate file and tested

guides/hack/05-functions/30-function-references.md
```diff
@@ -0,0 +1,183 @@
...
+the familiar method call syntax, without providing the call arguments. Just
+append type arguments (like `<>`) to the function call receiver
+(like `MyClass::bar`).
+
+```
+MyClass::bar<>; // a reference to static method 'MyClass::bar'
+```
+
+- **Private/protected** methods can be referenced using this syntax as long as
+  they are accessible to you in your local scope.
```
fredemmott, line 66: Once you have such a reference to a private/protected method, who can invoke it?

guides/hack/05-functions/30-function-references.md
```diff
@@ -0,0 +1,183 @@
...
+// these all have errors!
+buzz<>;
+buzz<_>;
+buzz<mixed>;
+```
+
+## Introspection
+
+Function references can be cached in APC (the name will be resolved again when
+they are rehydrated) or passed to memoized functions. However, other
```
fredemmott, line 140: rehydrated is an unusual term - maybe 'unserialized', 'retrieved from APC' or similar?

fredemmott (approved):
A few inline nits and most of the larger examples should be extracted out for automatic testing, but overall looks god

jjergus:
> most of the larger examples should be extracted out for automatic testing

I mostly just copied the examples from the internal announcement; they're kind of hard to extract due to a combination of

- almost all of them show a valid + an invalid usage of a shared declaration, so they'd probably have to be `*.hack.type-errors` examples which doesn't feel great
- most of them also require additional code (unimportant declarations + boilerplate) that would probably make them harder to understand
- I'm almost done working on improvements to my example extractor (#915) that will automatically add most boilerplate, which should alleviate the previous point at least partially

But I agree at least *some* of the examples should be typechecked (and ideally executed too with some reasonable output to check) so I'll work on that, but probably in a separate PR since I want to get this in so that I can link to it from the 4.79 release notes.

I'll apply all the other feedback here before merging.

fredemmott (approved):

guides/hack/05-functions/30-function-references.md
```diff
@@ -0,0 +1,184 @@
...
+
+* `HH\is_fun`
+* `HH\fun_get_function`
+* `HH\is_class_meth`
+* `HH\class_meth_get_class`
+* `HH\class_meth_get_method`
+
+Be very careful and deliberate when using these, as they are loosely typed but
+will throw exceptions for bad arguments. Use of reflection APIs in production is
+also discouraged because it often has a large performance impact.
```
periodic1236, line 162: Is this still true when considering how most people run HHVM outside of FB?

guides/hack/05-functions/30-function-references.md
```diff
@@ -0,0 +1,184 @@
...
+
+Internally, function/method references are represented using special data types
+that are intended to be opaque. This means they cannot (or should not) be cast
+directly to a string or another type, or be accessed in any other way besides
+calling them.
+
+It is *sometimes* useful to determine what a function reference is pointing to,
+e.g. for use in logging messages. For these *limited* use cases, if you know
+enough about the input and output formats you expect, HHVM provides the
+following helpers (but note they are not well supported and may change):
```
periodic1236, line 152: We don't need to waffle as much in the documentation. These are the only native APIs that do what they do, and the ones you _should_ use, but internally we have our own that wrap them and we recommend the latter instead. Maybe we can add an FB-only blurb that mentions this?

guides/hack/05-functions/30-function-references.md
```diff
@@ -0,0 +1,184 @@
...
+
+* `HH\is_fun`
+* `HH\fun_get_function`
+* `HH\is_class_meth`
+* `HH\class_meth_get_class`
+* `HH\class_meth_get_method`
```
periodic1236, line 158: Are there pages for these that we can link to? We probably need to add docblocks to their definitions though.

periodic1236 (approved):
Mostly LGTM

guides/hack/05-functions/30-function-references.md
jjergus: Yep, these are linkified automatically. Link goes to https://docs.hhvm.com/hack/reference/function/HH.is_fun/ etc.

Docblocks could be nice but I think it's pretty clear what each one does in the context of this guide, so not really necessary.

guides/hack/05-functions/30-function-references.md
jjergus: Sounds reasonable, do you have a suggestion on what exactly to put here? I'm not sure how much we want to encourage/discourage people from using these. If they're unsupported and likely to change, we should probably mention that.

We can add a FB-only block, but we're only hiding those using CSS so shouldn't mention anything sensitive.

guides/hack/05-functions/30-function-references.md
jjergus: I think it's true in general that "using reflection in production is slow" but these APIs are no worse than anything else (probably even much better than things like ReflectionFunction...) so you're probably right, it doesn't make a lot of sense to mention it here. I'll remove it.

guides/hack/05-functions/30-function-references.md
periodic1236, line 152: Nothing too sensitive here. The `HackCallable` class is what I would recommend for internal use.

```suggestion
If you need to determine what a function reference is pointing to, e.g. for
use in logging messages, and you know enough about the expected input and
output formats, HHVM provides the following helpers (but note they are not
well supported and may change):
```

PR #934, opened 2020-10-14
Merged 'break' and 'continue

bball07:


guides/hack/04-statements/32-continue.md
```diff
@@ -4,3 +4,22 @@ A `continue` statement terminates the execution of the innermost enclosing `do`,
 
 Although a `continue` statement must not attempt to break out of a finally block, a `continue` statement can terminate a loop that is
 fully contained within a finally block.
+
+A `break` statement can be used to interrupt the iteration of a loop statement and to break-out to the statement immediately following
```
Wilfred, line 8: Could you add headings to separate the `break` information from the `continue` information?

```suggestion
## `break`

A `break` statement can be used to interrupt the iteration of a loop statement and to break-out to the statement immediately following
```

Likewise we should have a heading at the top for the `continue` documentation.

Wilfred (changes requested):
Great start!

I think we should rename `32-continue.md` to `32-break-and-continue.md`. This will give us a better heading in the sidebar, so users can find the discussion of `break` more easily.

PR #942, opened 2020-10-28
Clarify use case of docker images

fredemmott:


Wilfred (approved):
LGTM. This paragraph is also repeated in README.md, so we might want to delete one or the other.

PR #945, opened 2020-11-02
Fixes comment in echo example

### LINKED ISSUE #944: Typo in documentation for Expressions And Operators: Echo

# Where is the problem?

- The problem is in this [Expressions And Operators: Echo](https://docs.hhvm.com/hack/expressions-and-operators/echo) in the first code section. 

# What is the problem?

- The comment which says what the echo is going to output is slightly wrong. doesn't include the decimal part in the comment.

 - Build ID: HHVM=HHVM-4.79.0:HSL=v4.41.0:2020-10-23T20:15:50+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/expressions-and-operators/echo
 - Page requested at: Mon, 02 Nov 2020 16:47:33 +0000

Prithvirajbilla:
Fixes #944


jjergus:
Good catch, thanks!

PR #948, opened 2020-11-13
example extractor: missing features + migrate everything

jjergus:
Following up on #915
This implements any missing features to make the system usable for all currently existing examples (mainly automatic boilerplate handling), and migrates all the examples.

The number one thing to review here is actually none of the commits here, but the actual build output. Here's the diff between the `build` directory in the master branch vs after this migration: https://gist.github.com/jjergus/ccae9b42ffb39b6847b2ad6ec4fbe5c5

The differences are:
- non-deterministic stuff (async curl etc.)
- file paths in error output, the new system replaces them with `/home/example` automatically
- some `.example.hhvm.out` files that didn't actually match the script output (we were never testing this)
- some changes that I intentionally made (see the first set of commits here)
- namespaced names that occur in various outputs -- the new system generates namespace names automatically based on each example path, which mostly don't match the original namespace names that people wrote manually

The commits here include:
- everything before "implement FilterBase::addBoilerplate()" are random changes that I made for the sole purpose of making the automatic migration easier (e.g. avoiding features that are hard to handle automatically, renaming things to match what would be autogenerated, etc.) -- not much worth reviewing there but they're all trivial so feel free to
- **"implement FilterBase::addBoilerplate()" -- the most important commit to review, implements the missing features from #915**
- "migration script" -- script that did the automated migration, will not be included in the final commit; it's a one-off script so pretty hacky
- "[auto] run migrate.php" -- script output, inlines all example code in .md guides
- 2 more commits with random trivial changes I made manually
- last 3 commits are deleting all `guides/.../*-examples` directories and then running `bin/build.php` to regenerate them (split into 3 commits for easier review) -- these are worth looking at to see how the new auto-extracted example files differ from the original manually written versions:
  - the changes that can be seen in the build output above are also seen here
  - but almost all changes in these commits are in the auto-generated boilerplate parts that are not part of the build output, e.g. the new auto-generated namespace names


src/markdown-extensions/extracted-code-blocks/FilterBase.php
```diff
@@ -169,37 +192,47 @@ public function filter(
           Files::EXAMPLE_TYPECHECKER_OUT,
           Files::TYPECHECKER_EXPECTF,
           Files::TYPECHECKER_EXPECTREGEX,
+          false, // needs example (TODO: see below)
         )
       ) {
-        static::processMissingTypecheckerOutput($hack_file_path);
+        static::processMissingTypecheckerOutput($hack_file_path, false /*TODO*/);
       }
+      /* TODO: Test if this is preferable to the HHVM output.
```
Wilfred, line 200: Should we remove this?

src/markdown-extensions/extracted-code-blocks/FilterBase.php
```diff
@@ -208,16 +241,134 @@ public function filter(
...
+              $$,
+              re"/[^a-zA-Z0-9]+([a-zA-Z0-9])/",
+              $match ==> Str\uppercase($match[1]),
+            ),
+        )
+        |> Str\join($$, '\\');
+      $code = 'namespace HHVM\\UserDocumentation\\'.$ns.";\n\n$code";
+    }
+
+    /* TODO: Uncomment later.
```
Wilfred, line 342: Is this finished?

api-examples/class.AsyncMysqlClient/adoptConnection/001-basic-usage.php.typechecker.expect
```diff
@@ -1 +0,0 @@
-No errors!
```
Wilfred, line 1: Shouldn't this file be deleted entirely? Or am I misunderstanding the GitHub UI here?

Wilfred (approved):
Beautiful! I've left a comment about some commented-out code, but looks really good overall.

src/markdown-extensions/extracted-code-blocks/FilterBase.php
jjergus: Oh, sorry, I forgot to mention this; I'm planning to uncomment this right after this lands, I just didn't want to do it in the same PR since it would make it even harder to review (my test plan relies on comparing old example files with the newly extracted ones, which would be much harder if these extra comments showed up everywhere)...

api-examples/class.AsyncMysqlClient/adoptConnection/001-basic-usage.php.typechecker.expect
jjergus: I verified they're deleted. It says "This file was deleted." in the GitHub UI for me, but that message disappears when you expand the file so it's pretty confusing.

jjergus:
The last update has the TODO part uncommented, so all extracted examples that have any auto-generated boilerplate now contain the warning.

jjergus:
The remaining TODO comments are removed in #949

PR #951, opened 2020-11-20
fix all api-examples

jjergus:
We weren't testing these at all in ExamplesTest. This changes that (the commit "add api-examples to ExamplesTest") and fixes any errors (mostly top-level code, migrated with an automated migration in the first commit).


jjergus:
hhast migration is https://github.com/hhvm/hhast/pull/312

lexidor:
Thank you for migrating these over. I can rebase #911 and write a proper `queryAsync` API example for it.

api-examples/__includes/async_mysql_require_env.inc.php
```diff
@@ -1,14 +1,19 @@
...
+<?hh
 
 namespace Hack\UserDocumentation\API\Examples\AsyncMysql;
 
-require __DIR__."/async_mysql_connect.inc.php";
+use Hack\UserDocumentation\API\Examples\AsyncMysql\ConnectionInfo as CI;
 
-use \Hack\UserDocumentation\API\Examples\AsyncMysql\ConnectionInfo as CI;
+<<__EntryPoint>>
+function async_mysql_skipif(): void {
```
fredemmott, line 8: `.inc` files shouldn't have `<<__EntryPoint>>`s

api-examples/__includes/async_mysql_require_env.inc.php
```diff
@@ -1,14 +1,19 @@
...
+  require __DIR__."/async_mysql_connect.inc.php";
 
-if (!extension_loaded('mysql') || !function_exists('mysqli_connect')) {
-  die('Skip');
-}
-if (!mysqli_connect(CI::$host, CI::$user, CI::$passwd, CI::$db, CI::$port)) {
-  die('Skip');
+  $pool = new \AsyncMysqlConnectionPool(darray[]);
+  try {
+    \HH\Asio\join(
```
fredemmott, line 13: Given callers/includes need modifying anyway, is it practical t oturn this into an async function and await, moving HH\Asio\join() out a level if needed? It'd be good to have no hh\asio\join in these at all

fredemmott (approved):
Thanks; some changes would be good, but fine with those being a separate PR later

jjergus:
Thanks, updated in https://github.com/hhvm/user-documentation/pull/951/commits/94f9877e0d4fa7ac8c7fd473dfcf36a66f763b4d and https://github.com/hhvm/user-documentation/pull/951/commits/ed9102608edf826953939fb6a26cdbc00197af31

I'm planning a separate PR to clean up some of the silly __EntryPoint functions that just call another function, and replace `require` with autoloader where possible.

jjergus:
There are 2 more failures I didn't know how to fix:

```
FAILED: api-examples/class.AsyncMysqlClient/setPoolsConnectionLimit/001-basic-usage.php
001+ 
001- dict\(5\) \{
002-   \["created_pool_connections"\]\=\>
003-   int\(2\)
004-   \["destroyed_pool_connections"\]\=\>
005-   int\(0\)
006-   \["connections_requested"\]\=\>
007-   int\(3\)
008-   \["pool_hits"\]\=\>
009-   int\(\d+\)
010-   \["pool_misses"\]\=\>
011-   int\(\d+\)
012- \}
```

(`AsyncMysqlClient::setPoolsConnectionLimit` doesn't seem to work, the expected exception is not thrown)

```
FAILED: api-examples/class.AsyncMysqlConnectionPool/__construct/001-basic-usage.php
003- Hit idle limit
```

(`idle_timeout_micros` also doesn't seem to work, the expected exception is not thrown here either)

These are not blocking me since all AsyncMysql tests are skipped in Travis anyway, but it would be good to fix them later (also maybe add a MySQL server to the Docker image so we can run those tests in CI).

fredemmott:
FYI I'll probably de-dockerize the pipeline when it moves over to github actions, but I'll look into adding a mysql instance too.

This repo will probably have a pretty distinct github actions pipeline (or two) rather than re-using the generic one that's currently copy-pasted around the place.

jjergus:
2 new commits are migrations from https://github.com/jjergus/hhast/commit/48210dc81ef740aa61c6010b8944e51e630305eb and https://github.com/jjergus/hhast/commit/48ce80a577b717e76d406e3aa84f48a9f94e0ac7

PR #952, opened 2020-11-21
test .example.out files match .expectf/regex

jjergus:
only last 2 commits are relevant, the rest is https://github.com/hhvm/user-documentation/pull/951

### DIFF
--- a/tests/lib/ExpectObj.php
+++ b/tests/lib/ExpectObj.php
@@ -15,7 +15,7 @@ final class ExpectObj<T> extends \Facebook\FBExpect\ExpectObj<T> {
 
   public function toMatchExpectregexFile(string $file): void where T = string {
     $this->toMatchRegExp(
-      \file_get_contents($file),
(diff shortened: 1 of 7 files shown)


tests/lib/ExpectObj.php
```diff
@@ -15,7 +15,7 @@ public function toMatchExpectFile(string $file): void where T = string {
 
   public function toMatchExpectregexFile(string $file): void where T = string {
     $this->toMatchRegExp(
-      \file_get_contents($file),
+      '/^'.\file_get_contents($file).'$/s',
```
fredemmott, line 18: How did this pass before ? :/ Did we have no expectregexp files?

fredemmott (approved):

tests/lib/ExpectObj.php
jjergus: Previously this was only called from `testExamplesTypecheck`, and we have no typechecker expectregex files.

PR #957, opened 2020-12-02
Delete 02-some-basics.md

bball07:
This discussion of precedence is duplicated, because the next page has a full precedence table. Delete it.


fredemmott:
Thanks - is it intentional that you're deleteting the entire file, rather than just the precedence section?

Wilfred:
@fredemmott yep (I suggested this in T65198640), I don't think this page adds anything. 'Sequence points' is a C concept that doesn't exist in Hack. The examples with `++` and `--` are also unhelpful, because Hack no longer allows things like `foo(++$x)`.

@bball07 looks good, but you'll need to add a redirect for this deleted page: https://github.com/hhvm/user-documentation/blob/master/src/Guides.php#L86

PR #958, opened 2020-12-03
migrate api-examples to .md with inline code

jjergus:
This is not as clearly useful as it was for `guides` (#948), since many of the examples are standalone files, but I think it's still preferable to migrate these too:

- we now have a single .md file per API page (class/function/method) instead of a directory with 3+ files per example so it's easier to navigate
- some examples end up with a silly .md file that is just a single code block, but most examples have (or should have) additional description before the code which was previously a separate .md file
- most importantly, this way we take advantage of the same auto-add-boilerplate system as all the examples in `guides` -- you write a code block with exactly what you want to see in the documentation, we add `namespace`, `__EntryPoint` etc. when the example code is extracted

Included in this PR:

- first, replace all `require` statements with `\init_docs_autoloader();` which makes the migration easier (don't have to keep track of how many `/../../../` we need everywhere)
- move the MCRouter skipif logic into a central function, the migration just calls this from the `.skipif` part of all MCRouter examples
- fixes to build code so that it can handle examples both in `guides` and in `api-examples` consistently
- run migration script -- replaces each *directory* in `api-examples` with a single .md file with the example code (+ description + extra files like skipif/expectf/...) inlined
- run build -- extracts all the newly inlined examples into `api-examples/_extracted` (not a great directory name -- we should probably make this behave consistently with `guides`, maybe extract everything into a central directory like `build/extracted-examples/`?)


fredemmott:
- should these be migrated to .hack files first?
- you mentioned the directory name - the 'correct' place would be a subdir of build/scratch/ (for intermediate build artifacts that aren't needed at runtime)
- sorry if covered on previous PRs: it's seems strange that the extracted files are commited - if they move to build/, they won't be.

fredemmott:
More generally:

- is there a change in rendered output?
- is it reasonable to add extra content to these too, or does the rendered structure make things weird?

jjergus:
> should these be migrated to .hack files first?

I was planning to do that after; it should be easier then because the auto-add-boilerplate logic should take care of adding/not adding <?hh as needed.

> it's seems strange that the extracted files are commited - if they move to build/, they won't be.

I think it's important that they stay committed

- we want the example output reviewed (especially changes to existing example output)
- if the example output changes unexpectedly, we want it to be a test failure, not silently replaced during the build (that would make the whole ExamplesTest pointless)
- performance: it would be slow to run all examples during every build (even in parallel like ExamplesTest)

So probably not `build/scratch` but a separate directory under `build` that is committed would be fine?

> is there a change in rendered output?

https://gist.github.com/jjergus/6243da1148ed747a665109251bbee8ec (removed `require`s and some different paths in error backtraces)

> is it reasonable to add extra content to these too, or does the rendered structure make things weird?

The .md file is placed under "Examples" header on each API page, but that would be easy to change if we want other text besides examples there.

fredemmott:
We definitely want the expect files committed; should the .php and .hack files go to a different place, and not be committed?

fredemmott (approved):
eh, this is fine, let's consistently sort out what to do with generated files - guides and examples - later

jjergus:
> We definitely want the expect files committed; should the .php and .hack files go to a different place, and not be committed?

Good question, my arguments don't really apply to php/hack files so we could avoid commiting those; I'm not sure if that's better, feels a bit weird to commit some extracted example-related files but not others...

PR #961, opened 2020-12-09
central directory for extracted examples

jjergus:
`api-examples/_extracted` --> `build/extracted-examples/api`
`guides/...-examples` --> `build/extracted-examples/guides/...`

- now `guides` and `api-examples` behave more consistently
- simplifies build code and `ExamplesTest` (first commit)
- much less clutter (no more generated files) in `guides` and `api-examples`

If you feel that the `build` directory should not contain any committed files, this could also be a new top-level directory, I don't really have an opinion on that -- but it made sense to me that everything generated by `build.php` is now in `build`, whether or not it's committed.


fredemmott (approved):
For future diff readability, could you add `build/extracted-examples/.gitattributes` containing:

```
*.hack linguist-generated=true
*.php linguist-generated=true
```

... and ditto for other extensions?

This might need to be `build/extracted-examples/**/*.hack linguist-generated=true` etc in the top-level gitattributes instead

Once this is done, the PR should do a 'click to show details' on changes to these, rather than attempting to render all the many many files.

jjergus:
> For future diff readability...

There shouldn't be a lot of these files in future diffs, only new examples that people are adding/modifying. It can be useful to review the example files (definitely the output/expect files, not sure about hack/php files but there's probably some value in reviewing that the code was extracted as expected...) so in general it might be better not to hide them?

Probably too late for that but for this PR, you can also look at individual commits and ignore/only spot check the ones labeled [auto].

PR #964, opened 2020-12-29
Consolidate 'some basics' into intro for async section

Wilfred:


Wilfred:
Looks like Travis has updated its composer version, which is why the build has failed. Seems unrelated to these changes.

PR #966, opened 2021-01-05
Use --ignore-platform-reqs to make composer happy with HHVM

Wilfred:


fredemmott:
Composer gives this message in lots of dependency situatations that actually aren't due to using HHVM; this isn't the right fix, and we need to fix the constraints in the dependencies

PR #972, opened 2021-01-15
Build and test on GitHub actions

fredemmott:
I've not yet added the deploy step; for now, this runs along side the TravisCI testing.

Still using Docker as we're going to want to keep building and publishing docker images - both to dockerhub, and to the live AWS site

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -13,6 +13,7 @@ jobs:
         hhvm: [ '4.88' ]
     runs-on: ${{matrix.os}}-latest
     steps:
+      - uses: actions/checkout@v2
       - name: Fetch docker images
         run: |
           docker pull hhvm/hhvm:${{matrix.hhvm}}-latest
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,0 +1,29 @@
+name: Continuous Integration
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '42 15 * * *'
+jobs:
+  build:
+    name: Build and Test
+    strategy:
+      matrix:
+        os: [ ubuntu ]
+        hhvm: [ '4.88' ]
+    runs-on: ${{matrix.os}}-latest
+    steps:
+      - name: Fetch docker images
+        run: |
+          docker pull hhvm/hhvm:${{matrix.hhvm}}-latest
+          docker pull hhvm/hhvm-proxygen:${{matrix.hhvm}}-latest
+      - name: Build
+        run: docker build -t hhvm/user-documentation:scratch -f .deploy/built-site.Dockerfile .
+      - name: Typecheck
+        run: docker run --rm -w /var/www hhvm/user-documentation:scratch hh_server --check .
+      - name: Run tests
+        run: docker run --rm -w /var/www hhvm/user-documentation:scratch vendor/bin/hacktest tests/
+      - name: Lint
+        run: docker run --rm -w /var/www hhvm/user-documentation:scratch vendor/bin/hhast-lint
+      - name: Verify codegen is unchanged
+        run: docker run --rm -w /var/www hhvm/user-documentation:scratch vendor/bin/hh-codegen-verify-signatures src


fredemmott:
Clean run at https://github.com/fredemmott/hhvm-user-documentation/actions/runs/489045605

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+  schedule:
+    - cron: '42 15 * * *'
+jobs:
+  build:
+    name: Build and Test
+    strategy:
+      matrix:
+        os: [ ubuntu ]
+        hhvm: [ '4.88' ]
```
jjergus, line 13: does this need to be "matrix" if it's just 1 element each?

jjergus (approved):

jjergus:
Do we also want to test everything outside of Docker? (since that's what most developers would likely run)

Could use hhvm/actions/hack-lint-test for that

.github/workflows/build-and-test.yml
fredemmott: we could move it to an env section, but I like the consistency with other projects

fredemmott:
> Could use hhvm/actions/hack-lint-test for that

I basically want to do something like this:

```
diff --git a/.github/workflows/build-and-test.yml b/.github/workflows/build-and-test.yml
index 8bf11f31..80fafdbf 100644
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -5,8 +5,20 @@ on:
   schedule:
     - cron: '42 15 * * *'
 jobs:
-  build:
-    name: Build and Test
+  developer:
+    name: Developer Build
+    strategy:
+      matrix:
+        os: [ ubuntu, macos ]
+        hhvm: [ '4.88' ]
+    runs-on: ${{matrix.os}}-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: hhvm/actions/hack-lint-test@master
+        with:
+          hhvm: ${{matrix.hhvm}}
+  dockerized:
+    name: Dockerized Build
     strategy:
       matrix:
         os: [ ubuntu ]
```

... but I'd need to add a 'build' step between composer and typecheck.

I think the best way to deal with this would be:

- split the current action into multiple actions:
  - 'prepare': install composer, hhvm, and dependencies
  - typecheck
  - lint
  - test
 - add a 'verify' one that delegates to typecheck, lint, and test
 - make the current shared action prepare && verify
 - make this project prepare, build, then verify

fredemmott:
Going to merge this as-is but will do that follow-up work next week

fredemmott:
Squashed as f80d50f15d16817d9c86c8f4a3d1b864617c6ae0

jjergus:
From what I remember the composite github action can't actually run other actions :( IIRC that's why we still need the checkout step duplicated in every project, I couldn't make it part of the shared action.

Maybe we can add a parameter to the shared action? e.g. a `build_script` parameter, if non-empty the action would run it between the existing composer and typecheck steps...

```
[ -z "${{inputs.build_script}}" ] || hhvm "${{inputs.build_script}}"
```

fredemmott:
Ah, yeah, I'm not seeing `uses` in the syntax.

Only half joking: perhaps we want a build system for our CI system? Define each step separately, and generate all the practical combinations?

jjergus:
Could work, but is that any better than having a single shared action, where some steps are optional/parametrized?

fredemmott:
Optional/parameterized requires that we think of sufficient options/parameters in advance or keep extending; something composable/composed gives more flexibility - e.g. if we wanted to split the build step up into multiple commands.

jjergus:
True, we could end up with some parameters that are overly specific and not useful, but "build_script" seems pretty reasonable? If we needed more build commands we could always just add them to that script (maybe don't run the build script via `hhvm` like I suggested above, so that it can be a shell script if needed).

Other parameters I was thinking of are "skip_lint" (instead of the current implicit check, which feels a bit unsafe) and "skip_tests" -- feels more reasonable to me than having an "everything-but-lint" action, "everything-but-tests" etc.

PR #973, opened 2021-01-16
Deploy from GitHub Actions

fredemmott:
Straight port of TravisCI stuff.

We should probably do less in Docker, however it's nice that we can quickly switch between 'full' and 'prod' (no dev dependencies etc) with stuff in docker.

Starting with a straight port to unblock moving, then we should:
- avoid docker where not needed
- if we still keep some basic testing in docker, add testing for standard non-docker developer flow too


fredemmott:
Right now this goes all the way up to building a repo-auth image. It is a draft as it does not yet talk to EC2 or DockerHub, and I  don't want to start that at 5pm on a friday :p

PR #980, opened 2021-01-26
Skip mcrouter examples for \HH\Asio\ functions if McRouter extension not present

fredemmott:
Fails on MacOS

Planning to add testing of actual developer flow, so need clean test run
on all platforms :)

### DIFF
--- a/api-examples/function.HH.Asio.mw.md
+++ b/api-examples/function.HH.Asio.mw.md
@@ -26,4 +26,6 @@ async function basic_usage_main(): Awaitable<void> {
     }
   }
 }
+```.skipif
+\Hack\UserDocumentation\API\Examples\MCRouter\skipif();
 ```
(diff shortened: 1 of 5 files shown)


fredemmott:
Related infra issues: #981, #982

jjergus (approved):
`\Hack\UserDocumentation\API\Examples\MCRouter\skipif()` checks that a server is running (so pretty much always fails, I don't think we have it set up anywhere), not just that the extension is present. Maybe we can have a separate one that just does

```hack
  if (!\extension_loaded('mcrouter') || !\extension_loaded('memcached')) {
    echo 'skip';
  }
```

(or just inline that here)?

fredemmott:
I think we'll also need that here; once I've de-dockerized the testing a bit, we should look at adding mysql and memcached servers to the CI env

fredemmott:
Rebased to merge the .skipif into the same commit

PR #990, opened 2021-01-28
Adding documentation for enum class

vsiles:
This is the current documentation for enum classes. Almost everything is already available in the released HHVM, but:
- the feature is not enabled by default at the moment (see section about flags)
- some features are only partially implemented (final / __Sealed / coeffects) and I'll update the doc once everything is released


vsiles:
The error I see is
> Fatal error: Uncaught exception 'HH\InvariantException' with message '/var/www/build/extracted-examples/guides/hack/11-built-in-types/35-enum-class/EnumClassNames.hack is outdated. Run `hhvm bin/build.php` and commit all generated files.' in /var/www/src/markdown-extensions/extracted-code-blocks/VerifyFilter.php:34

however when I log back to my instance and run the build, no new file is modified. Any suggestion ?

vsiles:
Updated the .hhconfig with enum classes flag. Still getting the "outdated" error message but after
```
$  ./vendor/bin/hh-autoload 
$ hhvm bin/build.php
$ git status
```
still reports no changes

vsiles:
Rebased per @fredemmott advices

jjergus:
- The build failure is because there are multiple code blocks with the same file name.
- If you fix that, you'll likely run into issues when code from one block tries to use definitions from a previous code block, because by default each code block is put into its own namespace.
  - To fix that, you can use file names like `EnumClassFull.interface.hack` + `EnumClassFull.implementation.hack` + `EnumClassFull.usage.hack` -- only the part before the first period (`.`) in the file name will be used for the autogenerated namespace name, so these will all end up in the same namespace (so make sure they don't have conflicting definitions, e.g. a function called `main` in multiple of them)
- There is another issue you might run into: You might get "unknown class" and similar *runtime* errors even if the relevant code blocks are in the same namespace -- this is because HHVM depends on an "autoloader" that must be initialized. We automatically put the necessary code at the beginning of each `__EntryPoint` function (you can see it in the autogenerated files), but anything before that function that references definitions from other files will fail.
  - This will hopefully be fixed eventually when HHVM has a "native autoloader" that doesn't need to be explicitly initialized, but in the meantime there's nothing we can do besides trying to work around it :(
  - The most common workaround is to put your `__EntryPoint` function in a separate code block from everything else, e.g. `EnumClassFull.definitions.hack` has all the classes etc. and `EnumClassFull.main.hack` has the entrypoint function.

I'm working on documenting these limitations and workarounds properly somewhere, but for now hopefully this helps...

vsiles:
@jjergus thanks it worked ! And I fixed the type checking errors :p
Now I got an hhvm error because I didn't set the right flag to support HHVM. Just tried to update the ini files. If it is not the right way to do it, I'll contact you later today

jjergus:
You can add .ini options for each example like this: https://gist.github.com/jjergus/4d2d430cd9d752f28ae81b3abdfa42b6

For examples that cannot run without errors (e.g. because they contain definitions that reference definitions from another file) you can also add `noexec` to the header instead of `no-auto-output`.

Sorry we really should have all this documented somewhere :/

jjergus:
One more thing: If you change some example code block in a way that should change its output (e.g. fix an error), you need to delete the example output files before running `bin/build.php` again. 

This is by design because we don't want example outputs to change unintentionally without people noticing, but now I realize it can be pretty annoying so I wonder if we can improve that somehow...

vsiles:
Victory \o/

PR #998, opened 2021-02-13
regenerate expect/output files if example code changes

jjergus:
This makes it so that if you update some example code block, running `hhvm bin/build.php` will automatically update the relevant output/expect files.

I originally intentionally avoided doing this, but it ended up being pretty annoying for anyone working on a new guide and iterating on their example code (they'd have to manually delete the correct output files from the build directory after each change), so it was probably the wrong decision.

The tradeoff here (and the reason why I originally didn't want to do this) is that if someone updates some example code block and *doesn't* expect the output to change (e.g. fixing style, improving readability, making it compatible with latest HHVM...), it is now possible that the output will change without them noticing. However, we still require any changes to output/expect files to be committed, so hopefully it would still be caught later when creating/reviewing a PR.

### DIFF
--- a/src/markdown-extensions/extracted-code-blocks/ExtractFilter.php
+++ b/src/markdown-extensions/extracted-code-blocks/ExtractFilter.php
@@ -28,7 +28,7 @@ final class ExtractFilter extends FilterBase {
   private static keyset<string> $writtenPaths = keyset[];
 
   <<__Override>>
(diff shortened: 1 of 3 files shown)


Wilfred (commented):
Belated LGTM.

We should probably add a chapter to the docs describing this format.

PR #999, opened 2021-02-17
[enum class] Updating constructors to be ready for coeffects

vsiles:
Activating coeffects will require all initializers to be at most
`write_props`. Here, pure is fine for most of them


jjergus:
Thanks!

Should we also mention this in the text of the guide?

vsiles:
There's already a part on the coeffect interaction. I might be a moving target for now so I'd like to leave it like this, and improve it if we get questions :D

PR #1002, opened 2021-02-24
[27-using--a-trait] Update docs for constants in traits

rcpooley:
Updated docs for new constants in traits feature.


Wilfred (approved):
Wording looks fine, but I think a few examples would be really helpful.

jjergus:
Thanks!

PR #1003, opened 2021-02-26
Replace usage of "memorizing" with "memoizing"

quittle:
This documentation is for the `<<__Memoize>>` attribute and presumably is a typo.


jjergus:
Thanks!

Issue #573, opened 2019-01-18
Explicit operator search support


fredemmott:
This probably needs to be hardcoded in the search - searching for the symbols for a particular operator should get you a useful result

PR #1011, opened 2021-03-25
Explicit operator support (#573)

AndrewDiMola:
hard-coded support for many, many operators


fredemmott:
- we definitely need a test for these :)
- a point to consider I'm 50-50 on:
  - I feel that these should be hardcoded as '+' = URL in the controller instead, as it'll be less brittle
  - ... but, the search terms you're mapping to clearly should deliver good results too. Why duplicate work

AndrewDiMola:
@fredemmott agh, sorry. total n00b here. assuming current direction, would that be adding a random sampling of operators -> expected results to https://github.com/hhvm/user-documentation/blob/master/tests/SearchTest.php#L26?

Still new to Hack too so let me know if anything looks weird, or if there's a better grouping, etc.

fredemmott:
Given the indirection of operators -> searching for words -> results, I think it should actually be a complete list. We want the test run to fail if *any* of these operators become unsearchable

fredemmott:
A few of these terms are current page titles, which is probably a bit too specific:

>   const string PRECEDENCE = 'expressions and operators: operator precedence';

perhaps just 'precedence' and make the test look at the content, e.g.
- searching for `+ precedence` gives a page that includes `||`, `>>` - some indication that we have an all-operator precedence table
- searching for `+ precedence` gives page that includes 'tightly' or some other domain-relevant word

>   const string INCREMENTING_DECREMENTING = 'incrementing and decrementing';

- similar to above for what to actual test
- I'd hope that 'increment decrement' would be sufficient and more flexible. If it isn't, we should improve that. This *might* be
  tweaking the scores, tweaking the scoring mechanism (e.g. rather than just doing prefix/suffix match, taking the relative length of the match into account), or - just dropping suffixes like 'ing'

AndrewDiMola:
I was having issues figuring out how to check the body of each returned result for a given query, and after talking with @jjergus, decided to start by using the existing infra for search query tests, with a test for each operator. Also, you were right @fredemmott -- I shouldn't have been coding in queries for the exact titles for operators. That's very brittle and prone to breaking if titles are changed... so I changed some of the constants too. Re: increment / ing (above comment), I think it does work as intended, just need to not use flexible keywords related to the page contents.

Images are in format of New --> Old

image
image

image
image

image
image

Issue #879, opened 2020-06-24
Document how lambdas capture local variables


jjergus:
We should explain they are captured "by value", not "by reference", with specific examples:

- example how `$captured_variable = ...` doesn't change the variable outside the lambda
- example how `Ref` can be used to work around that (or is this a bad suggestion that we shouldn't encourage?)
- example how `$captured_object->var = ...` works

If this is already explained somewhere, we could just link to that, but I couldn't find it anywhere.

--------------------------------

 - Build ID: HHVM=HHVM-4.57.0:HSL=v4.41.0:2020-05-21T18:29:34+0000:422a1f74d5f744078d1a5aba910cd7ced3e6f419
 - Page requested: /hack/functions/anonymous-functions
 - Page requested at: Wed, 24 Jun 2020 18:38:34 +0000
 - Controller: GuidePageController

PR #1012, opened 2021-03-25
Add Pass by Value example (#879)

AndrewDiMola:
Explaining that lambas pass by value

### DIFF
--- a/guides/hack/05-functions/02-anonymous-functions.md
+++ b/guides/hack/05-functions/02-anonymous-functions.md
@@ -24,6 +24,8 @@ echo("\n");
 echo($x); // $x is unchanged; result of 5
 ```
 
+If you need to mutate the reference, use the HSL `Ref` class.
+
 ## Type Inference
 
 Unlike named functions, type annotations are optional on anonymous functions.
--- a/guides/hack/05-functions/02-anonymous-functions.md
+++ b/guides/hack/05-functions/02-anonymous-functions.md
@@ -1,9 +1,27 @@
 Hack supports anonymous functions.
 
+In the example below, the anonymous function, `$f`, evaluates to a function that
+returns the value of `$x + 1`.
+
 ``` Hack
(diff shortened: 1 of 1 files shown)


AndrewDiMola:
unresolved: should we reference https://docs.hhvm.com/hsl/reference/class/HH.Lib.Ref/?

fredemmott:
> unresolved: should we reference https://docs.hhvm.com/hsl/reference/class/HH.Lib.Ref/?

yep :)

fredemmott:
We have autolinking - hopefully this will 'just work':

```
If you need to mutate, use the HSL `Ref` class
```

... rather than needing to fully-qualify or specify the target page.

AndrewDiMola:
> We have autolinking - hopefully this will 'just work':
> 
> ```
> If you need to mutate, use the HSL `Ref` class
> ```
> 
> ... rather than needing to fully-qualify or specify the target page.

worked on my local test!

fredemmott (approved):

Wilfred:
@AndrewDiMola @jjergus  I don't think this is quite right.

```
$x = 5;
$f = $x ==> $x + 1;
$six = $f($x); // pass by value

echo($six); // result of 6
echo("\n");
echo($x); // $x is unchanged; result of 5
```

This example would also work with pass-by-reference, because `$f` isn't modifying anything.

There's two things we should probably cover:

(1) Variables are scoped to the function that they're assigned in.

```
$x = 1;
$f = () ==> { $x = 2; }
$f();

$x; // still 1
```

(2) Closures capture variables at the definition site, not the call site (unlike e.g. JS). (This is a subtlety I only learnt recently!)

```
$x = 1;
$f = () ==> { return $x; }
$x = 2;

$f(); // 1
```

AndrewDiMola:
@Wilfred - Oh I see, in that example, I'm just getting the output of the function which of course is different but not demonstrating the modification behavior of the variable.

PR #1022, opened 2021-04-13
Anon functions 879

AndrewDiMola:
Continuation of https://github.com/hhvm/user-documentation/pull/1012

### DIFF
--- a/guides/hack/05-functions/02-anonymous-functions.md
+++ b/guides/hack/05-functions/02-anonymous-functions.md
@@ -9,22 +9,33 @@ $f = $x ==> $x + 1;
 $two = $f(1); // result of 2
 ```
 
-Anonymous functions pass _by value_, not by reference. This is also true for any
+## Default Functionality: Pass by Value
+Anonymous functions pass _by value_, not by reference, as shown in the example below.
+
+This is also true for any
 [object property](../expressions-and-operators/member-selection) passed to an
 anonymous function.
 
 ``` Hack
-$x = 5;
-$f = $x ==> $x + 1;
+$x = 1;
+$f = () ==> { $x = 2; }
+$f();
 
-$six = $f($x); // pass by value
-
-echo($six); // result of 6
-echo("\n");
-echo($x); // $x is unchanged; result of 5
+$x; // still 1
 ```
 
-If you need to mutate the reference, use the HSL `Ref` class.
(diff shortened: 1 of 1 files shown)


guides/hack/05-functions/02-anonymous-functions.md
```diff
@@ -1,9 +1,40 @@
...
+
 ``` Hack
 $f = $x ==> $x + 1;
 
-$two = $f(1);
+$two = $f(1); // result of 2
+```
+
+## Default Functionality: Pass by Value
+Anonymous functions pass _by value_, not by reference, as shown in the example below.
```
Wilfred, line 13: I know @jjergus used the term "pass by value" in #879, but I think it's clearer to talk about scope. I chatted about this with a few other Hacklang people and I'm not sure we have a standard term for this.

```
$v = vec[];
$f = () ==> { $v[0] = 1; }
$f(); // $v is vec[]

$v = Vector {};
$f = () ==> { $v[0] = 1; }
$f(); // $v is Vector { 1 }
```

Value versus reference semantics depends on the type of the value you're closing over. Similarly an object has reference semantics:

```
$o = new MyObject(5);
$f = () ==> { $o->value = 6; }
$f();

$o->value; // 5
```

The important thing here is that captured variables take the value when the closure is evaluated. This is extremely unusual: only PHP and Hack do this in a way that's observable to user code.

Wilfred (changes requested):
Sorry for the delay in the reviewing this. I've proposed tweaking the wording, let me know what you think.

guides/hack/05-functions/02-anonymous-functions.md
AndrewDiMola: This is still a bit confusing to me. Just speaking out loud, likely wrong understanding, the above example has `vec`, an array, and `Vector`, a collection. We're saying that Hack has pass by value-like behavior (doesn't persist) for `vec[]` and pass by reference-like behavior (persists) for `Vector{}`. 

Then, later, the next example, we have an object, which should also be similar to collection behavior because both are object types, right?  But the reference doesn't persist... because for closures we need to see where the variable is assigned, not called. My head is all sorts of jumbled here.

Wilfred:
```
$v = vec[];
$f = () ==> { $v[] = 1; }
$v[] = 2;

$f(); // $v is vec[2]
```

AndrewDiMola:
Circling back on a new local branch in a new PR

PR #1024, opened 2021-04-26
Update Shapes Doc

icechen1:
Added some extra content in the Shapes doc with some content from our own docs to clarify a a few things about Shapes that may be confusing to new engineers, namely, some additional context around how to create shapes, limitations, and mentioning Shapes::toDict.

Adapted with [@krisjan](https://github.com/kristjan) from some of our internal docs. We're happy to make more contribution like these to make the docs better!

### DIFF
--- a/guides/hack/11-built-in-types/28-shapes.md
+++ b/guides/hack/11-built-in-types/28-shapes.md
@@ -84,7 +84,7 @@ function uses_shape_internally(): void {
 }
 ```
 
-For large shapes, it is often convenient to define a type alias.
+For large shapes, it is often convenient to define a type alias.  This is useful because it creates a contract users of your function are expected to follow.
 
 ``` Hack
 type Server = shape('name' => string, 'age' => int);
@@ -93,6 +93,19 @@ type Server = shape('name' => string, 'age' => int);
 function takes_server(Server $s) {
   // ...
 }
+```
+
+While you can't instantiate these types directly, you can create a my_function_arguments_t by simply defining a shape that has all of the required fields. The typechecker will automatically infer if your shape can be accepted by code() or not.
+
+``` Hack
+$args = shape('name' => 'hello', 'age' => 10);
+$output = takes_server($args); // no error
+
+$args = shape('name' => null, 'age' => 10);
+$output = takes_server($args); // typechecker error: type mismatch
+
+$args = shape('name' => null, 'age' => 10, 'error' => true);
+$output = takes_server($args); // typechecker error: we have an extra field
 ```
 
 Since shapes are copy-on-write, updates can change the type.
@@ -209,4 +222,8 @@ function returns_wrong_shape(): shape('x' => int) {
 
 ## Converting Shapes
 
-You can use `Shapes::toArray` to convert a shape to a `darray`.
+You can use [`Shapes::toArray`](https://docs.hhvm.com/hack/reference/class/HH.Shapes/toArray/) to convert a shape to a `darray`, or You can use [`Shapes::toDict`](https://docs.hhvm.com/hack/reference/class/HH.Shapes/toDict/) to convert a shape to a `dict`.
+
+## Limitations
+
+Some limitations of shapes include only being able to index it using literal expressions (you can't index on a shape using a variable or dynamically formed string, for example), or to provide run-time typechecking, because it is actually just a dict (darray) at runtime.
--- a/guides/hack/11-built-in-types/28-shapes.md
+++ b/guides/hack/11-built-in-types/28-shapes.md
@@ -86,31 +86,36 @@ function uses_shape_internally(): void {
 
 For large shapes, it is often convenient to define a type alias.  This is useful because it promotes code re-use and when the same type is being used, and provides a descriptive name for the type.
 
-``` Hack
+``` shape-values-of-type-define-server.hack
 type Server = shape('name' => string, 'age' => int);
 
 // Equivalent to the previous takes_server function.
-function takes_server(Server $s) {
+function takes_server(Server $s) : void{
   // ...
+  return;
 }
 ```
 
 Any shape value that has all of the required fields (and no undefined fields - unless the shape permits them) is considered a value of type `Server`; the type is not specified when creating the value.
 
 ```shape-values-of-type.hack
+use type \HHVM\UserDocumentation\Guides\Hack\BuiltInTypes\Shapes\ShapeValuesOfTypeDefineServer\Server;
+
 function takes_server(Server $s): void {
   return;
 }
 
 function test(): void {
   $args = shape('name' => 'hello', 'age' => 10);
-  $output = takes_server($args); // no error
+  takes_server($args); // no error
 
   $args = shape('name' => null, 'age' => 10);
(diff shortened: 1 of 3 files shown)


guides/hack/11-built-in-types/28-shapes.md
```diff
@@ -84,7 +84,7 @@ function uses_shape_internally(): void {
 }
 ```
 
-For large shapes, it is often convenient to define a type alias.
+For large shapes, it is often convenient to define a type alias.  This is useful because it creates a contract users of your function are expected to follow.
```
fredemmott, line 87: So does a raw inline definition; there's two major benefits of a type alias:
- having a name for the type that can help thinking about it
- it helps re-use, and making it clear where the same type is being used

guides/hack/11-built-in-types/28-shapes.md
```diff
@@ -95,6 +95,19 @@ function takes_server(Server $s) {
 }
 ```
 
+While you can't instantiate these types directly, you can create a my_function_arguments_t by simply defining a shape that has all of the required fields. The typechecker will automatically infer if your shape can be accepted by code() or not.
```
fredemmott, line 98: I think this is a bit unclear; perhaps:

Any shape value that has all of the required fields (and no undefined fields - unless the shape permits them) is considered a value of  type `Server`; the type is not specified when creating the value.

guides/hack/11-built-in-types/28-shapes.md
```diff
@@ -95,6 +95,19 @@ function takes_server(Server $s) {
...
+
+``` Hack
+$args = shape('name' => 'hello', 'age' => 10);
+$output = takes_server($args); // no error
+
+$args = shape('name' => null, 'age' => 10);
+$output = takes_server($args); // typechecker error: type mismatch
+
+$args = shape('name' => null, 'age' => 10, 'error' => true);
+$output = takes_server($args); // typechecker error: we have an extra field
```
fredemmott, line 108: Perhaps add a `shape('name' => string, 'age' => int', ...)` example?

guides/hack/11-built-in-types/28-shapes.md
```diff
@@ -95,6 +95,19 @@ function takes_server(Server $s) {
...
+While you can't instantiate these types directly, you can create a my_function_arguments_t by simply defining a shape that has all of the required fields. The typechecker will automatically infer if your shape can be accepted by code() or not.
+
+``` Hack
+$args = shape('name' => 'hello', 'age' => 10);
+$output = takes_server($args); // no error
+
+$args = shape('name' => null, 'age' => 10);
+$output = takes_server($args); // typechecker error: type mismatch
+
+$args = shape('name' => null, 'age' => 10, 'error' => true);
```
fredemmott, line 107: replace with a string name so there's one error at a time, rather than both the type mismatch and extra field errors

guides/hack/11-built-in-types/28-shapes.md
```diff
@@ -95,6 +95,19 @@ function takes_server(Server $s) {
 }
 ```
 
+While you can't instantiate these types directly, you can create a my_function_arguments_t by simply defining a shape that has all of the required fields. The typechecker will automatically infer if your shape can be accepted by code() or not.
+
+``` Hack
```
fredemmott, line 100: no space; also please provide a filename, like:

~~~
```values-of-type.hack
~~~

Then, when built, a typechecker output file will be generated; that should ideally be committed, then if something changes with shapes, the build process will detect that the example is no longer valid.

File: guides/hack/11-built-in-types/28-shapes.md [hunk 6]
```diff
@@ -209,4 +222,8 @@ function returns_wrong_shape(): shape('x' => int) {
 
 ## Converting Shapes
 
-You can use `Shapes::toArray` to convert a shape to a `darray`.
+You can use [`Shapes::toArray`](https://docs.hhvm.com/hack/reference/class/HH.Shapes/toArray/) to convert a shape to a `darray`, or You can use [`Shapes::toDict`](https://docs.hhvm.com/hack/reference/class/HH.Shapes/toDict/) to convert a shape to a `dict`.
```
fredemmott, line 225: Dict needs to be preferred , and pre-existing: conversion should be discouraged.

Perhaps:

> Converting shapes to containers is strongly discouraged, however is necessary, this can be done with `Shapes:toDict()`.
>
> On older versions of HHVM, Shapes can also be converted to `darray`s with `Shapes::toArray()`; this should be avoided in new code, as `darray`s are currently an alias for the `dict` type, and will be removed from the language.

guides/hack/11-built-in-types/28-shapes.md
```diff
@@ -209,4 +222,8 @@ function returns_wrong_shape(): shape('x' => int) {
...
+
+## Limitations
+
+Some limitations of shapes include only being able to index it using literal expressions (you can't index on a shape using a variable or dynamically formed string, for example), or to provide run-time typechecking, because it is actually just a dict (darray) at runtime.
```
fredemmott, line 229: (or `darray` on older versions)

guides/hack/11-built-in-types/28-shapes.md
fredemmott: This would also need a filename to be specified for the example that defines the shape, or that definition will be undefined

guides/hack/11-built-in-types/28-shapes.md
fredemmott: Also, throughout: do not specify link targets if they're internal. `Foo::bar()` in backticks will be automatically converted to links.

fredemmott (changes requested):
Thank you so much! A few notes/suggestions inline

guides/hack/11-built-in-types/28-shapes.md
icechen1: oo this is really cool!

guides/hack/11-built-in-types/28-shapes.md
icechen1: I think this is already covered by the section below this one

guides/hack/11-built-in-types/28-shapes.md
icechen1: done, and committed!

icechen1:
@fredemmott updated the PR!

fredemmott (approved):
will merge shortly assuming CI run is clean

fredemmott (changes requested):
```
File "/var/www/build/extracted-examples/guides/hack/11-built-in-types/28-shapes/shape-values-of-type.hack", line 6, characters 23-28:
111
Unbound name: `HHVM\UserDocumentation\Guides\Hack\BuiltInTypes\Shapes\ShapeValuesOfType\Server` (Naming[2049])
```

an `hh_client` run in the top level after a build should pick up this problem;  a filename needs to be specified for the example defining the type, and a `use type` might be needed in the second example to pull it in from the auto-generated namespace

icechen1:
Updated PR once again to fix the typechecker errors, changed the examples to add some fixmes to silence the expected errors with the given snippet

PR #1030, opened 2021-05-17
Remove async block stub that's better documented elsewhere

Wilfred:


fredemmott (approved):
ok with the intent of this, though test needs fixing

Wilfred:
Yeah, I thought it would be uncontroversial, I just wanted CI to check my tests :)

Issue #1032, opened 2021-05-25
Bool constants are NOT capitalized


PeterTheobald:
Please complete the information below:

# Where is the problem?

https://docs.hhvm.com/hack/built-in-types/bool
Page describing built in type bool

# What is the problem?

On the first line it gives the misleading example of bool constants True and False, which cause errors (guess how I know) they should be true and false.

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: HHVM=HHVM-4.110.0:HSL=v4.108.1:2021-05-18T22:29:20+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/built-in-types/bool
 - Page requested at: Mon, 24 May 2021 21:36:03 +0000
 - Controller: GuidePageController

sheldor14:
This is fixed in https://github.com/hhvm/user-documentation/pull/1033

AndrewDiMola:
Closing after merge of #1033. Thanks for letting us know!

PR #1033, opened 2021-05-25
Clearing up the proper use of true/false in hack

sheldor14:
While helping a teammate out with hack (coming from python), we noticed that true and false were capitalized which isn't how hack has those values as. Just trying to make the documentation a bit clearer.


AndrewDiMola:
This is great. Thanks for fixing the docs! Will also close out #1032.

AndrewDiMola (commented):
image

PR #1039, opened 2021-05-31
Fix the link to php documentation

pavlov99:
Verified that project could be built without errors and the link is updated.


Wilfred:
Thanks. As far as I can see the pages are the same, but your URL seems canonical.

PR #1042, opened 2021-06-08
Update various community and resource links

fredemmott:


fredemmott:
image

Manually clicked links to check them :)

jjergus (approved):
At least one GitHub link could be useful somewhere (doesn't necessarily have to be in the footer of every page though). I'd probably link to https://github.com/hhvm/ instead of a specific project.

src/site/controllers/NonRoutableWebPageController.php
```diff
@@ -367,30 +367,28 @@ private function getFooter(): footer {
...
           <div class="footerPanel">
-            <h2>HHVM Community</h2>
+            <h2>Resources</h2>
             <ul>
-              <li><a href="http://hhvm.com">Website</a></li>
               <li><a href="http://hhvm.com/blog">Blog</a></li>
-              <li><a href="https://www.facebook.com/hhvm">Facebook</a></li>
-              <li><a href="https://twitter.com/HipHopVM">Twitter</a></li>
+              <li><a href="http://hhvm.com">Hack Website</a></li>
+              <li><a href="http://hacklang.org">HHVM Website</a></li>
```
lexidor, line 383: I can't edit this 😞 
`Hack Website` and `HHVM Website` are in the wrong order or `http://hhvm.com` and `http://hacklang.org` are in the wrong order.

PR #1043, opened 2021-06-09
use real examples in available-contexts-and-capabilities

jjergus:
The examples are now typechecked, so they're guaranteed to stay up-to-date. There's a valid and invalid example for each capability.

- AccessStaticVariable was renamed to AccessGlobals
- there is also ReadGlobals, but it's only usable with the experimental `readonly` operator so I left it out for now
- I'll add `write_this_props` after it's added to the `hhi`s (in 1--2 weeks)

Manual changes are in guides/hack/13-contexts-and-capabilities/08-available-contexts-and-capabilities.md, everything else is automatically extracted from there.


jjergus:
Screen Shot 2021-06-10 at 11 02 10-fullpage

PR #1045, opened 2021-06-22
String literals update: heredoc, nowdoc, variable substitution

AndrewDiMola:
Addressing feedback to split out 'Variable Substitution' into a subheading and simplify nowdoc and heredoc examples to just show literals, rather than using a whole main function.


guides/hack/02-source-code-fundamentals/19-literals.md
```diff
@@ -86,9 +86,52 @@ Escape sequence | Character name | Unicode character
...
+substitution rules](#variable-substitution)*.
 
+### Heredoc String Literals
+
+A heredoc string literal is a string literal delimited by `<<< id` and `id`. The literal can contain any source character.
+Certain other (and sometimes non-printable) characters can also be expressed as [escape sequences](#string-literals__double-quoted-string-literals).
+
+For example:
+
+```
```
jjergus, line 99: IIRC if you add a filename here (```heredoc.hack) it will wrap the code in an `__EntryPoint` function automatically, but *only in the extracted file* -- the guide will only show the exact part of the code that you have here.

This could be the best compromise -- we'd still have a "real" example that is executed and typechecked, but without having any boilerplate code cluttering the guide text.

jjergus (approved):
It could be nice to include a variable substitution in the examples, something like

```
$var = 42;
$s = <<<   ID
Wow, look at this text!
We can even have a semicolon here! ; or '' or ""!
Variable substitution: $var
ID;
echo ">$s<\n";
```

(same code in both examples -- the example output will demonstrate the difference)

guides/hack/02-source-code-fundamentals/19-literals.md
AndrewDiMola: @jjergus I actually did want to create an automated test for both of these but it looks like they flag incorrectly as errors (I think!) when building. Based on some internal threads multi-line needs to be wrapped in <? ?> to execute correctly VIA cli?

AndrewDiMola: added variable examples

PR #1046, opened 2021-06-25
add section about using the `UNSAFE_CAST` function to silence errors

enetsee:
Added an example of using the `UNSAFE_CAST` function to silence errors and highlighted the key difference and similarities to using `HH_FIXME` comments, specifically:
-  only applies to expressions with errors arising from a type mismatch
- does change the type of the expression in subsequent typechecking
-  _must_ be applied to each sub-expression with a type mismatch 
- has no runtime impact


lexidor:
I believe mistmatch may have been a typo. Line 130.

lexidor:
Something worth noting. Adding a suppression comment above an assignment statement makes the assigned variable assume the type `_`. This means that you inadvertently disable all checks for all uses of this variable. Unsafe cast does not suffer from this fate.

enetsee:
Interesting; I just tried this (so I could tell you which of `Tany` and `Terr` we see) but I still see the return type of `takes_int` (i.e. `int`) in this example:

```
function takes_int(int $i): int {
  return $i;
}

function takes_float(float $i): float {
  /* HH_FIXME[4110] */
  $x = takes_int($i);
  return $x;
}
```

lexidor:
```HACK
function json_decode_mixed(string $_json): mixed {
  return dict['hello' => 'world'];
}

function main(): void {
  $x = json_decode_mixed('{"hello": "world"}');
  /*HH_FIXME[4110]*/
  foreach ($x as $hello => $world) {
    takes_float($hello);
    takes_float($world);
  }
}

function takes_float(float $_): void {}
```
Here is an example where UNSAFE_CAST would be helpful. You could UNSAFE_CAST to a Traversable<string, string> (which is not enforceable). This would give types to `$hello` and `$world`.

lexidor:
> Interesting; I just tried this (so I could tell you which of `Tany` and `Terr` we see) but I still see the return type of `takes_int` (i.e. `int`) in this example:
> 
> ```
> function takes_int(int $i): int {
>   return $i;
> }
> 
> function takes_float(float $i): float {
>   /* HH_FIXME[4110] */
>   $x = takes_int($i);
>   return $x;
> }
> ```

Thanks for teaching me something about suppression comments. I thought Hack, always gave up on inferring a local variable if an error was suppressed. Good to know that Hack will still try.

enetsee:
Got it; so, in the foreach, `$x` get's treated as a `KeyedTraversable<nothing,nothing>` and both `$hello` and `$world` are, consequently, `nothing` in the calls to `takes_float`.

lexidor:
Yep, iterating over an error results in `nothing`. `nothing` can flow anywhere. I am again surprised by the inference. Previous versions inferred the even more unsound `TErr`. Simple subscripting still does infer `TErr`.

```HACK
function json_decode_mixed(string $_json): mixed {
  return dict['hello' => 'world'];
}

function main(): void {
  $x = json_decode_mixed('{"hello": "world"}');
  /*HH_FIXME[4063]*/
  $y = $x['hello'];
  takes_float($y);
}

function takes_float(float $_): void {}
```

enetsee:
That's a lovely example; because array indexing on nullable containers is a common pattern, it get's special cased. The problem is, we haven't really got a sensible type to suggest in general (without complement or subtraction across all types). The upshot is, `Hole`s aren't generated during typing at the moment but we could worst-case it to `nothing` - thanks for sharing!

guides/hack/30-silencing-errors/01-introduction.md
```diff
@@ -120,6 +121,48 @@ strict mode checks in partial files by using the error code in
...
+to `HH_FIXME` comments, the `UNSAFE_CAST` function _does_ change the type of the 
+expression.
+
+### Silencing Errors per Expression
+
+Whilst a single `HH_FIXME` comment will silence all related errors on the 
+proceeding line, the `UNSAFE_CAST` function must be applied to each 
+sub-expression that has a type mismatch.
+
+```
```
fredemmott, line 145: Can you change this to:

~~~
```silencing_errors_per_expression.hack
~~~

... then re-run build.php? That should auto-generate a .ex[pect file, and if anything changes in Hack that makes the behavior of this example change, we'll automatically find out.

File: guides/hack/30-silencing-errors/01-introduction.md [hunk 2]
```diff
@@ -120,6 +121,48 @@ strict mode checks in partial files by using the error code in
...
+To silence an error arising from a type mismatch on a particular expression, 
+add a call to `HH\FIXME\UNSAFE_CAST` with the expression as the first argument,
+an optional (string literal) comment, and explicit type hints indicating the 
+actual type of the expression and the expected type.
+
+The `UNSAFE_CAST` function also **has no runtime effect**. However, in contrast 
+to `HH_FIXME` comments, the `UNSAFE_CAST` function _does_ change the type of the 
+expression.
+
+### Silencing Errors per Expression
```
fredemmott, line 139: Can you move this above the 'with comments' part?

While it's a less general approach, people will often stop reading when they find the first thing that works, so we should make it so they see this first.

fredemmott (changes requested):
Thanks so much, just a few small things about structure :)

fredemmott (approved):

PR #1049, opened 2021-07-05
[enum class] Update about diamond inclusion

vsiles:
Diamond inclusion, if non ambiguous, is now supported

### DIFF
--- a/guides/hack/11-built-in-types/35-enum-class.md
+++ b/guides/hack/11-built-in-types/35-enum-class.md
@@ -245,6 +245,37 @@ enum class E1 : IBox extends E {
 hhvm.hack.lang.enable_enum_classes=1
 ```
 
+### Diamond shape scenarios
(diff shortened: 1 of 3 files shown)


vsiles:
The typechecking error is most probably because the latest release is not out yet.
No rush here, as I'll be on PTO until the end of July. We can check later :D

vsiles:
This is me trying to rebase my PR :D

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -136,6 +136,35 @@ enum class E1 : IBox extends E {
 // only he name is use for ambiguity
 ```
 
+### Diamond shape scenarios
+Enum classes support diamond shaped inheritance as long as there is no ambiguity, like int:
```
AndrewDiMola, line 140: just wondering if you're saying that diamond inheritance (like in this example) only work for built-in types? Everything else is clear!

AndrewDiMola (approved):

PR #1054, opened 2021-07-16
Hhvm 4.102 is lts, not 4.104

lexidor:
Took the start date of the release of hhvm 4.102 instead of the date at which ltsness was announced. Also removed italics for start date with the power of hindsight.


AndrewDiMola:
Looks like @fredemmott would be better suited at reviewing the accuracy of lts support

PR #1064, opened 2021-08-03
Tmp env variable

AndrewDiMola:
image

Probably should have it's own section with other important env variables? But it's a start?


fredemmott (changes requested):
Environment variables need to be a separate section:
- in general, they need to modify `$_ENV`,  not `$_SERVER`
- this particular one is *only* used by the typechecker. HHVM settings have no effect on the typechecker. It must be set in the process environment.

fredemmott:
Also, please squash merge commits when rebasing (git rebase -i is handy)

AndrewDiMola:
yeah, you're right -- I can create a new PR that's more tidy.

AndrewDiMola:
closing for a neater pr

PR #1067, opened 2021-08-05
correct documentation about GC options

fredemmott:


.github/workflows/build-and-test.yml
```diff
@@ -102,47 +102,6 @@ jobs:
           docker tag "$IMAGE_NAME" hhvm/user-documentation:latest
           docker push "$IMAGE_NAME"
           docker push "hhvm/user-documentation:latest"
-      - name: Identify live vs staging environment
```
Wilfred, line 105: What does the workflow have to do with GC options?

Wilfred (approved):
GC changes look good, but I think you've accidentally included some workflow changes.

fredemmott:
I pushed to the wrong branch, that commit is already in

PR #1075, opened 2021-08-24
Enum class rework

### LINKED ISSUE #1072: Example test2 uses wrong show function, thus not illuminating the point of the example

# Where is the problem?

In the section starting with "To access the additional information added by HH\MemberOf, one has to change the function signature in the following way:"

# What is the problem?

The example code in test2 is using "show_name", but the comments are suggesting that it should be using "show_name_from_names", otherwise there would be no error.

 - Build ID: HHVM=HHVM-4.121.0:HSL=v4.108.1:2021-08-16T19:09:48+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/built-in-types/enum-class
 - Page requested at: Wed, 18 Aug 2021 16:43:13 +0000

AndrewDiMola:
* Fixes #1072 
* Rewrote language for clarity
* Moved around sections 
* Split up examples
* Removed (what I consider to be) a redundant example


AndrewDiMola:
(this is my bad for making the other fix on master instead of a branch)

AndrewDiMola:
(need to rework this -- don't review yet)

AndrewDiMola:
cleaned up in https://github.com/hhvm/user-documentation/pull/1077

PR #1077, opened 2021-08-31
refactor of enum class docs

### LINKED ISSUE #1072: Example test2 uses wrong show function, thus not illuminating the point of the example

AndrewDiMola:
* Fixes #1072 
* Rewrote language for clarity
* Moved around sections 
* Split up examples
* Removed (what I consider to be) a redundant example


AndrewDiMola:
ugh, hold on.

AndrewDiMola:
Last 3 commits resolve the errors from the code examples (1., fixing the samples so that the two separated code snippets now share the same namespace; 2., removing old generated files generated from the original mega embedded, tested code block; 3., commenting out intentional error lines so the type-checker doesn't get [rightfully] angry).

This PR revolves around refactoring the page to be more readable and modular, with structure changes and language changes, so I recommend pulling the branch, running locally, and reading the entirety of the enum class docs page.

AndrewDiMola:
@Wilfred -- Let me know what you think of the refactor. I can also try and post images if that's easier? Though, it's a long page...

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -52,19 +69,21 @@ enum class Names : IHasName {
 hhvm.hack.lang.enable_enum_classes=1
 ```
 
-## Declaring a new enum class
+## Accessing enum class constants
```
fredemmott, line 72: What is an enum class constant?

My assumption based on the `class` keyword is that we /also/ supported something like `const FOO = bar;` in here

Perhaps calling this `enum class values` with '(also known as enum class constants)' in the longer text would be clearer?

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -52,19 +69,21 @@ enum class Names : IHasName {
...
+Once declared, enum class constants are accessed like normal enum constants using the `::` operator: `Names::Hello`, `Names::Bar`, ...
 
-* the `enum class` keyword rather than the `enum` keyword
-* that each constant is annotated with its precise type, as in `HasName Hello = ...`
+### Control over enum class constants
 
-Once declared, enum class constants are accessed like normal enum constants using the `::` operator: `Names::Hello`, `Names::Bar`, ...
+Using [coeffects](../contexts-and-capabilities/introduction.md), you can have control over which expressions are allowed as enum class constants.
+
+By default, all enum classes are under the `write_props` context. It is not possible to override this explicitly. The only work around must be a temporary one involving `HH_FIXME`s.
```
fredemmott, line 80: What is 'one' here? We should either explain it clearly (with an example) and say to avoid it and why, or just remove the mention entirely.

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -85,30 +101,68 @@ enum class EC : int {
 hhvm.hack.lang.enable_enum_classes=1
 ```
 
-If we now look at `EC::A`, its type is `HH\MemberOf<EC, int>`.
+The built-in enum type of `E::A` is just `E`. All we know is that constant `A` is declared within the enum: we know nothing of its underlying type.
 
+But if we look at the enum class `EC::A` its type is `HH\MemberOf<EC, int>`. We know that it's declared within the enum class `EC`, with type `int`.
 
-Let's have a look back at `Names::World`, its type is `HH\MemberOf<Names, HasName>`, and `Names::Bar` has type `HH\MemberOf<Names, ConstName>`.
+## Extending an Existing enum class (Inheritance)
```
fredemmott, line 108: The captialization here feels weird - perhaps `enum class` should be in backticks, or Enum and Class should be capitalized too?

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -85,30 +101,68 @@ enum class EC : int {
...
 
-The addition of this type layer is here to give more information about the enumeration, namely its exact type
-(`HasName` vs `IHasName`) and from which enum class it comes from (`Names`).
-`HH\MemberOf` is designed to allow a transparent access to the underlying value: `HH\MemberOf<Names, HasName> <: HasName`.
-This means that this layer can be ignored if one doesn’t need the additional information, and `Names::Hello` can be used directly as an instance of the `HasName` class:
+Enum classes can be composed together, as long as they implement the same base type:
```
fredemmott, line 110: composition is a distinct concept to inheritance, and we should avoid any risk of confusing the two; or perhaps say "because the values are included, this is sometimes referred to as 'enum composition'"

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -85,30 +101,68 @@ enum class EC : int {
...
+  Box<string> Color = new Box('red');
 }
 ```.ini
 hhvm.hack.lang.enable_enum_classes=1
 ```
 
-## Defining function that expects enum class
+In this example, `EExtend` inherits `Age` from `EBase`, which means that `EExtend::Age` is defined.
+
+As with ordinary class extension, using the `extends` keyword will create a subtype relation between the enums: `EExtend <: EBase`.
```
fredemmott, line 126: It's used throughout the docs, especially for newer features, but I think we should avoid the `<:` etc syntax for type relations everywhere - my impression is that only experts in type theory are familiar with it

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -85,30 +101,68 @@ enum class EC : int {
...
+// type error, Y::Color is declared twice, in E0 and in E1
+// only he name is use for ambiguity
+```.ini
+hhvm.hack.lang.enable_enum_classes=1
+```
+
+### Control over inheritance
+
+Enum classes support the `__Sealed` attribute, just like normal classes, which allows for tighter control over extension mechanics.
+However enum classes do not yet support the `final` keyword.
```
fredemmott, line 159: I found a loophole! Maybe worth documenting?

```Hack
<<__Sealed()>>
enum class Foo : int {}
enum class Bar : int extends Foo {}
```

```
Typing[4238] You cannot extend sealed enum class Foo [1]
-> Declaration is here [2]

test.hack:3:12
    1 | <<__Sealed()>>
[2] 2 | enum class Foo : int {}
[1] 3 | enum class Bar : int extends Foo {}
```

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -85,30 +101,68 @@ enum class EC : int {
...
+
+## Defining a Function that expects an enum class
 
-As previously explained, it is completely fine to use enum class constants as an instance of their declared type:
+It's completely fine to use enum class constants as an instance of their declared type.
```
fredemmott, line 163: Like enums, the values of an enum class are subtypes of the enum itself - e.g. `MyEnumClass::MY_VALUE` is a valid parameter for a function expecting a `MyEnumClass`.

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -85,30 +101,68 @@ enum class EC : int {
 hhvm.hack.lang.enable_enum_classes=1
 ```
 
-If we now look at `EC::A`, its type is `HH\MemberOf<EC, int>`.
+The built-in enum type of `E::A` is just `E`. All we know is that constant `A` is declared within the enum: we know nothing of its underlying type.
 
+But if we look at the enum class `EC::A` its type is `HH\MemberOf<EC, int>`. We know that it's declared within the enum class `EC`, with type `int`.
```
fredemmott, line 106: On a first-read-through, this felt very academic, and not practically useful - until I got to 'Defining a function that expects an enum class' below

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -130,7 +184,7 @@ function test1(): void {
 hhvm.hack.lang.enable_enum_classes=1
 ```
 
-To access the additional information added by `HH\MemberOf`, one has to change the function signature in the following way:
+To access the additional information added by `HH\MemberOf`, change the function signature in the following way:
```
fredemmott, line 187: My first question when I got to here is "why would I want to do that?"; I had to scroll back up to the top.

I think the answer is something along the lines of "If you wanted to restrict it to members of the enum that implement the interface, rather than just anything that implements the interface, use `HH\MemberOf`"?

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -139,23 +193,23 @@ function show_name_from_names(\HH\MemberOf<Names, IHasName> $x): string {
...
+  // show_name_from_names(OtherNames::Foo); // error, expected Names but got OtherNames
 }
 ```.ini
 hhvm.hack.lang.enable_enum_classes=1
 ```
 
-
-As explained in the introduction, this feature also allows to write dependently typed code. Let’s consider the more involved code:
+## Motivation: Dependent Typing
+Our main goal with the enum class is to allow more expressivity and replace some generated / boiler plate code: associating enum constants to different subtypes of the base type provides a lightweight form of dependent types that can be used as type-safe abstraction to write generic validator/sanitizer/loggers boiler-plate code.
```
fredemmott, line 212: - "The main goal of enum classes"
- this description feels very expert-focussed. E.g. what is a 'dependent type' ?

fredemmott: For the examples, should the simple one come first? Also the purpose of the Box one is unclear - what advantages does `get(Box::AGE)`  have over `Box::AGE->data` ?

I think this is pre-existing:

> the return value of the get function depends on which constant is passed as an input.

I think 'the return type' is what makes this interesting?

fredemmott (commented):
I don't know how to use enum classes, and my feedback here is mostly coming from that perspective; sorry, in hindsight it's probably not appropriate for a refactor PR, but submitting as comment as I want to keep it stored somewhere.

More broadly along the same lines, we should look at using enum classes in hh-clilib or hack-router for controller parameters (looking at ScriptController for an FB example), then linking/describing them here as an example real-world problem.

fredemmott:
Sorry for the FB-only link: re hh-clilib and hack-router, https://www.internalfb.com/intern/wiki/Scriptcontroller/writing/#defining-arguments is the internal example I'm thinking of

guides/hack/11-built-in-types/35-enum-class.md
Wilfred: @fredemmott this is the same as normal classes I think:

```
Typing[4238] You cannot extend sealed class Animal [1]
-> Declaration is here [2]

x.php:8:7
    4 | 
    5 | <<__Sealed()>>
[2] 6 | class Animal {}
    7 | 
[1] 8 | class Dog extends Animal {}
```

When using `__Sealed`, you need to say which child classes you're allowing.

guides/hack/11-built-in-types/35-enum-class.md
vsiles: No they are not. There is no value of type `MyEnumClass`. For example `E::Age` doesn't have type `E`. It has type `HH\MemberOf<E, Box<int>>` which is a subtype of `Box<int>`. You can consider `E` as a namespace rather than a type.

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -1,28 +1,45 @@
...
-This feature enables more types as base type for enumerations, as long as the type doesn’t have any generics:
-if an enum has type `t` as base type, then its enum constants are bound to values whose type is a subtype of `t`.
+The base type of an _enum class_ can be any type, as long as that type doesn’t contain any generics: if an enum has type `t` as base type, then its enum constants are bound to values whose type is a subtype of `t`.
 
-Our main goal here is to allow more expressivity and replace some generated / boiler plate code: associating enum constants to different subtypes
-of the base type provides a lightweight form of dependent types that can be used as type-safe abstraction to write generic validator/sanitiser/loggers boiler-plate code.
+## Enabling enum classes
 
-Here are a couple of simple examples:
+Pass the following flags to HHVM/Hack:
```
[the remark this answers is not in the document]
vsiles: IIRC we enabled enum class by defaults in Hack and HHVM.

guides/hack/11-built-in-types/35-enum-class.md
vsiles: Let's consider the `Random` enum class above, what I called 'constant" was `Random::X` and `Random::S`. You can call them members maybe ? To go with their type `HH\MemberOf<Random, ...>` ?

guides/hack/11-built-in-types/35-enum-class.md
vsiles: Indeed, let's stick to inheritance (as I've been asked to work on another mean of composition :D)

guides/hack/11-built-in-types/35-enum-class.md
vsiles: "The declared type" here would be `Box<int>`, not `E`

vsiles (commented):
Some comments

guides/hack/11-built-in-types/35-enum-class.md
fredemmott: Yep - I don't mean to say that this is special, just that it might be worth highlighting the fact that enum classes don't support the final keyword can be worked around.

guides/hack/11-built-in-types/35-enum-class.md
fredemmott: Whoops; I thought this would work, but perhaps it should be given as an example and explicitly explain why it is invalid?

```Hack
enum class Foo: string {
  string BAR = 'BAZ';
}

function do_stuff(Foo $value): void {
  var_dump($value);
}

<<__EntryPoint>>
function main(): void {
  do_stuff(Foo::BAR);
}
```

We shouldn't use the word namespace though unless we're very clear that `MyEnumClass\foo` is invalid.

vsiles: An "invalid" example might do the trick. Going into the details why this doesn't work involves explaining lots of subtyping bits that will clutter the doc I'm afraid

PR #1079, opened 2021-09-05
remove spec file and href

AndrewDiMola:
Moving await as an expression spec documentation to internal docs and removing related href


Wilfred (approved):
Looks good. Please add a redirect from await-as-an-expression-spec to await-as-an-expression so old links still go somewhere useful.

AndrewDiMola:
oh, that's a good idea, didn't even think of broken 404s.

PR #1083, opened 2021-09-10
Refactor of keywords / reserved names page

AndrewDiMola:
Refactor of keywords page to be more exhaustive and discourage other built-in types and literals when naming things.  Not sure if there warrants separation between keywords and "contextual keywords"

image


AndrewDiMola:
ref: #1083

lexidor:
Nothing with your refactoring perse, but something that jumped out to me. This list of keywords contains things that have long since been removed from the language, like `and`, `or`, `xor`, `instanceof`. It may be worth looking at them to see if their special behavior as an operator has been completely removed. For example, the following code typechecks and runs without issue.

```HACK
const int and = 6;

function and(bool $a, bool $b): bool {
  return $a && $b;
}

final class and {
  public function and(): int {
    return 7;
  }

  public static function create(): and {
    return new and();
  }
}

<<__EntryPoint>>
function main(): void {
  $and = and::create();
  if (and(true, and + $and->and() === 13)) {
    echo "Is `and` a keyword?\n";
  }
}

```

lexidor:
`null` is missing from the list of built-in types.

AndrewDiMola:
I ended up making a section for deprecated keywords and removing the types, opting to link to the full section instead

image

test.hack
```diff
@@ -0,0 +1,13 @@
+namespace Hack\GettingStarted\MyFirstProgram;
+namespace HH\Lib\Str;
```
lexidor, line 2: This file is declared in two namespaces. The MyFirstProgram and Str namespace.

fredemmott, line 2: This should be `use namespace HH\Lib\Str;` ...

test.hack
```diff
@@ -0,0 +1,13 @@
...
+
+<<__EntryPoint>>
+function main(): void {
```
fredemmott, line 5: This currently defines `HH\Lib\Str\main` which is unwanted

test.hack
```diff
@@ -0,0 +1,13 @@
...
+
+  $x = "Apples Oranges and ";
+  $y = "Bananas";
+  splice($x, $y, 1);
```
fredemmott, line 9: ... and this should be `Str\splice`

AndrewDiMola:
Oh geez @fredemmott @lexidor did not mean to add that file in this commit

test.hack
AndrewDiMola: good feedback! was just messing around and accidentally added when I wasn't paying attention -- removed.

test.hack
AndrewDiMola: good feedback! was just messing around and accidentally added when I wasn't paying attention -- removed.

test.hack
AndrewDiMola: good feedback! was just messing around and accidentally added when I wasn't paying attention -- removed.

AndrewDiMola:
dangers of `git add .` and working too late at night

fredemmott (approved):

Issue #826, opened 2020-03-06
[ HTML renderer ] Output of scripts is sometimes rendered as a programming language


lexidor:
Example page: https://docs.hhvm.com/hack/functions/some-basics

![image](https://user-images.githubusercontent.com/31805625/76092047-05dd4180-5fbf-11ea-997b-90e897e2c429.png)
![image](https://user-images.githubusercontent.com/31805625/76092108-1beb0200-5fbf-11ea-8092-ed23a05664a8.png)

Coffeescript :confused:???

I believe that the language is being sniffed / guessed.

I think the Output should always be plain text.

lexidor:
This example is clearer, since it tries to syntax highlight string literals.

![image](https://user-images.githubusercontent.com/31805625/76092366-9156d280-5fbf-11ea-8f41-f78dd623b649.png)

Issue #988, opened 2021-01-27
Improve syntax highlighting


fredemmott:
We're currently using HLJS's PHP support. We should ideally switch to using the vscode-hack grammar.

An alternative is to switch to the ffp-based highlighting that hh-apidoc usually does.

Wilfred:
Are you aware of examples that are particularly poor, or is this just general maintenance?

PR #1084, opened 2021-09-10
fix output highlighting after highlightjs

AndrewDiMola:
Add a class for output blocks, then revert color to remove keyword highlighting in output text.

Ref: #826 

# Example 1
Screen Shot 2021-09-10 at 12 24 09 AM
Screen Shot 2021-09-10 at 12 24 54 AM

# Example 2
Screen Shot 2021-09-10 at 12 26 47 AM
Screen Shot 2021-09-10 at 12 26 20 AM

# Example 3
Screen Shot 2021-09-10 at 12 27 15 AM
Screen Shot 2021-09-10 at 12 28 09 AM

### DIFF
--- a/sass/codeBlocks.scss
+++ b/sass/codeBlocks.scss
@@ -25,6 +25,6 @@ p + .highlight {
   }
 }
 
-.output-block span{
+.output-block span {
   color: inherit;
 }
--- a/sass/codeBlocks.scss
+++ b/sass/codeBlocks.scss
@@ -24,3 +24,7 @@ p + .highlight {
     font-size: 60%;
   }
 }
+
+.output-block span{
+  color: inherit;
+}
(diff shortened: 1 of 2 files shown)


sass/codeBlocks.scss
```diff
@@ -24,3 +24,7 @@ p + .highlight {
     font-size: 60%;
   }
 }
+
+.output-block span{
```
fredemmott, line 28: space before `{`

fredemmott (approved):

fredemmott:
On a tangent: ideally we'd replace highlightjs with something that actually supports Hack. https://github.com/hhvm/hhast-inspect has the perfectly accurate but labor-intensive approach (there is a distinct CSS class for each AST node type). 

Re-using the syntax highlighting definition from https://github.com/slackhq/vscode-hack would be technically less accurate, but probably better in practice.

AndrewDiMola:
Yeah, I did see #988 and agree this is more covering a band-aid with a band-aid. Proper fix is more time intensive and on my todos when I can spend more time on utility.

Issue #1017, opened 2021-04-08
Document removal of varray/darray


fredemmott:
Please complete the information below:

# Where is the problem?

"PHP arrays" mentions

# What is the problem?

It does not document that they are now equivalent to vec/dict; they should probably be treated like other historical "you shouldn't use this, but might come across it and need to know" parts of the docs now rather than removed completely.

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: HHVM=HHVM-4.102.0:HSL=v4.94.0:2021-03-30T21:17:53+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/arrays-and-collections/introduction
 - Page requested at: Thu, 08 Apr 2021 15:22:01 +0000
 - Controller: GuidePageController

PR #1085, opened 2021-09-10
document php arrays alias and clean up code samples

AndrewDiMola:
In commit:
* Clean up code samples across docs using darray or varray
* Add FYI on alias behavior and push towards Hack Arrays

Open questions:
* Anything to change here?: https://docs.hhvm.com/hack/silencing-errors/error-codes#4045-array-without-type-parameter
* Or here? https://docs.hhvm.com/hack/built-in-types/shapes#converting-shapes ( +limitations section below )

ref: #1017


AndrewDiMola:
@Wilfred any thoughts here? Cleaned up the PHP arrays stuff in this PR.

sass/test.hack
```diff
@@ -0,0 +1,28 @@
...
+}
+
+<<__EntryPoint>>
+function run(): void {
+  f(new C(new Animal()));
+  f(new C(new Cat())); // accepted
+
+  g(vec[new Animal(), new Animal()]);
+  g(vec[new Cat(), new Cat(), new Animal()]); // arrays are covariant
+}
```
Wilfred, line 28: I'm guessing you didn't mean to commit this? It doesn't look like this file is used anywhere.

If you want to demonstrate covariance of arrays, the easiest way is probably to contrast it with `Vector`, where we can treat a `vec<Dog>` as a `vec<Animal>` but we can't treat as `Vector<Dog>` as a `Vector<Animal>` and use it as a `Vector<Dog>` again after.

```
class Animal {}
class Dog extends Animal {
  public function bark(): void {}
}
class Cat extends Animal {}

function add_item_vec(vec<Animal> $items): vec<Animal> {
  // vec is a value type, so this creates a copy.
  $items[0] = new Cat();
  return $items;
}

function add_item_vector(Vector<Animal> $items): Vector<Animal> {
  // Vector is a reference type, so this modifies the original.
  $items[0] = new Cat();
  return $items;
}

function demo(): void {
  $dogs = vec[new Dog()];
  add_item_vec($dogs);
  $dogs[0]->bark(); // this is fine because we didn't modify $dogs.

  $dogs2 = Vector {new Dog()};
  add_item_vector($dogs2);
  $dogs2[0]->bark(); // hack correctly complains $dogs2 is now a Vector<Animal>.
}
```

guides/hack/03-expressions-and-operators/75-comparisons.md
```diff
@@ -30,7 +30,7 @@ Arrays are compared by length, then by lexicographic ordering on
 elements:
 
 ``` Hack
-varray[5] < varray[1, 2]   // true (LHS has fewer elements)
+vec[5] < vec[1, 2]   // true (left-hand side has fewer elements)
```
Wilfred, line 33: Nice, I like the reduction in jargon :)

sass/test.hack
AndrewDiMola: ughh again with the test file, sorry

guides/hack/03-expressions-and-operators/75-comparisons.md
AndrewDiMola: thanks! a lot of instances of LHS which took me forever to realize was slang...

PR #1090, opened 2021-09-17
Update to HHVM 4.127

fredemmott:
- remove documentation of partially abstract type constants (removed from the language
- Use 'UTC' timezone instead of 'Etc/UTC'
- Update to HHVM 4.127


AndrewDiMola (approved):
Looks good.

Q: How do we manage API pages and methods like these when features are removed.

https://docs.hhvm.com/hack/reference/class/ReflectionTypeConstant/isAbstract/

fredemmott:
> Q: How do we manage API pages and methods like these when features are removed.


They'll be removed/updated as they're removed/updated from the runtime/HHI.

If people need solid documentation for old versions, currently we require that they grab an old version of the docs (e.g. via docker)

PR #1091, opened 2021-09-29
Section for Experimental features + readonly docs

AndrewDiMola:
New section for experimental features and for the `readonly` experimental keyword. Because of how the site is built, adding the section alongside the API reference is pretty tricky: the pages won't generate as is in the sidebar or render unless they're part of one of the product folders ("hack", "hhvm"), so I moved it as a new section & folder in /hack.

Also: Tightened up language in the readonly keyword file but also prefer someone to scan it for accuracy. Intentionally left out some implementation details / talk.

image

Q: Is this just FYI or will people be able to actually use this functionality in prod hhvm if the attribute is included?

FYI / Q: Also removed https://docs.hhvm.com/hsl-experimental/reference/ which only renders on the /home page (hack + hhvm) but not /hack individually. It's related to: https://github.com/hhvm/hsl-experimental/. Keep? Remove? Either way, should be on both site controllers or neither site controllers.

FYI / Q: Because these are experimental, assume the code sample checker fails to pass these examples, right? Was having issues... so I just marked examples in the  ``` Hack format instead of using the auto-tested example system, but when / if we merge, these tests should be converted.

### DIFF
--- a/guides/hack/51-experimental-features/02-read-only.md
+++ b/guides/hack/51-experimental-features/02-read-only.md
@@ -42,6 +42,9 @@ function test(readonly Bar $x) : void {
 }
 ```
 
+## How is it different from Coeffects?
+[Coeffects](https://docs.hhvm.com/hack/contexts-and-capabilities/available-contexts-and-capabilities) affects an entire function (and all the functions it calls), whereas readonly affects values / expressions.
+
 ## Applications
 This is a list of all of the valid locations of the keyword.
 
@@ -121,7 +124,7 @@ function foo(): void {
 ```
 
 ### Functions / Methods
-`readonly` can appear as a modifier on instance methods.
+`readonly` can appear as a modifier on instance methods, signaling that `$this` is readonly.
 
 ``` Hack
 class C {
@@ -171,8 +174,8 @@ function returns_mutable(readonly Foo $x): Foo {
 ### Closures
 A function type can be marked readonly: `readonly function(T1): T`.
 
-## Interactions with [Context and Capabilities](https://docs.hhvm.com/hack/contexts-and-capabilities/available-contexts-and-capabilities)
-If your function only has the `ReadGlobals` capability (i.e. is marked `read_globals` or `policied`) it can only access class static variables if they are wrapped in a readonly expression:
+## Interactions with [Coeffects](https://docs.hhvm.com/hack/contexts-and-capabilities/available-contexts-and-capabilities)
+If your function only has the `ReadGlobals` capability (i.e. is marked `read_globals`) it can only access class static variables if they are wrapped in a readonly expression:
 
 ``` Hack
 function read_static()[read_globals]: void {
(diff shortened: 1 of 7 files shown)


guides/hack/51-experimental-features/02-read-only.md
```diff
@@ -0,0 +1,213 @@
...
+  $y = readonly $x;
+}
+```
+
+### Functions / Methods
+`readonly` can appear as a modifier on instance methods.
+
+``` Hack
+class C {
+  public readonly function bar(): Bar {
```
fredemmott, line 128: I guess this is more about feature design than docs: any note we can add on 'readonly' vs pure or other coeffects here?

guides/hack/51-experimental-features/02-read-only.md
```diff
@@ -0,0 +1,213 @@
...
+function returns_mutable(readonly Foo $x): Foo {
+  return $x; // error, $x is readonly
+}
+```
+
+### Closures
+A function type can be marked readonly: `readonly function(T1): T`.
+
+## Interactions with [Context and Capabilities](https://docs.hhvm.com/hack/contexts-and-capabilities/available-contexts-and-capabilities)
+If your function only has the `ReadGlobals` capability (i.e. is marked `read_globals` or `policied`) it can only access class static variables if they are wrapped in a readonly expression:
```
fredemmott, line 175: `policied` is not currently documented

guides/hack/51-experimental-features/02-read-only.md
```diff
@@ -0,0 +1,213 @@
...
+  return readonly new Foo();
+}
+
+function test(): void {
+  $x = readonly returns_readonly(); // this is required to call returns_readonly()
+}
+```
+
+## Converting to non-readonly
+Sometimes you may encounter a readonly value that isn’t an object (i.e. a readonly int, due to the deepness property of readonly). In those cases, instead of returning a readonly int, you’ll want a way to tell Hack that the value you have is actually a value type. You can use the function `HH\Readonly\as_mut()` to convert any primitive type from readonly to mutable.
```
fredemmott, line 200: s/i.e./e.g./

Are objects not supported, or just a very bad idea?

fredemmott (approved):

guides/hack/51-experimental-features/02-read-only.md
lexidor: I happen to know this. `as_mut()` validates that the argument is a valid type.https://github.com/facebook/hhvm/commit/8114b9f18c28d754fc48b17d64689a57ead44818

guides/hack/51-experimental-features/02-read-only.md
AndrewDiMola: Added a note on coeffects affecting the full function scope v. readonly only affecting expressions / values.

guides/hack/51-experimental-features/02-read-only.md
AndrewDiMola: Adding note that it's usage is only for primitive types and value-type collections of primitive types

PR #1095, opened 2021-10-07
Fix classes, interfaces, methods, traits links in HSL sidebar

AndrewDiMola:
Fix an issue where the table of contents only ever links to /hack reference documentation. Check the HREF in the embedded images or test on the live site for for context.

## Current Docs
Screen Shot 2021-10-07 at 2 32 22 PM

## Proposed Fix
Screen Shot 2021-10-07 at 2 20 47 PM
ed

### DIFF
--- a/src/APINavData.php
+++ b/src/APINavData.php
@@ -12,7 +12,7 @@ namespace HHVM\UserDocumentation;
 
 final class APINavData {
   private APIIndex $index;
-  private string $product;
+  private APIProduct $product;
 
   private function __construct(APIProduct $product) {
     $this->index = APIIndex::get($product);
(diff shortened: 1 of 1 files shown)


src/APINavData.php
```diff
@@ -12,8 +12,11 @@
 
 final class APINavData {
   private APIIndex $index;
+  private string $product;
```
fredemmott, line 15: Is the `APIProduct` enum usable here? It be best to keep it as strictly typed as possible until the last momenet, casting to string if necessary when creating the URL - though AIUI it shouldn't need an explicit cast.

AndrewDiMola: really good point -- resolves fine. merging in with that change.

PR #1096, opened 2021-10-07
Document basics of encoding/locale Str\

fredemmott:
I've not documented creating/modifying locales, as that's really
documenting locales, not string behaviors; that said, I don't know
where the best place to document locales is. Maybe it's here; for now
I've aimed to document just enough of `Locale\` for people to get
started with HSL strings.

While I'm here
- remove `__toString()` example and docs as `__toString()` is deprecated
- remove definition of 'numeric string'; there isn't a consistently used
  definition
- reword

This is a bit less precise, but I think the precision here took away
from readability for new users.


guides/hack/11-built-in-types/16-string.md
```diff
@@ -1,45 +1,105 @@
...
+# Functions without a locale parameter
+
+In HHVM 4.130 and above, this is the default locale used by the `Str\` functions that do not take an explicit
+locale - that is, `Str\foo($bar)` is equivalent to `Str\foo_l(Locale\bytes(), $bar)`. In prior versions, the
+active native/libc locale would be used instead; `Str\foo_l(Locale\get_native(), $bar)` can be used instead to
+emulate the prior behavior.
+
+This behavior was changed as:
+- functions such as `Str\format()` were frequently used to to generate machine-readable strings, leading to subtle
+  bugs when locale waas changed. For example, `Str\format('%.2f', 1.23)` could return eith '1.23' or '1,23'.
```
lexidor, line 83: waas and eith

lexidor:
An informative read. Documenting the Locale objects and their methods might be generated from the hhvm docblocks. If they are helpful, we could link to them.

AndrewDiMola (approved):

PR #1099, opened 2021-10-12
Proposal: Add collapse for non-active nav groups

AndrewDiMola:
Was talking with @Wilfred about the benefits of a collapsible sidebar because there are so many elements on the left-hand side that it's sort of an endless scroll. Open to suggestions / any thoughts on current v. proposed. For section readability, considering adding horizontal rows or alternating backgrounds. 

Also... in sidebar, $orange is hard for me to read, so I'm using $blue instead.

Having concerns about hiding too much though. What're all of your thoughts.

## Proposed / Sidebar
Screen Shot 2021-10-11 at 8 30 01 PM

## Proposed / Reference Sidebar
image

## Curent Sidebar
image

### DIFF
--- a/sass/navigation.scss
+++ b/sass/navigation.scss
@@ -22,9 +22,8 @@
       .subList {
         display: block;
       }
-
-      h4 > a {
-        text-transform: uppercase;
+      h4 {
+        background-color: #DADADA;
       }
     }
 
@@ -45,7 +44,7 @@
         font-size: 16px;
 
         a {
-          color: $darkblue;
+          color: $orange;
           padding-bottom: 10px;
           padding-top: 10px;
         }


Wilfred:
Wow, this is a huge improvement! It's so much easier to see what's available in the sidebar here. It also makes the flaws of the grouping (e.g. Types vs Built In Types) more obvious, but that's to be expected :)

sass/navigation.scss
```diff
@@ -11,13 +11,28 @@
...
     position: relative;
 
+    .navGroup.navGroupActive {
+      .subList {
+        display: block;
+      }
+
+      h4 > a {
+        text-transform: uppercase;
+      }
```
Wilfred, line 28: Previously we had a sort of TITLE CASE thing going on: upper case, but smaller font and a different colour.

It's less obvious to me that the upper case section is active. How about reusing the hover style?

```suggestion
      a {
        background-color: #DADADA;
      }
```

sass/navigation.scss
```diff
@@ -27,11 +42,10 @@
 
       h4 {
         font-family: $base-font;
-        font-size: 12px;
-        text-transform: uppercase;
+        font-size: 16px;
 
         a {
-          color: $orange;
+          color: $darkblue;
```
Wilfred, line 48: It looks OK, but we do use orange for links everywhere else on the site. How bad does it it look if we use orange here too?

If we do use darkblue here, we should probably use the same colour for the little arrow on the active item in the sidebar.

sass/navigation.scss
AndrewDiMola: picture reference: 
image

PR #1107, opened 2021-11-05
Rework devcontainer to use proxygen server

fredemmott:
This means that `hhvm` is the actual docker command; the bad news is that if
hhvm crashes, the container goes down.

I'm doing this anyway as there doesn't appear to be support in
devcontainer.json for a separate long-running command - only via docker
compose, which:

- feels like overkill
- appears not to support devcontainer variables like
  containerWorkspaceFolder; this means that it can't work with GitHub
  codespaces, as the `workspaceMount` devcontainer.json option is not
  supported on codespaces

The lack of support for `workspaceMount` is also why there's the fun
messing around with symlinking /var/www and configuration files.

### DIFF
--- a/.deploy/init.sh
+++ b/.deploy/init.sh
@@ -15,6 +15,10 @@ echo "** Installing apt dependencies"
 apt-get clean
 apt-get update -y
 
+# Some environments (e.g. VSCode containers) will copy the exterior locale
+# settings, which can break things if the current locale isn't usable in the
+# container; using the `C` locale makes sure that the `locales` package
+# post-install succeeds.
 LC_ALL=C apt-get install -y ruby php-cli zip unzip locales
 
 echo "** Updating locales"
--- a/.deploy/init.sh
+++ b/.deploy/init.sh
@@ -7,13 +7,15 @@ if [ ! -e /docker_build ]; then
   exit 1
 fi
 
+export DEBIAN_FRONTEND=noninteractive
+
 echo "** Installing apt dependencies"
 # This is done by the dockerfile, but the intermediate issue can be cached, so do
 # it again here.
 apt-get clean
 apt-get update -y
 
-apt-get install -y ruby php-cli zip unzip locales
+LC_ALL=C apt-get install -y ruby php-cli zip unzip locales
 
 echo "** Updating locales"
 locale-gen en_US.UTF-8
--- a/.deploy/init.sh
+++ b/.deploy/init.sh
@@ -2,35 +2,17 @@
 
 set -ex
 
-if [ ! -e /docker_build ]; then
+if ! [ -e /.docker_build ]; then
   echo "This script should only be ran from a Dockerfile"
   exit 1
 fi
 
-export DEBIAN_FRONTEND=noninteractive
-
-echo "** Installing apt dependencies"
-# This is done by the dockerfile, but the intermediate issue can be cached, so do
-# it again here.
-apt-get clean
-apt-get update -y
-
-# Some environments (e.g. VSCode containers) will copy the exterior locale
-# settings, which can break things if the current locale isn't usable in the
-# container; using the `C` locale makes sure that the `locales` package
-# post-install succeeds.
-LC_ALL=C apt-get install -y ruby php-cli zip unzip locales
-
-echo "** Updating locales"
-locale-gen en_US.UTF-8
-
-echo "** Installing composer"
-mkdir /opt/composer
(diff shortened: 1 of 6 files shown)


.deploy/init.sh
```diff
@@ -7,13 +7,15 @@ if [ ! -e /docker_build ]; then
...
+export DEBIAN_FRONTEND=noninteractive
+
 echo "** Installing apt dependencies"
 # This is done by the dockerfile, but the intermediate issue can be cached, so do
 # it again here.
 apt-get clean
 apt-get update -y
 
-apt-get install -y ruby php-cli zip unzip locales
+LC_ALL=C apt-get install -y ruby php-cli zip unzip locales
```
fredemmott, line 18: vscode executes stuff a bit differently, including implicitly copying locale around. Installing the `locales` package will fail if the locale is set to `en_US.UTF-8` before we have ran `locale-gen en_US.UTF-8`... which we can't do earlier as it's provided by the locales package.

fredemmott:
https://github.com/microsoft/vscode-remote-release/issues/4093 and https://github.com/microsoft/vscode-remote-release/issues/3034 are the primary design constraints leading to this solution :(

.deploy/init.sh
Atry: Shall we move `apt-get install` to the `Dockerfile`, to allow docker caching across Codespaces?

Atry (approved):
Overall looks good with some minor questions. See comments.

Atry:
By the way, shall we add a step in the Github Actions to test the devcontainer?

For example: https://github.com/stuartleeks/devcontainer-build-run

.deploy/init.sh
fredemmott: This script is used by a few different dockerfiles at the moment, though it's probalby worth splitting the stuff that doesn't require a copy of the source into a separate script which can then be called from the dockerfile and cache

.devcontainer/devcontainer.json
```diff
@@ -4,7 +4,16 @@
...
-  "image": "hhvm/hhvm:4.134-latest",
+  "build": {
+    "dockerfile": "Dockerfile",
+    "args": {
+      "WORKSPACE": "${containerWorkspaceFolder}"
+    }
+  },
+  "runArgs": [
+    "--env", "LC_ALL=C"
+  ],
```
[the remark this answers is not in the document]
fredemmott: This way it's only passed to the HHVM server process, not terminals, build commands etc, which should have the UTF-8 locale

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM hhvm/hhvm-proxygen:4.134-latest
+
+ARG WORKSPACE
+
+RUN rm -rf /var/www; ln -s ${WORKSPACE} /var/www
+RUN ln -sf /var/www/hhvm.dev.ini /etc/hhvm/site.ini
+RUN touch /docker_build
```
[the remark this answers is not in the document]
fredemmott: Looks like it could use /.dockerenv nowadays

.deploy/init.sh
fredemmott: ... which won't work as that script isn't available until the source is mounted.

.devcontainer/Dockerfile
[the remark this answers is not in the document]
fredemmott: Going to make it a dockfile, but stick with touching this. The methods in that stackoverflow post are valid when running, but not when building.

fredemmott:
Either fixed or responded, except for:

> By the way, shall we add a step in the Github Actions to test the devcontainer?
> 
> For example: https://github.com/stuartleeks/devcontainer-build-run

Given that this is re-using our existing testing container scripts where possible, I don't think it adds much, especially as it's a third-party 'pet project' - and it limits experimentation with a better compose-based. Definitely worth adding if GitHub introduce a more complete official one in the future.

.deploy/init.sh
```diff
@@ -2,36 +2,18 @@
 
 set -ex
 
-if [ ! -e /docker_build ]; then
+if ! [ -e /.docker_build ]; then
   echo "This script should only be ran from a Dockerfile"
   exit 1
 fi
```
Atry, line 8: `.dockerenv` does not exist on buildkit but the cgroup approach should work.


See https://stackoverflow.com/a/69860299/955091

Atry (approved):

PR #1109, opened 2021-11-09
link clearing capabilities to memoize attributes

### LINKED ISSUE #674: Link __Memoize attribute to HH/clear_static_memoization
The documentation on special internal attributes describes `<<__Memoize>>` and `<<__MemoizeLSB>>`. That might also be a helpful place to mention the existence of https://docs.hhvm.com/hack/reference/function/HH.clear_static_memoization/ and the LSB equivalent because they're useful for testing/working with memoized functions.

AndrewDiMola:
* @ryangreenberg had a great idea to link the HH clearing functions for `__Memoize` and `__MemoizeLSB` on the [attributes page](https://docs.hhvm.com/hack/attributes/predefined-attributes).
* Fixes #674

### DIFF
--- a/guides/hack/20-attributes/07-predefined-attributes.md
+++ b/guides/hack/20-attributes/07-predefined-attributes.md
@@ -221,7 +221,9 @@ The types of the parameters are restricted to the following: `null`, `bool`, `in
 
 The interface type `IMemoizeParam` assists with memoizing objects passed to async functions.
 
-To clear the cache and avoid potential memory leaks, use `HH\clear_static_memoization`.
+You can clear the cache with `HH\clear_static_memoization`. This should only be used **in tests** where:
+- the component being tested is meant to be immutable/idempotent for the entire request
+- the test needs to cover multiple initial states, where only one would truly be reachable in a single request
 
 ### Exceptions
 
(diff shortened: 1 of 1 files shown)


guides/hack/20-attributes/07-predefined-attributes.md
```diff
@@ -221,6 +221,8 @@ The types of the parameters are restricted to the following: `null`, `bool`, `in
 
 The interface type `IMemoizeParam` assists with memoizing objects passed to async functions.
 
+To clear the cache and avoid potential memory leaks, use `HH\clear_static_memoization`.
```
fredemmott, line 224: This should not be done in production code; it's really only intended for unit tests, and only if the tests require breaking the isolation of 'a unit'. It shouldn't generally be used.

`<<__Memoize>>` is frequently abused as a 'run at most one time', so clearing caches can really break stuff.

fredemmott (changes requested):
I'm fine with the content here, but it does need some more warnings :)

guides/hack/20-attributes/07-predefined-attributes.md
AndrewDiMola: hmm do we have a way of embedding macro blocks? something like:
image

fredemmott: if you grep -i src/ for `fbonly` you'll see one special case of this, which could be generalized

PR #1111, opened 2021-11-11
[RFC] Move readonly documentation to non experimental section ahead of release, improve docs

jamesjwu:
Readonly is getting ready for release, so I'm preparing the documentation for when the time comes to release the feature. I've updated the docs to have five sections with code examples (mainly just splitting and elaborating on what existed already). 

I couldn't really find a good category to put readonly, since it isn't directly a type, and also isn't just an expression, statement, control flow, etc. Instead it's an advanced feature for specific use cases, so I made a new section called Advanced Features to put it in. 

Not sure if this makes sense, would love feedback on it and also on the different sections.

image
image
image
image
image
image
image
image


jamesjwu:
(For what it's worth, coeffects, reified generics, and enum classes all feel like features that could go in this "advanced" section as well)

Atry:
Shall we compare `readonly` fields with `<<__Memoized>>` nullary methods? When should we use `readonly` fields instead of  `<<__Memoized>>` nullary methods?

guides/hack/16-readonly/02-syntax.md
```diff
@@ -0,0 +1,96 @@
...
+
+``` Hack
+function readonly_closure_example(): void {
+  $x = new Foo();
+  $f = readonly () ==> {
+    $x->prop = 4; // error, $x is readonly here!
+  };
+}
+```
+One way to make sense of this behavior is to think of closures as objects with an __invoke function (which is how HHVM implements all closures), and whose properties are the values it captures. A readonly closure is then defined as a closure whose __invoke function is annotated with readonly.
```
fredemmott, line 92: Use backticks for `__invoke`

Also, not sure if we sohuld be mentioning `__invoke` here - while it's a clear way to explain, `__invoke` itself is a PHPism

guides/hack/16-readonly/06-advanced-semantics.md
```diff
@@ -0,0 +1,59 @@
...
+  $ro_regular_f(); // error, $ro_regular_f is a readonly reference to a regular function
+  $ro_ro_f = readonly $ro_f; // readonly (readonly function(): T)
+  $ro_ro_f(); // safe
+}
+```
+
+## Converting to non-readonly
+Sometimes you may encounter a readonly value that isn’t an object (i.e. a readonly int, due to the deepness property of readonly). In those cases, instead of returning a readonly int, you’ll want a way to tell Hack that the value you have is actually a value type. You can use the function `HH\Readonly\as_mut()` to convert any primitive type from readonly to mutable.
+
+Use `HH\Readonly\as_mut()` strictly for primitive types and value-type collections of primitive types (i.e. a vec of int).
```
fredemmott, line 46: e.g.

src/Guides.php
```diff
@@ -127,18 +127,19 @@ public static function getGuidePageRedirects(
...
-	  'some-basics' => tuple('functions', 'introduction'),
-	  'calling-a-function' => tuple('functions', 'format-strings'),
-	],
         'generics' => dict[
           'constraints' => tuple('generics', 'type-constraints'),
           'open-and-closed-generic-types' => tuple('generics', 'type-erasure'),
           'reified-generics' => tuple('reified-generics', 'reified-generics'),
           'reified-generics-migration' => tuple('reified-generics', 'reified-generics-migration'),
-	  'type-arguments' => tuple('generics', 'type-parameters'),
+	        'type-arguments' => tuple('generics', 'type-parameters'),
```
fredemmott, line 142: Looks like a tab crept in - hackfmt/vscode formatting?

fredemmott (approved):
Happy to address these separately, except for comments about __invoke

fredemmott:
@atry can you give an example of the pattern you're thinking of?

Though, @jamesjwu , another memoize question: is it possible to have a readonly memoized method?
- all memoized methods write to an invisible property of the object, so no
- ... but that can be considered an implementation detail

jamesjwu:
@fredemmott we're allowing it as the second option right now, due to it being an implementation detail and also already having quite a restrictive feature. Thanks for the comments! I'll fix up these docs

guides/hack/16-readonly/01-introduction.md
```diff
@@ -0,0 +1,49 @@
...
+
+### How does it work?
+Expressions in Hack can be annotated with the `readonly` keyword. When an object or reference is readonly, there are two main constraints on it:
+* **Readonlyness:** Object properties can not be modified (i.e. mutated).
+* **Deepness:** All nested properties of a readonly value are readonly.
+
+### Readonlyness
+Object properties of readonly values can not be modified (i.e. mutated).
+
+``` Hack
```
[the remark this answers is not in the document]
jamesjwu: This is cool! qq: this means that each test file would need to be self contained, though, right? Which means adding a lot of class definitions that will increase the verbosity of the examples

That said, I'm all for this since it would make each example runnable on its own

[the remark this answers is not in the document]
fredemmott: - If they're sharable between examples, add a `.inc.hack` file with them
- verbosity: I'm not seeing any usage of this, but it looks like if you add `// @example-start` and `// @example-end` comments, only the code between them will be rendered

src/Guides.php
jamesjwu: I added it because originally it was definitely misformatted; hope that's okay!

fredemmott (approved):

Atry:
> @Atry can you give an example of the pattern you're thinking of?

``` hack
class MyClass {
  readonly int $foo = 42;
}
```

vs


``` hack
class MyClass {
  <<__Memoize>>
  function foo { return 42; }
}
```

PR #1114, opened 2021-11-15
Re-add the fbonly messages to the HSL pages

fredemmott:
Visible if `<body>` has the `isFbViewer` class instead of
`isNotFbViewer`; the default classes depend on if the source IP is
visible.

Test plan:

- extended unit tests
- no visible change with default CSS classes
- screenshot with `isFbViewer` attached:

image


AndrewDiMola (approved):

fredemmott:
db173b87

PR #1116, opened 2021-11-16
Add file attributes to remove some unnecessary type errors

jamesjwu:
This just adds <<file:__EnableUnstableFeatures("readonly")>> to the example tests to decrease the number of unnecessary typechecker outputs.

It would be nice to omit these, and we can do so once the new version of HHVM is released, since the file attributes will become unnecessary.


jamesjwu:
(All of the changes are generated, with the exception of the 5 .md files)

fredemmott (approved):

PR #1117, opened 2021-11-17
[readonly] Fix a bunch of small typos in readonly documentation

jamesjwu:
See title, only markdown changes were made


guides/hack/16-readonly/01-introduction.md
```diff
@@ -39,8 +39,8 @@ function test(readonly Bar $x) : void {
 }
 ```
 
-### How is `readonly` different from contexts and capabilities that control property mutation (such as write_props)?
-[Contexts](https://docs.hhvm.com/hack/contexts-and-capabilities/available-contexts-and-capabilities) such as `write_props` affect an entire function (and all the functions it calls), whereas readonly affects specific values / expressions. 
+### How is `readonly` different from contexts and capabilities that control property mutation (such as `write_props`)?
+[Contexts](https://docs.hhvm.com/hack/contexts-and-capabilities/available-contexts-and-capabilities) such as `write_props` affect an entire function (and all of its callees), whereas readonly affects specific values / expressions.
```
fredemmott, line 43: can you remove `https://docs.hhvm.com/` from this please, so links are local (and show local changes) when developing

fredemmott (approved):

PR #1121, opened 2021-11-23
Add clarity to constants page re: self operator and visibility and type inference

AndrewDiMola:
* Add information on self operator
* Rewrite for clarity for areas like type inference and visibility modifiers

image

image

### DIFF
--- a/guides/hack/06-classes/18-constants.md
+++ b/guides/hack/06-classes/18-constants.md
@@ -16,10 +16,10 @@ function main(): void {
 ```
 
 ## Visibility
-Class constants have public visibility and can not be declared as `public`, `protected`, or `private`.
+Class constants are always public, and can not be explicitly declared as `public`, `protected`, or `private`.
 
 ## Selection
-Inside a parent class, use `self::foo` to access a named constant `foo`. Outside a parent class, a class constant's name must be fully qualified with the class and constant name (e.g. `Bar::foo`). For more information, see [scope-resolution operator, `::`](../expressions-and-operators/scope-resolution.md).
+Inside a parent class, use `self::foo` to access a named constant `foo`. Outside a parent class, a class constant's name must be fully qualified with the class and constant name (e.g. `Bar::foo`). For more information, see [scope-resolution operator, `::`](../expressions-and-operators/scope-resolution).
 
 ## Type Inference
 If a class constant's type is omitted, it can be inferred. For example:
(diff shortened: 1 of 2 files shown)


guides/hack/06-classes/18-constants.md
```diff
@@ -9,17 +10,24 @@ class Automobile {
...
 function main(): void {
-  $col = Automobile::DEFAULT_COLOR;
+  $col = Automobile::DEFAULT_COLOR; // or: $col = self::DEFAULT_COLOR;
   echo "\$col = $col\n";
 }
 ```
 
-If a class constant's type is omitted, it is inferred from the initializer, which must be present. In this case, that type is `string`.
+## Visibility
+Class constants have public visibility and can not be declared as `public`, `protected`, or `private`.
```
fredemmott, line 19: Class constants are always public, and can not be explicitly declared as `public`, `protected`, or `private`.

guides/hack/06-classes/18-constants.md
```diff
@@ -9,17 +10,24 @@ class Automobile {
...
+
+## Selection
+Inside a parent class, use `self::foo` to access a named constant `foo`. Outside a parent class, a class constant's name must be fully qualified with the class and constant name (e.g. `Bar::foo`). For more information, see [scope-resolution operator, `::`](../expressions-and-operators/scope-resolution.md).
```
fredemmott, line 22: Can you normalize this link to match the path?

If you prefer the `../` form, we probably need to improve the internal link test to be able to verify these

fredemmott (approved):

guides/hack/06-classes/18-constants.md
AndrewDiMola: Sure, didn't even notice the different format. Will update.

guides/hack/06-classes/18-constants.md
```diff
@@ -16,10 +16,10 @@ function main(): void {
 ```
 
 ## Visibility
-Class constants have public visibility and can not be declared as `public`, `protected`, or `private`.
+Class constants are always public, and can not be explicitly declared as `public`, `protected`, or `private`.
 
 ## Selection
-Inside a parent class, use `self::foo` to access a named constant `foo`. Outside a parent class, a class constant's name must be fully qualified with the class and constant name (e.g. `Bar::foo`). For more information, see [scope-resolution operator, `::`](../expressions-and-operators/scope-resolution.md).
+Inside a parent class, use `self::foo` to access a named constant `foo`. Outside a parent class, a class constant's name must be fully qualified with the class and constant name (e.g. `Bar::foo`). For more information, see [scope-resolution operator, `::`](../expressions-and-operators/scope-resolution).
```
fredemmott, line 22: It's more the `../` than the `.md` :)

PR #1122, opened 2021-11-24
Enum Class Labels documentation

vsiles:
Introducing doc for enum class labels. This is currently an experimental
feature which should be released by the end of the year. A label is an
alternative way to access an enum class value, remembering which binding
was used: `E::valueOf(E#A) = E::A`.


AndrewDiMola:
Hey @vsiles, wondering if this released yet / if not, when it would. I think it would make sense to add the docs here if it's not released https://docs.hhvm.com/hack/experimental-features and if it is, remove the experimental blurb on the main page.

vsiles:
I'll land the final diff today/tomorrow so it will be release "soon" (when HHVM release picks my diff). I'd jsut keep the doc in here if that's ok, and remove the header / merge once it is deployed

AndrewDiMola:
np! was just wondering if you were waiting on me. feel free to keep it here.

vsiles:
@AndrewDiMola I removed the section about experimental. If you don't have any comment on the content itself, I'll land it once the feature is released (sometime next week)

vsiles:
@AndrewDiMola  I tried to rebase but got a strange CI error. Will wait for you to wake up see If I did something stupid or if the error is unrelated

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -0,0 +1,96 @@
+## Values v. Bindings
+
+With [enum types](enumerated-types) and [enum classes](enum-class), most of the focus is given to their values.
+Expressions like `E::A` denote the value of `A` in `E`, but the fact that `A` was used to access it is lost.
```
AndrewDiMola, line 4: I could be wrong, but did you mean to say that the fact that the enum `E` was used to access the value at `A' is lost? Or is it correct to say that `A` is lost because the value at `A` is only cared about.

(Same note for line 24)

vsiles: We can "keep" the fact about `E` is kept a bit: you can have a `function f(E $x)` and pass in `E::A`. However in `f`, anything about `A` is gone

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -0,0 +1,96 @@
...
+
+This example defines two constants `E::A: \HH\MemberOf<E, int>` and `E::B: \HH\MemberOf<E, int>`. Enum class labels add more definitions to the mix:
+
+- `E#A: \HH\EnumClass\Label<E, int>` is the label to access `E::A`
+- `E#B: \HH\EnumClass\Label<E, int>` is the label to access `E::B`
+- `E::nameOf` is a static method expecting a label and returning its string representation: `E::nameOf(E#A) = "A"`
+- `E::valueOf` is a static method expecting a label and returning its value: `E::valueOf(E#A) = E::A`
+
+So we can rewrite the earlier example in a more resilient way:
+```EnumClassLabel.example.achk no-auto-output
```
AndrewDiMola, line 48: should be .hack

AndrewDiMola:
@vsiles we lag behind HHVM releases -- as of writing, right now on [4.136](https://github.com/hhvm/user-documentation/blob/main/src/codegen/PRODUCT_TAGS.php#L10). Guessing it's still unstable / requires explicit tags until we reach [4.139](https://github.com/facebook/hhvm/releases/tag/HHVM-4.139.0).

vsiles:
> @vsiles we lag behind HHVM releases -- as of writing, right now on [4.136](https://github.com/hhvm/user-documentation/blob/main/src/codegen/PRODUCT_TAGS.php#L10). Guessing it's still unstable / requires explicit tags until we reach [4.139](https://github.com/facebook/hhvm/releases/tag/HHVM-4.139.0).

Ok ! I'll make the fix and depending on how long the release process is here I can wait before landing. Do you have an ETA on the releases ?

fredemmott:
There's docker issues until D33002177 is merged (it provides a way to disable hack's ASLR poking); I expect a 4.139.1 once that's done, and it should be in 4.140 too

vsiles:
Is there a way to get notified of these release ?

fredemmott:
@vsiles if you subscribe to https://github.com/hhvm/hhvm.com/pull/170 you'll get a notification when the release is live

fredemmott:
> Anyway, would it be ok to land this diff with the __EnableUnstableFeatures flag and a mention saying this attribute won't be necessary once HHVM is updated ?

This is completely fine

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -0,0 +1,96 @@
...
+
+### The `#` character is no longer a one line comment
+This feature relies on the fact that Hack and HHVM no longer consider the character `#` as a one line comment. Please use `//` for such purpose.
+
+### Labels and values cannot be exchanged
+If a method is expecting a label, one cannot pass in a value, and vice versa: `full_print(E::A)` will result in a type error and so will `partial_print(E#A)`.
+
+### `MemberOf` is covariant, `Label` is invariant
+A label should be considered as a way to attach a type to a binding. Therefore it is invariant: `E#A` is not of type `\HH\EnumClass\Label<E, arraykey>`.
+This can be misleading at first, because `\HH\MemberOf` is invariant (it is data after all): `E::A` is of type `\HH\MemberOf<E, arraykey>`.
```
vsiles, line 96: TODO:change this invariant to covariant

vsiles:
@AndrewDiMola  ok so I managed to edit the files I wanted using `codespaces`, but the build fails because I can't refresh the php files: `hhvm bin/build.php` fails with the following error
```
Fatal error: require_once(/workspaces/user-documentation/bin/../vendor/autoload.hack): File not found in /workspaces/user-documentation/bin/build.php on line 17
```
(and the file is indeed missing)

vsiles:
I updated the doc putting the attribute back until we sort this hhvm version situation. If everything's green, I'll take a review and land this :)

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -0,0 +1,102 @@
+# Disclaimer
+This feature is on its way to be deployed. Most recent HHVM versions no longer need the `EnableUnstableFeatures` file attribute
+to support this. More information available at  https://github.com/hhvm/user-documentation/pull/1122
+
+## Values v. Bindings
+
+With [enum types](enumerated-types) and [enum classes](enum-class), most of the focus is given to their values.
```
AndrewDiMola, line 7: Can you use the full path for links?
* (enumerated-types) -> (/hack/built-in-types/enum)
* (enum-class) -> (/hack/built-in-types/enum-class)

This format lets our test suite push error notifications if page titles change and someone forgets to update links in other docs pages (like this one).

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -0,0 +1,102 @@
...
+
+function partial_print(\HH\MemberOf<E, int> $value) : void {
+  echo $value . "\n";
+}
+```
+Now, `full_print(E#A)` will echo `A 42` and `full_print(E#B)` will echo `B 42`.
+
+## Full v. Short Labels
+
+We refer to labels like `E#A` as *fully qualified* labels: the programmer wrote the full enum class name. However there are some situations where Hack can infer
```
AndrewDiMola, line 68: I would rewrite for specificity as:

* [_before_]: However there are some situations where Hack can infer this part, and we can save the programmer that bit. For example,
* [_recommended_]: However there are some situations where Hack can infer the class name; for example,

AndrewDiMola:
Not on you but we will need to fix the highlighter to adapt for label syntax (see code example)
image

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -0,0 +1,102 @@
...
+this part, and we can save the programmer that bit. For example, the previous calls could be written as `full_print(#A)` and `full_print(#B)`, leaving `E` implicit.
+This is only allowed when there is enough type information to infer the right enum class name. For example, `$x = #A` is not allowed and will result in a type error.
+
+### Special case of function calls
+
+When the first argument of a function is a label, we provide an alternative notation to call it. This was done to reflect some generated code patterns this feature helped removed:
```
AndrewDiMola, line 74: We're trying to keep implementation details out of the docs so I think we should remove this line:

> This was done to reflect some generated code patterns this feature helped removed

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -0,0 +1,102 @@
...
+  set(#A, 42);
+  set#A(42);
+}
+```
+
+As you can see, the short name can be written *before* the opening parenthesis of the function call. This only works for a single label argument, which must be the first one.
+
+## Known corner cases
+
+### The `#` character is no longer a one line comment
```
AndrewDiMola, line 94: Minor, but we use the terminology "single-line comment"

AndrewDiMola: huh, if this true, we should remove  the '#' in "Single-line comments start with // or #, and end with a newline." in 
 [hack/source-code-fundamentals/comments](https://docs.hhvm.com/hack/source-code-fundamentals/comments)

AndrewDiMola (approved):
Looks good! After the merge, I'm going to add a link to enum labels in the getNames section of the new emum page (which you don't have in your branch yet) @ https://docs.hhvm.com/hack/built-in-types/enum#library-methods__getvalues--getnames

guides/hack/11-built-in-types/36-enum-class-label.md
vsiles: Indeed, that change has been approved a couple month ago now and `#` is no longer a single-line comment.

vsiles:
@AndrewDiMola Thank you for the review ;) If you don't have further feedback I'll merge it tomorrow (CET)

AndrewDiMola (approved):
looks good! typo on line 94: single if you don't mind an extra commit, otherwise I'll patch it after.

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -0,0 +1,103 @@
...
+  set(#A, 42);
+  set#A(42);
+}
+```
+
+As you can see, the short name can be written *before* the opening parenthesis of the function call. This only works for a single label argument, which must be the first one.
+
+## Known corner cases
+
+### The `#` character is no longer a signle-line comment
```
lexidor, line 95:
```suggestion
### The `#` character is no longer a single-line comment
```

PR #1126, opened 2021-12-03
Remove external mention of the hsl

lexidor:
The hsl is built-in. Let's use TypeAssert as an example instead.

### DIFF
--- a/guides/hack/01-getting-started/03-starting-a-real-project.md
+++ b/guides/hack/01-getting-started/03-starting-a-real-project.md
@@ -1,7 +1,7 @@
 ## Starting A Real Project
 
 Real projects generally aren't a single file in isolation; they tend to have
-dependencies such as the [Hack Standard Library], and various optional tools.
+dependencies such as the [TypeAssert](https://github.com/hhvm/type-assert), and various optional tools.
 
 A good starting point is to:
 - [install Composer]


guides/hack/01-getting-started/03-starting-a-real-project.md
```diff
@@ -1,7 +1,7 @@
 ## Starting A Real Project
 
 Real projects generally aren't a single file in isolation; they tend to have
-dependencies such as the [Hack Standard Library], and various optional tools.
+dependencies such as the [TypeAssert](https://github.com/hhvm/type-assert), and various optional tools.
```
fredemmott, line 4: - 'the TypeAssert library' or 'TypeAssert'
- How about 'such as libraries, frameworks, or various optional tools' ? TypeAssert:
  - is a 'necessary evil'
  - it's not clear why it's useful to someone who doesn't yet understand how hack's type system (and runtime enforcement) works
 
 There's things like 'hacktest' that might be a better example, but I don't think one's really needed, and putting something here feels like 'most projects will want this' that I don't realyl want to imply

lexidor: I'll be touching up this file anyway #1127. Changing this link was a knee jerk reaction on hhvm/hsl#168. After rereading this page, I realized that it could use a lot of little touch ups. Closing for now. Thanks for the insights.

PR #1128, opened 2021-12-03
Enum refactor, public methods, and Types introduction

### LINKED ISSUE #648: No documentation about enum functions (::assert and ::coerce etc)

# Where is the problem?

https://docs.hhvm.com/hack/types/enumerated-types

# What is the problem?

There is no mention of any of the enum functions on the enum page, or anywhere else in the docs. Content from https://github.com/hhvm/hack-langspec/blob/master/spec/13-enums.md#the-predefined-enumerated-type-methods might be interesting to reuse.

This documentation used to exist (still shows up in `bunny s enum assert`, first result).

 - Build ID: 2019-05-07T20:37:42+0000:9e9731fda3c705fc7f2980868c1e6269f23b46d0
 - Page requested: /hack/types/enumerated-types
 - Page requested at: Wed, 08 May 2019 23:34:26 +0000

### LINKED ISSUE #800: Mention as/?as for converting enum values
Converting to/from enum values is a common action with enums; it would be useful to explain how to do this at https://docs.hhvm.com/hack/built-in-types/enumerated-types (or link to conversion docs elsewhere if preferred).

From an internal discussion:

```Hack
enum Duck: string {
  MALLARD = 'mallard';
  CANVASBACK = 'canvasback';
}

$definitely_duck as Duck; // converts to Duck, throws if not
$maybe_duck ?as Duck; // converts to ?Duck; null if is not a valid value
```

> "That would be great to include in the enum docs at https://docs.hhvm.com/hack/built-in-types/enumerated-types"

### LINKED ISSUE #924: Enums doc page does not document the `coerce` method

# Where is the problem?

The section describing Enums does not document the `coerce` method.

# What is the problem?

The section describing Enums does not document the `coerce` method.

 - Build ID: HHVM=HHVM-4.74.0:HSL=v4.41.0:2020-10-01T21:06:44+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/built-in-types/enumerated-types
 - Page requested at: Thu, 01 Oct 2020 21:46:13 +0000

AndrewDiMola:
Refactor of enum docs and types introduction. Too long to screenshot everything.

# Enum rewrite
* `enumerated-types.md` changed to `enum.md` to help with search, as the enum page does not return for an `enum` search query.
* new documentation and examples for public static methods: `assert()`, `assertAll()`, `coerce()`, `isValid()`.
* remove experimental instructions for enum inclusion.
* refactor into sections (e.g. quickstart) and tighten language.
* fixes #924
* fixes #648 
* fixes #800

image
image
image
image

# Types introduction
* convert links to testable format.
* write an actual introduction, explaining concepts like `mixed` and `nothing`, iterating out details of subtype inheritance.

image
image


AndrewDiMola:
I'd also like to write something on #800 but from what I can tell `as` `?as` don't completely map 1:1 to `assert` and `coerce` (not sure if that's a future expectation)... though maybe `is` does map to `isValid`

lexidor:
The current behavior of `is` on enums is the following:
`is` matches isValid. When the LHS is not an arraykey, the result is false. If the LHS is an arraykey and the enum contains one or more members with the same value as the LHS, true is returned. If the LHS is an integer and the enum contains one or more members of type string that are the canonical representation of the LHS, true is returned. If the LHS is an intish string and the enum contains one or more values of type int that match the `(int) $lhs` cast of the LHS, true is returned. All other cases return false.

The behavior of `as` on enums is the following:
If the `is` expression is true, the `as` expression resolves to the LHS. If the `is` expression is false and the `as` operator is not prefixed with a question mark, a TypeAssertionException is thrown. If the `is` expression is false and the `as` operator is prefixed with a question mark, the expression resolves to `null`.

This part of Hack is unsound. Passing the string "1" to a function which expects an enum with an int member 1 will throw a TypeError. But both `is` and `as` refine the string "1" to type EnumWithAnIntOneMember. This may be changed in the future to be a sound membership check. See https://hhvm.com/blog/2021/10/05/hhvm-4.136.html future changes and https://hhvm.com/blog/2018/08/28/hhvm-4.2.0.html section Supported Types.

guides/hack/11-built-in-types/34-enum.md
```diff
@@ -0,0 +1,197 @@
...
+  writeText("Hello", Position::Bottom);
+  writeText("Today", Position::Left);
+}
+```
+
+## Library Methods
+All enums implement these public static methods.
+
+### `getNames() / getValues()`
+Returns a map-like array of enum constant values and their names.
```
lexidor, line 80: It is true that all enums implement these two methods. An enum with duplicate values will throw an exception from getNames. This is done because the returned dict could not contain all members, since it is keyed by the values of the members. It might be handy to document this behavior here.

guides/hack/11-built-in-types/34-enum.md
```diff
@@ -0,0 +1,197 @@
...
+  foreach ($values as $key => $value) {
+    echo "\tkey >$key< has value >$value<\n";
+  }
+}
+```
+
+### `assert()` / `coerce()`
+`assert($value)` checks if a value exists in an enum, and if it does, returns the value; if the value does not exist, throws an `UnexpectedValueException`.
+
+`coerce($value)` checks if a value exists in an enum, and if it does, returns the value; if the value does not exist, returns `null`.
```
lexidor, line 110: Both assert and coerce may cast the value given from int to string if no direct int match is found. The same goes for strings that are canonical representations of ints. They will be cast to an integer if no string match is found. This is maybe expected for coerce, but noteworthy and strange for assert.

AndrewDiMola: I can add as:

image

lexidor: My description was incomplete and this incompleteness leaked into your description as well. The cast string -> int / int -> string value is only returned if the result of the cast is a member of the enum. The current description hint towards unconditionally returning the result of the cast. A more accurate description would be:

Both library methods will, when no exact match is found, attempt to do a cast to the other arraykey type. If the cast is not reversible / lossless, or the resulting value is still not a member of the enum after the cast, the failure result occurs. Where failure for assert is throwing a UnexpectedValueException and for coerce it is returning null.

lexidor: And string in the format of a number is a wider description than and intish string. "3.5" is a string in the format of a number, but not an intish string.

AndrewDiMola: Yeah, I agree that it is too wide of a description. Was writing from a flawed viewpoint earlier, but has been updated.

Going to use your second paragraph above as a note with some slight wording changes. Thanks!

guides/hack/11-built-in-types/34-enum.md
AndrewDiMola: I made this much clearer, thanks!

image

AndrewDiMola:
> The current behavior of `is` on enums is the following: `is` matches isValid. When the LHS is not an arraykey, the result is false. If the LHS is an arraykey and the enum contains one or more members with the same value as the LHS, true is returned. If the LHS is an integer and the enum contains one or more members of type string that are the canonical representation of the LHS, true is returned. If the LHS is an intish string and the enum contains one or more values of type int that match the `(int) $lhs` cast of the LHS, true is returned. All other cases return false.
> 
> The behavior of `as` on enums is the following: If the `is` expression is true, the `as` expression resolves to the LHS. If the `is` expression is false and the `as` operator is not prefixed with a question mark, a TypeAssertionException is thrown. If the `is` expression is false and the `as` operator is prefixed with a question mark, the expression resolves to `null`.
> 
> This part of Hack is unsound. Passing the string "1" to a function which expects an enum with an int member 1 will throw a TypeError. But both `is` and `as` refine the string "1" to type EnumWithAnIntOneMember. This may be changed in the future to be a sound membership check. See https://hhvm.com/blog/2021/10/05/hhvm-4.130.html future changes and https://hhvm.com/blog/2018/08/28/hhvm-3.28.0.html section Supported Types.
> 
> _Edit: Repaired the blog links. My keyboard has "learned" the words hhvm-4.2.0.html and assumed hhvm-3.28.0.html was a typo in need of correction._

I think this will go a long way, x-linking to the pages on type refinement and included a set of examples. I'm wary about putting notes about things that might be changed, though I agree it's unexpected. Going to add as:
image

guides/hack/11-built-in-types/34-enum.md
lexidor: I believe you may have flipped the examples around. `getNames()` returns `value` => `MemberName` and `getValues()` returns `MemberName` => `value`. 🙂

I have changed the `getNames()` description, as _Can not contain duplicate keys_ does not apply to `getValues()`. If you want, you may use my text verbatim, use elements, or not use it all.

## getValues() / getNames()

Returns a `dict` of enum constant values and their names.

 - `getValues()` returns a `dict` where the keys are the enum names and the values are the enum constant values.
   - In the example below, the keys/values would be: `"Top" => 0`, `"Bottom" => 1`, etc.
 - `getNames()` returns a `dict` which is the flipped result of `getValues()`\*; the keys are the enum constant values and the values are the enum's named constants.
   - Following the same example, the keys/values would be: `"0 => "Top"`, `1 => "Bottom"`

\* Dicts can not contain duplicate keys. `getNames()` can not return all members in enums with duplicate constant values. Calling `getNames()` on an enum with duplicate constant values will throw an `HH\InvariantException`, since the alternative would have been silently discarding members. If you want to silently discard members instead of throwing an exception, use `HH\Lib\Dict\flip(Position::getValues())`

```HACK
enum Position: int {
  Top = 0;
  Bottom = 1;
  Left = 2;
  Right = 3;
  Center = 4;
  Centre = 4;
}

<<__EntryPoint>>
function main(): void {
  \var_dump(Position::getNames());
}
```

```
'HH\InvariantException' with message
'Enum Position has overlapping values Center and Centre'
```

AndrewDiMola: oh, you're totally right! Thank you :-)

AndrewDiMola: Going to merge but I can always stealth edit. Fixed the mistake and wrote some more context on why `getNames()` presents an issue.

image

PR #1132, opened 2021-12-06
Use notation for union, not intersection

### LINKED ISSUE #1131: Typo about ?? operator's type

# Where is the problem?

- section "Usages"
 
# What is the problem?

- It is written that (T & nothing) is just T. This is very confusing I think. I'm unsure where & is formally defined but this looks like an intersection of types. nothing is the empty type, so T&nothing is not equivalent to T (except if T is nothing). The type of the ?? operator is a "union". And indeed "T union nothing" is equivalent to T.

 - Build ID: HHVM=HHVM-4.136.0:HSL=v4.108.1:2021-12-02T18:46:15+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/built-in-types/nothing
 - Page requested at: Mon, 06 Dec 2021 07:46:41 +0000

lexidor:
Fixes #1131

### DIFF
--- a/guides/hack/11-built-in-types/74-nothing.md
+++ b/guides/hack/11-built-in-types/74-nothing.md
@@ -36,7 +36,7 @@ function returns_an_int(?int $nullable_int): int {
   $throwing_expr_lambda = () ==> throw_as_an_expression(new \Exception());
 
   // You can't write a statement on the RHS of an operator, because it operates on expressions.
-  // The type of the `??` operator is `(nothing & int)`, which simplifies to `int`,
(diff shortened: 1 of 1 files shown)


lexidor:
Explanation for why a union is correct instead of an intersection. I had the notation mixed up when I authored this. https://github.com/hhvm/user-documentation/issues/1131#issuecomment-986565566

lexidor:
Converted to draft because I need to rerun the build locally to pass CI. This will have to wait a little while.

Wilfred:
Rebuilt and merged, thanks!

PR #1139, opened 2021-12-20
Update to 4.140

fredemmott:
I intend to merge as 4 commits:

- Update versions
- Update examples/expectf for 4.140
- Make `expect::toMatchExpect*File` operate on paths rather than strings
- example typechecking fix for 4.139+

Probably best to review separately too.

coeffects violation in Pair::concat() is a bug in HHVM.

refs #1122


fredemmott:
Error is reproducible, but only when running tests in docker; get expected output when file is standalone, or when running tests outside of docker.

fredemmott:
.stderr file: error: daemon_stubs.c: caml_disable_ASLR: failed to set personality

fredemmott:
Going to:
- rebase on #1148 once that's merged
- follow up with an update to 4.143

Wilfred (approved):

PR #1141, opened 2021-12-23
rename shapes to filter in results for shape; boost guides search scores

AndrewDiMola:
* `shape` is a top 10 search query; the guides page does not list in search because the title is "shapes". Re-titling to "shape" to fit actual search interest and boost its score.
* guides pages seem to perform poorly for top 10 search queries like `string`, `vec` (and `dict`, `keyset`), and others like `format`.

This change actually doesn't boost guides pages for `vec` or `format` to the top, but it does put them on more equal footing.

Current:
* image
* image
* image

Proposed:
* image
* image
* image


Wilfred (approved):
Seems like an improvement! You'll need to add a redirect from shapes to shape to make the build happy.

PR #1146, opened 2022-01-06
clarify use of `#` and `#!` in comments guide

AndrewDiMola:
* Remove `#` as a valid comment character
* Add link to Enum Class Label docs
* Add information on `#!` to special characters
* Use full paths for hrefs

image

### DIFF
--- a/guides/hack/02-source-code-fundamentals/10-comments.md
+++ b/guides/hack/02-source-code-fundamentals/10-comments.md
@@ -29,4 +29,3 @@ A number of special comments are recognized; they are:
 * `// strict` and `// partial` in [headers](/hack/source-code-fundamentals/program-structure)
 * `/* HH_FIXME[1234] */` or `/* HH_IGNORE_ERROR[1234] */`, which
   suppresses typechecker error reporting for error 1234.
-* `#!` in `#!/path/to/executable` (e.g. `/usr/bin/hhvm`), which is the shebang line (interpreter directive).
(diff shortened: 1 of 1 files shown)


guides/hack/02-source-code-fundamentals/10-comments.md
```diff
@@ -16,14 +16,17 @@ Hack has three comment syntaxes.
...
 
 A number of special comments are recognized; they are:
 
-* `// FALLTHROUGH` in [switch statements](../statements/switch.md)
-* `// strict` and `// partial` in [headers](program-structure.md)
+* `// FALLTHROUGH` in [switch statements](/hack/statements/switch)
+* `// strict` and `// partial` in [headers](/hack/source-code-fundamentals/program-structure)
 * `/* HH_FIXME[1234] */` or `/* HH_IGNORE_ERROR[1234] */`, which
   suppresses typechecker error reporting for error 1234.
+* `#!` in `#!/path/to/executable` (e.g. `/usr/bin/hhvm`), which is the shebang line (interpreter directive).
```
fredemmott, line 32: - this shouldn't be described as a comment
- should use `/usr/bin/env hhvm` instead
- should note that it is only permitted as the very first line

AndrewDiMola: I was thinking that it was a special type of comment like the others (because of the # in #!) but the others do all fall into the structure of being within valid comment syntax so you're right it doesn't belong there.

PR #1149, opened 2022-01-11
Rework SASS, remove all Ruby dependencies

fredemmott:
This is a little clunky, but given the build's completely broken at the
moment, I'd rather ship this first, and we can revisit it later if
there's a need.

We need to unblock other contributions asap.

The primary goal is to remove the dependency on Ruby SASS, as the
version we were using is no longer installable. Switch to Dart SASS.

This led to:
- stop using 'transition()` mixing from Bourbon:
  - newer bourbons don't have it
  - pretty much every browser supports it nowadays
  - if they don't... we've lost a transition, which isn't a huge deal
- ... and now we don't need Bourbon
- Remove stale references to the Hack font in the CSS
- Upgrade from font-awesome 4.x to 5.x as I couldn't find a good
  non-Rubygems source for 4.5


public/fonts/fontawesome-free-5.15.4-web
```diff
@@ -0,0 +1 @@
+../../build/final/fontawesome-free-5.15.4-web/webfonts
```
fredemmott, line 1: This is a symlink

sass/core.scss
```diff
@@ -1,8 +1,8 @@
 @import "variables";
-@import "bourbon";
 @import "reset";
 
-@import "font-awesome";
+@import "fontawesome";
+@import "solid";
```
fredemmott, line 5: this was previously part of font-awesome

sass/variables.scss
```diff
@@ -39,7 +38,7 @@ $side-padding-mob: 10px;
 /* Mixins */
 
 @mixin codetext {
-  font-family: Hack, "Lucida Console", Monaco, "Courier New", monospace;
```
fredemmott, line 42: not an actual change as we've not been making the webfont available for a very long time

AndrewDiMola (approved):

PR #1160, opened 2022-02-14
Prevent hh-apidoc from being upgraded to an incompatible version

Atry:
According to https://github.com/hhvm/hh-apidoc/releases/tag/v0.8.0 , hh-apidoc is not backward compatible with 0.6.0.


fredemmott:
Change is good, however:

> This PR should fix the nightly CI

There is a lock file; this should be an unobservable change by itself. It should have no effect on CI, and the failure appears unrelated.

fredemmott (approved):

PR #1162, opened 2022-02-14
Add VSCode run-and-debug config for building the site

fredemmott:
This makes it easy to use the vscode debugger for the build process


fredemmott:
e.g. 

image

Entered by selecting 'Build' just to the right of 'run and debug' in the top left while on the debugger tab, or in the bottom bar.

Atry (approved):

PR #1164, opened 2022-02-14
Document inherited methods

### LINKED ISSUE #1125: Showing me a type without the [inherited] methods defined on it is pretty useless
Trying to use Hacklang for Advent of code.

A documentation file like this: https://docs.hhvm.com/hsl/reference/interface/HH.Lib.File.CloseableReadHandle/

Is basically useless? Maybe I'm not sure how to use this website correctly... but I want to see the methods defined on a type. I'm trying to read in a text file, and every thing just links to interfaces like this with no methods defined. Even if they're not *defined* on this type, and come from a class that this type extended or an interface it implemented, I think this documentation would only be useful if I could see those methods here.

fredemmott:
This (more than slightly) needs a refactoring cleanup... will do that in
a followup, then look at the type/newtype stuff (which is currently not
supported at all, and has been requested for a while)

fixes #1125

### DIFF
--- a/src/build/HHAPIDocBuildStep.php
+++ b/src/build/HHAPIDocBuildStep.php
@@ -42,10 +42,8 @@ final class HHAPIDocBuildStep extends BuildStep {
       self::findSources(BuildPaths::HHVM_TREE.'/hphp/system/php/', $exts),
       self::findSources(BuildPaths::HHVM_TREE.'/hphp/runtime/ext/', $exts),
     );
-    $hhi_sources = self::findSources(
-      BuildPaths::HHVM_TREE.'/hphp/hack/hhi/',
-      $exts,
-    );
+    $hhi_sources =
+      self::findSources(BuildPaths::HHVM_TREE.'/hphp/hack/hhi/', $exts);
     Log::i("\nParsing builtins");
     list($runtime_defs, $hhi_defs) = \HH\Asio\join(async {
       return tuple(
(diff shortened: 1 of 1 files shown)


fredemmott:
First commit is #1163

Second commit gets me this (ignoring the squished layout because of minimum browser zoom):


image

src/build/HHAPIDocBuildStep.php
```diff
@@ -107,14 +101,46 @@ public function buildAll(): void {
...
+      }
+      if ($def is ScannedInterface) {
+        $hh_apidoc_index['interfaces'][$def->getName()] = $documentable;
+        continue;
+      }
+      if ($def is ScannedTrait) {
+        $hh_apidoc_index['traits'][$def->getName()] = $documentable;
+        continue;
+      }
+    }
```
Atry, line 136: I would use `Dict\group_by` to avoid mixing up of functional programming style code and the mutable loop.

fredemmott: `Dict\group_by()` doesn't produce shapes, and would need a call to reflection (or `get_class()` to significantly clean this up

fredemmott: There also currently isn't functional style here?

Atry (approved):

src/build/HHAPIDocBuildStep.php
Atry: I think it would be good if we have a high level API in `Facebook\HHAPIDoc`

fredemmott: HHAPIDoc does have one; docs.hhvm.com uses its' own as there's multiple sources (hhvm, HHI, hsl, hsl-experimental) and they can contain contradictory definitions (hhvm + hhi) which need resolving.

PR #1165, opened 2022-02-17
Use cmake -B and -S flags to specify the build path

Atry:
According to https://cmake.org/cmake/help/latest/manual/cmake.1.html , `-B` and `-S` flags are the right way to specify the build path


fredemmott (changes requested):
- I'm not sure this is 'recommended' as such - it's *a* way, but the way we're doing things is still documented, and (sorry if I missed it) I'm not seeing a deprecation notice or similar
- Unless there is a clear recommendation, I prefer the existing syntax: it's compatible with `cmake --build` and `cmake --install`, but `-S` and `-B` appear to be specific to the configure stage. For context, these are useful when using a backend other than `make` - for example, xcode projects or `ninja`. It will run `make`, `ninja`, `msbuild`, `xcbuild`, or whatever it was configured with, with consistent arguments. This is very handy for cross-platform CI.

I'm not certain on those two points, but regardless, we shouldn't do this *yet*: we still support Ubuntu 18.04 (cmake 3.10) and Debian Buster (cmake 3.13), and these options appear to be new in CMake 3.14.

- from a user perspective, I think we should aim for as few version-specific instructions as possible
- from our perspective, applying this to our own build system would require forking the packaging scripts between versions

Atry:
How about add the `-B` and `-S` flags to macOS build? Homebrew should install the latest cmake.

fredemmott (changes requested):
That's true, however:
- there's still not an indication it's preferred
- I'd rather us avoid unnecessary differences between any of our platforms; keeping stuff the same makes it easier to focus on what's actually different and potentially behavior-changing

PR #1166, opened 2022-02-19
Clarity updates in text

AndrewDiMola:
Quick updates based on a conversation re: usage of 'deprecated' with @fredemmott, and some other minor updates.


guides/hack/30-silencing-errors/05-error-codes.md
```diff
@@ -362,7 +362,7 @@ if ($y) {
 }
 ```
 
-This error is deprecated, as it was making it harder to convert
+This error is obsolete, as it was making it harder to convert
```
fredemmott, line 365: IMO obsolete implies no longer relevant, or replaced. In this case, it might be better just to say 'no longer raised'

PR #1170, opened 2022-03-03
home page reclassification

AndrewDiMola:
image
image


fredemmott (approved):
It'd be good to add redirects for all these pages that are moving

AndrewDiMola:
> It'd be good to add redirects for all these pages that are moving

This shouldn't change the URLs around at all. `/hack/contexts-and-capabilities` for example is the same path no matter what overarching category it's within on the front page.

Issue #1171, opened 2022-03-07
Stack example is not correctly implemented in Generics Introduction


xiaoleichen:
Please complete the information below:

# Where is the problem?

Generics -> Introduction
https://docs.hhvm.com/hack/generics/introduction

# What is the problem?

Stack is incorrectly implemented. The pop operation doesn't pop the item off the stack. For example:
```
  $stInt = new Stack<int>();
  $stInt->push(10);
  $stInt->push(20);
  $stInt->push(30);
  echo 'pop => '.$stInt->pop()."\n";
  $stInt->push(40);
  echo 'pop => '.$stInt->pop()."\n";
```
outputs:
```
pop => 30
pop => 30
```
It's supposed to be:
```
pop => 30
pop => 40
```

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: HHVM=HHVM-4.147.0:HSL=v4.108.1:2022-03-01T17:24:53+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/generics/type-parameters
 - Page requested at: Mon, 07 Mar 2022 05:09:27 +0000
 - Controller: GuidePageController

AndrewDiMola:
@Wilfred could we use `C\pop_back` on the vec in this example?

AndrewDiMola:
within:

```
public function pop(): T {
    if ($this->stackPtr > 0) {
      $this->stackPtr--;
      return $this->stack[$this->stackPtr];
    } else {
      throw new StackUnderflowException();
    }
  }
```

lexidor:
I am not Wilfred, but can answer nevertheless. You can reduce this example to a `[]=` in push and `C\is_empty()` in isEmpty, and `C\pop_backx()` + `$this->isEmpty()` in pop. No need to keep track of a stack pointer.

AndrewDiMola:
> I am not Wilfred, but can answer nevertheless. You can reduce this example to a `[]=` in push and `C\is_empty()` in isEmpty, and `C\pop_backx()` + `$this->isEmpty()` in pop. No need to keep track of a stack pointer.

Of course! Will try this implementation now. Thanks @lexidor!

Atry:
> I am not Wilfred, but can answer nevertheless. You can reduce this example to a `[]=` in push and `C\is_empty()` in isEmpty, and `C\pop_backx()` + `$this->isEmpty()` in pop. No need to keep track of a stack pointer.
> 
[...]

PR #1173, opened 2022-03-10
Minor fixes for Issues

### LINKED ISSUE #1137: <<__Deprecated>> annotation

# Where is the problem?
The <<__Deprecated>> annotation

# What is the problem?
The <<__Deprecated>>annotation is kinda misleading/doesn't square well with FB internal tooling and serves a different purpose internally. Discussion https://fb.workplace.com/groups/hackstyle/posts/6684528621618432

 - Build ID: HHVM=HHVM-4.136.0:HSL=v4.108.1:2021-12-18T03:29:23+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/attributes/predefined-attributes
 - Page requested at: Mon, 20 Dec 2021 17:23:32 +0000

### LINKED ISSUE #1142: Incorrect parentheses

# Where is the problem?

Contexts & Capabalities > Local Operators

# What is the problem?

Incorrect parentheses in the english, also confusing.

I see the wording "In the world where all these contexts (and matching capabilities exist), some or all may be included within the defaults context." In english grammar, remarks in parentheses should be removable without changing the meaning of the sentence. Perhaps "exist" should be outside of the parentheses here.

But while we're at it, I'm still pretty confused at what "In the world where" means in this documentation. Am I to understand these contexts/capabilities don't already exist built in to HHVM? I thought all contexts/capabilities are defined in the "typechecker and runtime" so wouldn't that imply the set of contexts/capabilities that exist is hardcoded? 

 - Build ID: HHVM=HHVM-4.136.0:HSL=v4.108.1:2022-01-03T04:48:47+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/contexts-and-capabilities/local-operations
 - Page requested at: Tue, 04 Jan 2022 16:37:51 +0000

### LINKED ISSUE #1147: Mention that type aliases can take generic arguments/params
There should be a mention, either in the [Type Aliases](https://docs.hhvm.com/hack/types/type-aliases) section or the [Generics](https://docs.hhvm.com/hack/generics/introduction) section that type aliases can have generic parameters.

 - Build ID: HHVM=HHVM-4.136.0:HSL=v4.108.1:2022-01-07T22:03:28+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/types/type-aliases
 - Page requested at: Mon, 10 Jan 2022 20:33:02 +0000

### LINKED ISSUE #1154: string encoding docs should mention `Regex\`
This page talks a lot about how `Str\` works with bytes vs characters; it should also mention that `Regex\` is primarily byte-based, unless the `/u` modifier is present in the pattern.

The `Regex\` docs should probably also mention this and link back to the string page.

 - Build ID: HHVM=HHVM-4.140.1:HSL=v4.108.1:2022-01-14T20:09:46+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/built-in-types/string
 - Page requested at: Fri, 28 Jan 2022 17:35:19 +0000

### LINKED ISSUE #1168: missing ;s in code example

# Where is the problem?

The first code example for Exceptions, on the Concurrent page

# What is the problem?

await async { throw new Exception('foo'); }
await async { throw new Exception('bar'); }

should each end with a ;

 - Build ID: HHVM=HHVM-4.147.0:HSL=v4.108.1:2022-02-23T01:00:33+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/asynchronous-operations/concurrent
 - Page requested at: Thu, 24 Feb 2022 22:05:39 +0000

### LINKED ISSUE #1172: Typo $this->p1 in Type Refinement
# Where is the problem?

Last code example and last paragraph.

# What is the problem?

Replace both occurrences of `$this->p1` with `$this->p` (there is no `$p1` property).

 - Build ID: HHVM=HHVM-4.147.0:HSL=v4.108.1:2022-03-01T17:24:53+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/types/type-refinement
 - Page requested at: Wed, 09 Mar 2022 00:37:38 +0000

AndrewDiMola:
* Fixes #1172 
* Fixes #1168
* Fixes #1142 
* Fixes #1147
* Fixes #1154 
* Fixes #1137

### DIFF
--- a/guides/hack/10-types/79-type-aliases.md
+++ b/guides/hack/10-types/79-type-aliases.md
@@ -2,6 +2,7 @@ We can create an alias name for a type, and it is common to do so for non-trivia
 that alias can be used in almost all contexts in which a type specifier is permitted.  Any given type can have multiple aliases, and a type alias
 can itself have aliases.
 
+## Quickstart
 A type alias can be created in two ways: using `type` and `newtype`.
 
 ```Hack
@@ -9,10 +10,14 @@ type Complex = shape('real' => float, 'imag' => float);
 newtype Point = (float, float);
 ```
 
+A type alias can also be created with [Generics](/hack/generics/introduction) as parameters.
+
+## Using `type`
 An alias created using `type` (such as `Complex` above) is a *transparent type alias*. For a given type, that type and all transparent aliases
 to that type are all the same type and can be freely interchanged. There are no restrictions on where a transparent type alias can be defined,
 or which source code can access its underlying implementation.
 
+## Using `newtype`
 An alias created using `newtype` (such as `Point` above) is an *opaque type alias*. In the absence of a type-constraint (see `Counter`
 example below), each opaque alias type is distinct from its underlying type and from any other types aliasing it or its underlying type.
 Only source code in the file that contains the definition of the opaque type alias is allowed access to the underlying implementation. As
@@ -33,8 +38,9 @@ function distance(Point $p1, Point $p2): float {
 }
 ```
 
-Being in the same source file as the alias definition, the functions `create_Point` and `distance` have direct access to the integer
-fields in any `Point`'s tuple.  However, any file that includes this file does not.
+## `Type` v. `Newtype` | (Trasparent v. Opaque)
+
+Looking at the earlier example, being in the same source file as the alias definition, the functions `create_Point` and `distance` have direct access to the `float` fields in any `Point`'s tuple.  However, any file that includes this file does not.
 
 Similarly, if a source file defines the following opaque alias:
 
--- a/guides/hack/11-built-in-types/16-string.md
+++ b/guides/hack/11-built-in-types/16-string.md
@@ -110,3 +110,9 @@ HHVM currently supports UTF-8, and single-byte encodings that are supported by t
 Other encodings may be supported in the future, however `Locale\set_native()` is likely to be restricted to
 the current locales; for example, UTF-16 can not be supported by `Locale\set_native()`, as UTF-16 strings can
 contain null bytes.
+
+# Working with `Regex\`
+Similar to `Str\`, functions in the `Regex\` namespace operate on bytes, not characters, unless `/u` is present in the pattern.
+
+- `Regex\replace("\u{1f600}", re"/./", 'Char')` is `CharCharCharChar`
+- `Regex\replace("\u{1f600}", re"/./u", 'Char')` is `Char`
--- a/guides/hack/13-contexts-and-capabilities/02-local-operations.md
+++ b/guides/hack/13-contexts-and-capabilities/02-local-operations.md
@@ -10,7 +10,7 @@ Consider the following potential (although not necessarily planned) contexts (wi
 * `write_prop`, representing the permission to mutate properties of objects
 * `dynamic`, representing the permission to cast a value to the `dynamic` type
 
-In all of the below cases, the relevant local operations are only legal due to the existence of the matching capabilities within the context of the function. In the world where all these contexts (and matching capabilities exist), some or all may be included within the `defaults` context.
+In all of the below cases, the relevant local operations are only legal due to the existence of the matching capabilities within the context of the function.
 
 ```hack
 function io_good()[io]: void {
--- a/build/extracted-examples/guides/hack/10-types/85-type-refinement/f5.hack
+++ b/build/extracted-examples/guides/hack/10-types/85-type-refinement/f5.hack
@@ -6,7 +6,7 @@ namespace HHVM\UserDocumentation\Guides\Hack\Types\TypeRefinement\F5;
 class C {
   private ?int $p = 8; // holds an int, but type is ?int
   public function m(): void {
-    if ($this->p is int) { // type refinement occurs; $this->p1 is int
+    if ($this->p is int) { // type refinement occurs; $this->p is int
       $x = $this->p << 2; // allowed; type is int
       $this->n(); // could involve a permanent type refinement on $p
       //      $x = $this->p << 2;   // disallowed; might no longer be int
--- a/guides/hack/10-types/85-type-refinement.md
+++ b/guides/hack/10-types/85-type-refinement.md
@@ -73,7 +73,7 @@ case, which involves a property:
 class C {
   private ?int $p = 8; // holds an int, but type is ?int
   public function m(): void {
-    if ($this->p is int) { // type refinement occurs; $this->p1 is int
+    if ($this->p is int) { // type refinement occurs; $this->p is int
       $x = $this->p << 2; // allowed; type is int
       $this->n(); // could involve a permanent type refinement on $p
       //      $x = $this->p << 2;   // disallowed; might no longer be int
@@ -83,6 +83,6 @@ class C {
 }
 ```
 
-Inside the true path of the `if` statement, even though we know that `$this->p1` is an `int` to begin with, once any method in this class
+Inside the true path of the `if` statement, even though we know that `$this->p` is an `int` to begin with, once any method in this class
 is called, the implementation must assume that method could have caused a type refinement on anything currently in scope. As a result,
 the second attempt to left shift is rejected.
--- a/guides/hack/15-asynchronous-operations/14-concurrent.md
+++ b/guides/hack/15-asynchronous-operations/14-concurrent.md
@@ -38,8 +38,8 @@ $x = 0;
 try {
   concurrent {
     $x = await async { return 1; };
-    await async { throw new Exception('foo'); }
-    await async { throw new Exception('bar'); }
+    await async { throw new Exception('foo'); };
+    await async { throw new Exception('bar'); };
   }
 } catch (Exception $e) {
   var_dump($x, $e->getMessage());
--- a/guides/hack/20-attributes/07-predefined-attributes.md
+++ b/guides/hack/20-attributes/07-predefined-attributes.md
@@ -59,8 +59,9 @@ plus an ellipsis indicating a trailing variable-argument list.
 
 ## __Deprecated
 
-This attribute can be applied to a function to indicate that it has been *deprecated*; that is, it is obsolete, and calls to it should be
-revised. This attribute has two possible attribute values.  Consider the following example:
+If you mark a function with `__Deprecated`, the Hack typechecker will find all static invocations of that function and mark them as errors so that you can find those invocations and fix them as needed. For runtime invocations, any function marked with `__Deprecated` will still be called successfully, but with runtime logging so that you can find and fix those dynamic invocations later.
+
+Consider the following example:
 
 ```Hack
 <<__Deprecated("This function has been replaced by do_that", 7)>>
@@ -68,10 +69,9 @@ function do_this(): void { /* ... */ }
 
 ```
 
-The presence of this attribute on a function has no effect, unless that function is actually called, in which case, for each call to that
-function, the type checker issues a diagnostic containing the text from the first attribute value.  The optional `int`-typed second attribute
-value (in this case, 7) indicates a *sample rate*. Assuming the program will still execute, every 1/sample-rate calls (as in, 1/7) to that
-function will be diagnosed at runtime.
+* For each dynamic call to `do_this()`, the typechecker will log a diagnostic containing the message from the attribute value.
+* The optional `int` indicates a *sample rate*, meaning that every `1/{sample-rate}` calls to that
+function will be diagnosed at runtime. In the example above, the sample rate is `1/7`.
 
 ## __DynamicallyCallable
 


guides/hack/10-types/79-type-aliases.md
```diff
@@ -33,8 +38,9 @@ function distance(Point $p1, Point $p2): float {
 }
 ```
 
-Being in the same source file as the alias definition, the functions `create_Point` and `distance` have direct access to the integer
-fields in any `Point`'s tuple.  However, any file that includes this file does not.
+## `Type` v. `Newtype` | (Trasparent v. Opaque)
```
lexidor, line 41: Thank you for fixing all these issues. This typo jumped out at me.

```suggestion
## `Type` v. `Newtype` | (Transparent v. Opaque)
```

guides/hack/11-built-in-types/16-string.md
```diff
@@ -110,3 +110,9 @@ HHVM currently supports UTF-8, and single-byte encodings that are supported by t
 Other encodings may be supported in the future, however `Locale\set_native()` is likely to be restricted to
 the current locales; for example, UTF-16 can not be supported by `Locale\set_native()`, as UTF-16 strings can
 contain null bytes.
+
+# Working with `Regex\`
+Similar to `Str\`, functions in the `Regex\` namespace operate on bytes, not characters, unless `/u` is present in the pattern.
```
lexidor, line 115: Not all functions in `Str\` operate on bytes. Many functions in that namespace have a locale aware version f.e. `Str\length_l()` which returns the length in characters instead of bytes.

guides/hack/10-types/79-type-aliases.md
AndrewDiMola: Thanks! Law of Conservation of Typos: they can not be created or destroyed...

guides/hack/11-built-in-types/16-string.md
AndrewDiMola: Good point. When I wrote this I was aware of the locale versions but wanted to say something like, "Similar to the default behavior of `Str\`", but maybe it's not a fair statement to say that is the default. At the very least reading this line on its own conveys a "this is always the truth" stance, which isn't true per the `_l()` variants.

AndrewDiMola: Do you think this is clear, if the preceding part is removed? Wondering if it could be more precise.

> Functions in the `Regex\` namespace operate on bytes, not characters, unless `/u` is present in the pattern.

lexidor: It is clear, but there are ways to be more precise. I don't know if this is an improvement though. (Flags are not always preceded by a `/`, since custom pattern delimiters like `#[ab]#u` are supported.)

Functions in the `Regex\` namespace operate on bytes (ascii strings). If the string being inspected is Utf8, you should use a `Regex\Pattern` with the `u` flag. Failing to do so may result in one multi-byte character being interpreted as multiple characters.

guides/hhvm/02-basic-usage/03-server.md
```diff
@@ -38,9 +47,8 @@ http://your.site:8080/index.php
 
 You can also use `curl` and other programs to access the HHVM server as well.
 
-### Possible Fatal Error
-
-If the code you are running is written in [Hack (<?hh)](/hack/) and you run into a [fatal error regarding not running the typechecker](/hhvm/FAQ/faq#running-code__how-do-i-fix-the-not-running-the-hack-typechecker-fatal-error), then you must do one of the following:
+## Debugging a Fatal Error
```
lexidor, line 50: The default value for `hhvm.hack.lang.look_for_typechecker` has been false for a _really_ long time. I don't know if this is still worth mentioning. Earlier hhvm versions (around 3.23 iirc) would refuse to run any Hack code if the typechecker was not running.

All code for hhvm has been Hack since 4.0, PHP is no longer supported. This sentence is an historical note only.

AndrewDiMola: @lexidor Thanks! That part did feel weird when I read it. Glad you pointed it out.

AndrewDiMola: @lexidor In that case, the preceding section should probably say something like:

`http://your.site:8080/index.hack` instead of `http://your.site:8080/index.php` (for an example of a client request to the server), right?

lexidor: Hack code can still be written in dot-php files. But we should indeed encourage dot-hack everywhere.

build/extracted-examples/guides/hack/12-generics/01-introduction/Stack.inc.hack
```diff
@@ -4,25 +4,23 @@
...
-    $this->stackPtr++;
   }
 
   public function pop(): T {
-    if ($this->stackPtr > 0) {
-      $this->stackPtr--;
-      return $this->stack[$this->stackPtr];
+    $stack = $this->stack;
+    if (!C\is_empty($stack)) {
+      return C\pop_backx(inout $stack);
```
lexidor, line 23: This method does not remove the last element of the stack property. You have made an implicit copy when assigning to a local variable. `C\pop_backx()` is altering this local copy, not the vec in the property.

AndrewDiMola:
A lot of issue here. Going to request a review and merge in before it gets even longer.

build/extracted-examples/guides/hack/12-generics/01-introduction/Stack.inc.hack
AndrewDiMola: Ah, you're right. I'm just returning the modified local. The reason I made a local was because `C\pop_backx()` said it had to operate on a local property, one marked as `inout`. Passing in the global was making the type checker angry.

AndrewDiMola: I could just do something like setting the global from the local, then return the value.

AndrewDiMola: @lexidor what do you think of:

```
  private vec<T> $stack;

  public function __construct() {
    $this->stack = vec[];
  }

  public function push(T $value): void {
    $this->stack[] = $value;
  }

  public function pop(): T {
    $stack = $this->stack;
    if (!C\is_empty($stack)) {
      $value = C\pop_backx(inout $stack);
      $this->stack = $stack;
      return $value;
    } else {
      throw new StackUnderflowException();
    }
  }
}
```

lexidor: Yes, that is correct. The typechecker disallows you from passing using a property as the "source" (not the correct term) for an inout. You can reassign the local variable to the property _or_ use a Vector of a vec. I am torn on which decision is best here. On the one hand, the docs clearly favor Hack arrays over collections everywhere else. On the other hand, making a copy to pop and then altering it incurs an `O(N)` cost, since hhvm needs to copy all the elements and increment their refcounts when you write to the local. A Vector is a reference type, so it does not do CoW.

lexidor: Sorry for leading you up a wrong path when I was suggesting you wouldn't need a stackPtr anymore. Doing this with a vec and no stackPtr incurs an O(N) cost on pop.

I have three implementations of a stack, a Vector based, dict based, and vec based stack.

The Vector based is the "best" stack. It does not hold on to elements that are already popped. The garbage collector will be able to collect elements that were previously only referenced in an "unreachable" section of the stack. The dict based stack is a minor correction on the old Stack from the example. By using a dict instead of a vec, push() will always be `[$next_index]=` instead of `[]=`. It is correct, but the memory layout of the elements is not contiguous. The vec based stack is the best you can do with Hack arrays. You need to keep track of the next unused index, since `[]=` must be used when appending to the end of a vec. You can't simply use `[$next_index]=`, this throws an OutOfBoundsException. So you need to pick the right form of the write operation depending on where the next element goes.

I have copied the implementations below.

```HACK
interface StackLike<T> {
    public function push(T $element): void;
    public function pop(): T;
}

final class VectorStack<T> implements StackLike<T> {
    private Vector<T> $elements = Vector {};

    public function push(T $element): void {
        $this->elements->add($element);
    }

    public function pop(): T {
        try {
            return $this->elements->pop();
        } catch (InvalidOperationException $_) {
            throw new StackUnderflowException();
        }
    }
}

final class DictStack<T> implements StackLike<T> {
    private dict<int, T> $elements = dict[];
    private int $stackPtr = 0;

    public function push(T $element): void {
        $this->elements[$this->stackPtr] = $element;
        $this->stackPtr++;
    }

    public function pop(): T {
        if ($this->stackPtr > 0) {
            $this->stackPtr--;
            return $this->elements[$this->stackPtr];
        }
        throw new StackUnderflowException();
    }
}

final class VecStack<T> implements StackLike<T> {
    private vec<T> $elements = vec[];
    private int $size = 0, $capacity = 0;

    public function push(T $element): void {
        if ($this->size === $this->capacity) {
            $this->elements[] = $element;
            $this->capacity++;
        } else {
            $this->elements[$this->size] = $element;
        }
        $this->size++;
    }

    public function pop(): T {
        if ($this->size > 0) {
            $this->size--;
            return $this->elements[$this->size];
        }
        throw new StackUnderflowException();
    }
}

final class StackUnderflowException extends Exception {}

<<__EntryPoint>>
function test(): void {
    $ops = vec<(function(StackLike<int>): mixed)>[
        $s ==> $s->pop(),
        $s ==> $s->push(1),
        $s ==> $s->pop(),
        $s ==> $s->push(1),
        $s ==> $s->push(2),
        $s ==> $s->push(3),
        $s ==> $s->pop(),
        $s ==> $s->pop(),
        $s ==> $s->pop(),
        $s ==> $s->push(1),
        $s ==> $s->push(2),
        $s ==> $s->pop(),
        $s ==> $s->push(3),
        $s ==> $s->pop(),
        $s ==> $s->pop(),
        $s ==> $s->pop(),
    ];

    $s = new VectorStack();
    foreach ($ops as $op) {
        try {
            $res = $op($s);
        } catch (StackUnderflowException $e) {
            $res = \get_class($e);
        }

        if ($res is arraykey) {
            echo $res."\n";
        } else {
            echo "[push]\n";
        }
    }
}
```

guides/hack/10-types/79-type-aliases.md
```diff
@@ -33,8 +38,9 @@ function distance(Point $p1, Point $p2): float {
 }
 ```
 
-Being in the same source file as the alias definition, the functions `create_Point` and `distance` have direct access to the integer
-fields in any `Point`'s tuple.  However, any file that includes this file does not.
+## `Type` v. `Newtype` | (Transparent v. Opaque)
```
Wilfred, line 41: I think we should spell out 'versus' rather than writing 'v.', and I'm not sure what `|` means here. How about "Choosing between `type` and `newtype`"?

Also, we should only put valid syntax between backticks, and `type` and `newtype` are case sensitive.

guides/hhvm/02-basic-usage/03-server.md
```diff
@@ -1,46 +1,48 @@
-Server mode is how you will use HHVM to serve web requests. The HHVM process starts up and continuously waits to serve web requests.
+Use HHVM server mode to create a HHVM process that continuously serves web requests, with these advantages:
 
-Multiple requests can, of course, be served simultaneously, and HHVM also caches code to be shared across requests as well.
+✔️ Multiple requests served simultaneously
```
Wilfred, line 3: Could we use plain bullet points here? I'm not sure if ticks will look good on all browsers.

guides/hhvm/02-basic-usage/03-server.md
```diff
@@ -1,46 +1,48 @@
...
+And we also could have overridden other defaults like the [server type](/hhvm/basic-usage/proxygen) or source root of your project files. For example:
 
-HHVM will also use the default INI configuration `server.ini` (normally found in `/etc/hhvm/` in Linux distros and `/usr/local/etc/hhvm/` in MacOS).
+```
+$ hhvm --mode server -d hhvm.server.type={$SERVER_TYPE} -d hhvm.server.source_root={$PROJECT_FOLDER}
+```
 
-## Client access to HHVM in server mode
+### INI Configuration Values
+HHVM uses the default [INI configuration](/hhvm/configuration/INI-settings) specified in `server.ini`, typically found at these locations:
```
Wilfred, line 37: If we can be definitive, users don't need to worry about whether their situation is typical. How about this?

"The default ini locations are ..."

guides/hack/05-functions/05-format-strings.md
```diff
@@ -1,6 +1,6 @@
-The Hack typechecker can also check that format strings are being used
-correctly.
+The Hack typechecker can check that format strings are being used correctly.
```
Wilfred, line 1: This is already a wording improvement, but "can" implies it's optional. How about "The Hack typechecker will check that ..." or "The Hack typechecker checks that ..."?

guides/hack/05-functions/05-format-strings.md
```diff
@@ -36,3 +35,19 @@ function use_it(): void {
...
 check that arguments match the type in the format string.
+
+## Format Specifiers
+See `PlainSprintf` for in-depth information on all format specifiers. 
+
+| Specifier | Expected Input | Expected Output                                                                                                     |
+|-----------|----------------|---------------------------------------------------------------------------------------------------------------------|
+| b         | int            | Binary (`63` => `111111`)                                                                                           |
+| c         | int            | ASCII character (`63` => `?`)                                                                                       |
+| d, u      | int            | Signed int (d: `-1` => `-1`) \| Unsigned int (u: `-1` => `18446744073709551615`)                                    |
```
Wilfred, line 46: Could we split `d` and `u` onto separate lines, since they give different output?

guides/hack/05-functions/05-format-strings.md
```diff
@@ -36,3 +35,19 @@ function use_it(): void {
...
+| e, E      | float          | Scientific Notation, as lowercase or uppercase (e: `0.34` => `3.4e` \| E: `0.34` => `3.4E`)                         |
+| f, F      | float          | Locale Floating Point Number (f) \| Non-Locale Floating Point Number (F)                                            |
+| g         | float          | Locale Floating Point Number or Scientific Notation (`3.141592653` => `3.14159` \| `3141592653` => `3.1415e+9`) (🥧) |
+| o         | int            | Octal (`63` => `77`)                                                                                                |
+| s         | string         | string                                                                                                              |
+| x, X      | int            | Hexadecimal, as lowercase or uppercase (x: `63` => `3f` \| X: `63` => `3F`)                                         |
```
Wilfred, line 52: Likewise here, I'm not sure what `|` means. In C-like languages, including Hack, `||` means logical OR and `|` means bitwise OR.

guides/hack/11-built-in-types/16-string.md
```diff
@@ -110,3 +110,9 @@ HHVM currently supports UTF-8, and single-byte encodings that are supported by t
 Other encodings may be supported in the future, however `Locale\set_native()` is likely to be restricted to
 the current locales; for example, UTF-16 can not be supported by `Locale\set_native()`, as UTF-16 strings can
 contain null bytes.
+
+# Working with `Regex\`
+Functions in the `Regex\` namespace operate on bytes (ASCII strings). If the string being inspected is UTF-8, use a pattern with the `u` flag. Failing to do so may result in one multi-byte character being interpreted as multiple characters. For example:
```
Wilfred, line 115: Bytes can be anything, including UTF-8 strings or ASCII strings.
```suggestion
Functions in the `Regex\` namespace operate on bytes. If the string being inspected is UTF-8, use a pattern with the `u` flag. Failing to do so may result in one multi-byte character being interpreted as multiple characters. For example:
```

guides/hack/13-contexts-and-capabilities/02-local-operations.md
```diff
@@ -10,7 +10,7 @@ Consider the following potential (although not necessarily planned) contexts (wi
 * `write_prop`, representing the permission to mutate properties of objects
 * `dynamic`, representing the permission to cast a value to the `dynamic` type
 
-In all of the below cases, the relevant local operations are only legal due to the existence of the matching capabilities within the context of the function. In the world where all these contexts (and matching capabilities exist), some or all may be included within the `defaults` context.
+In all of the below cases, the relevant local operations are only legal due to the existence of the matching capabilities within the context of the function.
```
Wilfred, line 13: This is definitely an improvement too, but perhaps we should avoid 'legal'? How about this?

```suggestion
In all of the below cases, the relevant local operations typecheck due to the matching capabilities within the context of the function.
```

guides/hack/10-types/79-type-aliases.md
```diff
@@ -33,8 +38,9 @@ function distance(Point $p1, Point $p2): float {
...
+
+Looking at the earlier example, being in the same source file as the alias definition, the functions `create_Point` and `distance` have direct access to the `float` fields in any `Point`'s tuple.  However, any file that includes this file does not.
```
fredemmott, line 43: include/require are very rarely used nowadays and pretty discouraged; perhaps something along the lines of 'no other file' instead?

guides/hack/11-built-in-types/16-string.md
fredemmott, line 115: Just checking: `Regex\` renders correctly?

build/extracted-examples/guides/hack/12-generics/01-introduction/Stack.inc.hack
lexidor: I feel I need to apologize once again for suggesting using `C\pop_backx()`. When I suggested this on #1171 I was thinking about `return C\pop_backx(inout $this->stack)`. This is not allowed. I misremembered the rules about which arguments qualify. Please do not use `C\pop_backx()` in this Stack example. The dance with the local variable incurs CoW, and therefore makes `->pop()` an order N function. Someone somewhere is going to copy this Stack into their code and blow through their response time budgets with stacks larger than 10,000 elements. I have ran a benchmark of VecStack, VectorStack, and MyStack (called PopBackStack). I timed how long it took to pop a stack of `n` elements until it is empty. The 2^10 rows operate on 1024 elements, the 2^16 rows operate on 65535 elements. The exact numbers are meaningless, but the trend and the comparative results do count.

Running in CLI mode

```
VectorStack   2^10 small 11.14503 ms >> __hhvm_launder_value(523776)
VecStack      2^10 small 6.334611 ms >> __hhvm_launder_value(523776)
PopBackStack  2^10 small 14.475042 ms >> __hhvm_launder_value(523776)
VectorStack   2^11 small 4.915844 ms >> __hhvm_launder_value(2096128)
VecStack      2^11 small 4.439112 ms >> __hhvm_launder_value(2096128)
PopBackStack  2^11 small 20.500871 ms >> __hhvm_launder_value(2096128)
VectorStack   2^12 small 0.314585 ms >> __hhvm_launder_value(8386560)
VecStack      2^12 small 0.289365 ms >> __hhvm_launder_value(8386560)
PopBackStack  2^12 small 38.393394 ms >> __hhvm_launder_value(8386560)
VectorStack   2^13 small 0.569701 ms >> __hhvm_launder_value(33550336)
VecStack      2^13 small 0.63312 ms >> __hhvm_launder_value(33550336)
PopBackStack  2^13 small 143.488665 ms >> __hhvm_launder_value(33550336)
VectorStack   2^14 small 1.611967 ms >> __hhvm_launder_value(134209536)
VecStack      2^14 small 1.313712 ms >> __hhvm_launder_value(134209536)
PopBackStack  2^14 small 545.340548 ms >> __hhvm_launder_value(134209536)
VectorStack   2^15 small 2.074609 ms >> __hhvm_launder_value(536854528)
VecStack      2^15 small 2.45884 ms >> __hhvm_launder_value(536854528)
PopBackStack  2^15 small 2147.56281 ms >> __hhvm_launder_value(536854528)
VectorStack   2^16 small 4.20304 ms >> __hhvm_launder_value(2147450880)
VecStack      2^16 small 4.763756 ms >> __hhvm_launder_value(2147450880)
PopBackStack  2^16 small 7962.39552 ms >> __hhvm_launder_value(2147450880)
```

Running in a webserver with a warm jit (did not speed up PopBackStack)

```
VectorStack   2^10 small 0.09595 ms >> __hhvm_launder_value(523776)
VecStack      2^10 small 0.067733 ms >> __hhvm_launder_value(523776)
PopBackStack  2^10 small 1.962322 ms >> __hhvm_launder_value(523776)
VectorStack   2^11 small 0.110098 ms >> __hhvm_launder_value(2096128)
VecStack      2^11 small 0.121941 ms >> __hhvm_launder_value(2096128)
PopBackStack  2^11 small 7.672828 ms >> __hhvm_launder_value(2096128)
VectorStack   2^12 small 0.185667 ms >> __hhvm_launder_value(8386560)
VecStack      2^12 small 0.221244 ms >> __hhvm_launder_value(8386560)
PopBackStack  2^12 small 33.073805 ms >> __hhvm_launder_value(8386560)
VectorStack   2^13 small 0.360809 ms >> __hhvm_launder_value(33550336)
VecStack      2^13 small 0.630479 ms >> __hhvm_launder_value(33550336)
PopBackStack  2^13 small 125.909002 ms >> __hhvm_launder_value(33550336)
VectorStack   2^14 small 0.700814 ms >> __hhvm_launder_value(134209536)
VecStack      2^14 small 0.847374 ms >> __hhvm_launder_value(134209536)
PopBackStack  2^14 small 487.473917 ms >> __hhvm_launder_value(134209536)
VectorStack   2^15 small 1.515368 ms >> __hhvm_launder_value(536854528)
VecStack      2^15 small 1.660033 ms >> __hhvm_launder_value(536854528)
PopBackStack  2^15 small 2020.03276 ms >> __hhvm_launder_value(536854528)
VectorStack   2^16 small 3.364754 ms >> __hhvm_launder_value(2147450880)
VecStack      2^16 small 3.381403 ms >> __hhvm_launder_value(2147450880)
PopBackStack  2^16 small 8155.71524 ms >> __hhvm_launder_value(2147450880)
```

The bench script

```HACK
<<__EntryPoint>>
function test(): void {
    \header('Content-Type: text/plain');

    for ($i = 3; $i < 7; ++$i) {
        bench(fill(new VectorStack(), $i), $i, 'small');
        bench(fill(new VecStack(), $i), $i, 'small');
        bench(fill(new PopBackStack(), $i), $i, 'small');
    }
}

final class TestObject {
    public function __construct(private int $value) {}

    public function getValue(): int {
        return $this->value;
    }
}

function fill(
    StackLike<TestObject> $stack,
    int $powers_of_ten,
): StackLike<TestObject> {
    $depth = (int)(10 ** $powers_of_ten);

    for ($i = 0; $i < $depth; ++$i) {
        $stack->push(new TestObject($i));
    }

    return $stack;
}

function bench(
    StackLike<TestObject> $stack,
    int $powers_of_ten,
    string $kind_of_object,
): void {
    $start = clock_gettime_ns(CLOCK_MONOTONIC);
    $total = 0;
    try {
        for (; ; ) {
            $total += $stack->pop()->getValue();
        }
    } catch (StackUnderflowException $e) {
        $end = clock_gettime_ns(CLOCK_MONOTONIC);
        echo Str\format(
            "%-13s 10e%d %s %.9g ms >> __hhvm_launder_value(%d)\n",
            \get_class($stack),
            $powers_of_ten,
            $kind_of_object,
            ($end - $start) / 1000000.,
            $total,
        );
    }
}
```

guides/hack/10-types/79-type-aliases.md
```diff
@@ -2,17 +2,22 @@ We can create an alias name for a type, and it is common to do so for non-trivia
...
 
+## Quickstart
 A type alias can be created in two ways: using `type` and `newtype`.
 
 ```Hack
 type Complex = shape('real' => float, 'imag' => float);
 newtype Point = (float, float);
 ```
 
+A type alias can also be created with [Generics](/hack/generics/introduction) as parameters.
```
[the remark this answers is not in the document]
AndrewDiMola: hmm -- this was in response to https://github.com/hhvm/user-documentation/issues/1147. But if it's not true I can remove.

guides/hack/10-types/79-type-aliases.md
AndrewDiMola: That's fair. Stylistically, I always write `v.` versus `vs.` but the whole word is probably better, and less confusing. Same for `|` as shorthand for or. Good feedback all around.

guides/hhvm/02-basic-usage/03-server.md
AndrewDiMola: Sure. Was trying to get fancy, but didn't think of that as a concern.

guides/hack/05-functions/05-format-strings.md
AndrewDiMola: `The Hack typechecker checks that` is better. Updating.

guides/hack/05-functions/05-format-strings.md
AndrewDiMola: Sorry, it's my shorthand for or. Probably something that only makes sense in my head.

guides/hack/11-built-in-types/16-string.md
AndrewDiMola: image

AndrewDiMola: Ideally, when the namespace reference feature is live, I'd like this to automatically link to that page, much like other functions link directly to their reference page.

build/extracted-examples/guides/hack/12-generics/01-introduction/Stack.inc.hack
AndrewDiMola: For the Vector code, also a good candidate to add to API-Examples for this page, to render an Examples section with that code.

https://docs.hhvm.com/hack/reference/class/HH.Vector/

And the dict and vec implementations might be great to add here, at the bottom of the page: https://docs.hhvm.com/hack/arrays-and-collections/vec-keyset-and-dict

AndrewDiMola: @lexidor That benchmark is fascinating. I almost want to formalize and add it somewhere as an example of better understanding why you should are about O(n), but that's probably outside the scope of these docs.

Unless anyone objects, I'll use the Vector code, but thank you very much for your implementation @Wilfred.

fredemmott: `Vector`, `Map`, and `Set` are strongly discouraged - we should try really hard to avoid them except when documenting the specific class.

This is both out of a general 'they're replaced by Hack arrays' thing, but also that 'hack has too many array types, when would I use which?' is a common question - the answer is use hack arrays unless you're working on existing code that uses something else.

AndrewDiMola: Good point. I'll use the vec implementation.

guides/hack/10-types/79-type-aliases.md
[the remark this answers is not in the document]
Wilfred: Oh, gotcha. I interpreted this sentence as "you can use generics to create a type alias" whereas #1147 want to clarify that "a type alias can include generics".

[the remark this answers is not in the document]
AndrewDiMola: Fixed

fredemmott (approved):
Why are `size` and `capacity` tracked in the `Stack` class instead of using `C\count()` where needed?

Additionally, anything in the vec at a position > capacity will be leaked

I'm fine with these variables if there is an educational intent behind them, but I'm not currently seeing this in the text.

If there's not a specific desire for having the size and capacity properites in `VecStack`, please remove both before landing

AndrewDiMola:
I think that was initially over concerns for `pop_back`. I'll try a different implementation.

Was gonna try something like this: not working at the moment, gonna debug / rewrite tomorrow:

```
use namespace HH\Lib\C;

interface StackLike<T> {
    public function push(T $element): void;
    public function pop(): T;
}

class StackUnderflowException extends \Exception {}

class VecStack<T> implements StackLike<T> {
    private vec<T> $elements = vec[];

    public function push(T $element): void {
        $this->elements[] = $element;
    }

    public function pop(): T {
        $count = C\count($this->elements);
        if ($count > 0) {
            $element = $this->elements[$count - 1];
            Vec\drop($this->elements, $count - 1);
            return $element;
        }
        throw new StackUnderflowException();
    }
}
```

fredemmott:
`$this->elements = Vec\take($this->elements, $count - 1);`

PR #1174, opened 2022-03-12
Switch 'hello, world' to HSL IO from PHP inherited stuff

fredemmott:
- PHP functionality like this is already undocumented in the reference
  docs
- new users should be pointed towards the best practice
- if HSL IO is used anywhere else, the PHP stuff stops working, as HSL
  IO sets `O_NONBLOCK`, which the PHP functionality isn't aware of

refs sunfishcode/hello-world-vs-io-errors#3


Atry:
I think we should have a free function to write strings to stdout asynchronously. It does not feel good to write such a long statement in Hello World.

Atry (approved):

fredemmott:
Yep  - in general, convenience of the HSL for super-common stuff is an area that needs work

AndrewDiMola:
> if HSL IO is used anywhere else, the PHP stuff stops working, as HSL
IO sets O_NONBLOCK, which the PHP functionality isn't aware of

Interesting, didn't know this.

Chaining all those just for a hello world output feels kinda bad. Is there any way to recommend an alias (an echo-like) if that's truly the easiest way to output something? That does seem scary.

fredemmott:
There should probably also be something like var_dump for debugging, but there's a few design questions:

Do we actually want to make it easier? I think this is a relatively clear yes, but we should explicitly make this decision. c++ `<iostream>` would be an example of choosing 'no'

There's then several much less clear questions:

- does it act like an async function? i.e. is `await` or `\HH\Asio\join()` required?
- if `await` or `HH\Asio\join()` is not required, does it:
  - temporarily flip off O_NONBLOCK, do the write, then restore the previous value?
  - consistently act as if `HH\Asio\join()` was passed?
  - act as if `await` was used in async functions, `HH\Asio\join()` in non-async functions?
  - require `await` in async functions, implicitly join in non-async-functions?
- keyword, normal function, or auto-imported function?
  - this ties into whether or not it is 'magic', e.g. implicit prep
  - also is a tradeoff between usability and encouraging doing things 'correctly'
  
 Also, to clarify, when `O_NONBLOCK` is set, PHP IO can partially succeed (including 0 bytes written) when without they would block until it is written. For example, `echo $ten_mb_blob` would usually output all 10mb, but with O_NONBLOCK could write 0 bytes. If nothing else has been written, 1kb, 4kb, and 16kb are fairly common real-world results.

lexidor:
As it stands, `IO\request_output()->writeAllAsync()` is synchronous behind the scenes. https://github.com/hhvm/hsl/blob/main/src/io/_Private/ResponseWriteHandle.php

We really should not be depending on that btw. Having other Awaitables able to progress whilst writing to the output would decrease the wall time on write heavy programs.

I am not convinced adding auto-await magic would be a good idea. Hack (and JS) devs have been taught, an Async function can not be interrupted between awaits. It can only be suspended at an await point. If the IO intrinsic could suspend an async function, this may cause confusion.

fredemmott:
> is synchronous behind the scenes.

Only sometimes :)

https://github.com/hhvm/hsl/blob/main/src/io/stdio.php#L28

guides/hack/01-getting-started/01-getting-started.md
```diff
@@ -30,9 +30,11 @@ $ touch .hhconfig
 Create a file called `my_project/hello.hack` with the following code:
 
 ```hello.hack no-auto-output
+use namespace HH\Lib\IO;
+
 <<__EntryPoint>>
 function main(): void {
-  echo "Hello World!\n";
+  await IO\request_output()->writeAllAsync("Hello World!\n");
```
lexidor, line 37: You cannot await an a non-async function ;)

lexidor (changes requested):

PR #1176, opened 2022-03-15
Documentation for abstract enum class

vsiles:
Not a big change, just documenting abstract enum classes.
HHVM should allow them without the experimental flag since the last
release so I didn't document that part. Let's see how CI likes it

EDIT: CI doesn't like it much, but let's review the text before and wait for an update / add the temp experimental flag in a second pass.


vsiles:
@fredemmott @AndrewDiMola  @Wilfred Any feedback on this update ?

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -169,6 +169,33 @@ If either `D1`, `D2` or `D3` tries to define a constant named `Age`, there will
 
 Though the `final` keyword is not supported, Enum classes support the [`__Sealed`](../attributes/predefined-attributes#__sealed) attribute. Using `__Sealed`, you can specify which other enum classes, if any, are allowed to extend from your enum class.
 
+
+## Abstract enum classes
+
+Like regular classes, enum classes come in two flavors: a concrete one and an abstract one. An abstract enum class can declare abstract members, where only their type and name are provided.
```
AndrewDiMola, line 175:
```suggestion
Like regular classes, enum classes come in two flavors: concrete and abstract. An abstract enum class can declare abstract members (properties and methods), where only their type and name are provided.
```
I like being explicit when referring to what a member is, because that was confusing to me when I was a new reader.

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -169,6 +169,33 @@ If either `D1`, `D2` or `D3` tries to define a constant named `Age`, there will
...
+```EnumClassIntro.Abstract.hack no-auto-output
+// abstract enum class with some abstract members
+abstract enum class AbstractNames : IHasName {
+  abstract HasName Foo;
+  HasName Bar = new HasName('bar');
+}
+```
+
+Abstract members do not support default values, and can't be accessed directly. They only map a name to a type.
+One must extends abstract enum classes into concrete ones, and provide all the abstract members to build a valid enum class.
```
AndrewDiMola, line 186:
```suggestion
For an enum class with abstract members to successfully build and typecheck, you must extend your [abstract class](/hack/classes/inheritance#abstract-classes) with implementations of all abstract members.
```
Switching to active voice and rewording because "One must extends abstract enum classes" felt a bit off. If any of my other suggestions are not strictly correct, feel free to ignore.

guides/hack/11-built-in-types/35-enum-class.md
```diff
@@ -169,6 +169,33 @@ If either `D1`, `D2` or `D3` tries to define a constant named `Age`, there will
...
+
+```EnumClassIntro.Concrete.hack no-auto-output
+enum class ConcreteNames : IHasName extends AbstractNames {
+  HasName Foo = new HasName('foo'); // one must provide all the abstract members
+  // Bar is inherited from AbstractNames
+}
+```
+
+All concrete members are inherited, and can't be redefined.
```
AndrewDiMola, line 195: I'm wondering if overridden is the correct term here.

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -132,6 +132,10 @@ enum class F : Foo {
 // E#A === F#A is true
 ```
 
+## Enum class labels and abstract enum classes
+Abstract enum classes support labels like any other enum class. The main difference is that an abstract enum class only provides the `nameOf` static method.
+Since some of its members may be abstracted away, they do no provide the `valueOf()` or `getValues()` static methods.
```
AndrewDiMola, line 137: I recommend replacing the word 'they'. When talking of two topics, the reader might be confused about what specifically is being referred to, especially with 'its' as the previous pronoun.

guides/hack/11-built-in-types/36-enum-class-label.md
```diff
@@ -161,4 +165,4 @@ Secondly, `HH\MemberOf` is covariant in `TType`. Since `B extends A`, it means t
...
 
-On the other hand, `HH\EnumClass\Label` is invariant in `TType`. It means that while `G#Y` is of type `HH\EnumClass\Label<G, B>`, it is not of type `HH\EnumClass\Label<G, A>`. Labels are opaque handles to access data; one can think about them as maps from names to types. Their typing has to be more strict, especially if we want to be able to extend this concept to other parts of a class (reflection like access to methods, properties, …). To make sure these possible extensions remain possible, we enforce a stricter typing for labels then for values.
\ No newline at end of file
+On the other hand, `HH\EnumClass\Label` is invariant in `TType`. It means that while `G#Y` is of type `HH\EnumClass\Label<G, B>`, it is not of type `HH\EnumClass\Label<G, A>`. Labels are opaque handles to access data; one can think about them as maps from names to types. Their typing has to be more strict, especially if we want to be able to extend this concept to other parts of a class (reflection like access to methods, properties, …). To make sure these possible extensions remain possible, we enforce a stricter typing for labels then for values.
```
AndrewDiMola, line 168:
```suggestion
On the other hand, `HH\EnumClass\Label` is invariant in `TType`. It means that while `G#Y` is of type `HH\EnumClass\Label<G, B>`, it is not of type `HH\EnumClass\Label<G, A>`. Labels are opaque handles to access data; one can think about them as maps from names to types. Their typing has to be more strict, especially if we want to be able to extend this concept to other parts of a class (reflection like access to methods, properties, …). To make sure these possible extensions remain possible, we enforce a stricter typing for labels than for values.
```

Quick grammar fix. Also, in general, I prefer using 'you' over 'one' to help keep the documentation set as one voice.

AndrewDiMola (approved):

guides/hack/11-built-in-types/35-enum-class.md
vsiles: Ok, so I'll be more explicit, but it will be constants. Enum classes don't have properties or methods :)

guides/hack/11-built-in-types/35-enum-class.md
vsiles: Ok, I'll rephrase this. But you're confusing classes and enum classes. I'll ping you once this is updated.

guides/hack/11-built-in-types/35-enum-class.md
vsiles: You mean using override instead of redefined. Yes, that's probably more accurate.

vsiles:
@AndrewDiMola Have a look at my updates when you can and let's ship this next week, no rush ;)

AndrewDiMola (approved):
Looks good. Thanks, @vsiles!

PR #1177, opened 2022-03-16
Improve names and remove HH_FIXMEs in UNSAFE_CAST example

Wilfred:


AndrewDiMola:
Looks like you need to rebuild for the test files but otherwise seems good.

PR #1180, opened 2022-03-21
enclose main function in an async context

AndrewDiMola:
If I understand right, should be an async function. 

Taken from: https://docs.hhvm.com/hack/XHP/introduction#runtime-validation

## Before
```
<<__EntryPoint>>
function intro_examples_tag_matching_validation_run(): void {
  intro_examples_tag_matching_validation_using_string();
  await intro_examples_tag_matching_validation_using_xhp();
}
```
## After
```
<<__EntryPoint>>
async function intro_examples_tag_matching_validation_run(): Awaitable<void> {
  intro_examples_tag_matching_validation_using_string();
  await intro_examples_tag_matching_validation_using_xhp();
}
```


fredemmott (approved):
Tests need fixing, otherwise looks good

PR #1183, opened 2022-03-24
Make VecStack::pop() constant time

lexidor:
The new implementation of VecStack introduced in #1173 uses `Vec\take()` in the implementation of `function pop(): T`. `Vec\take()` has a linear time complexity:
```HACK
/**
 * Returns a new vec containing the first `$n` elements of the given
 * Traversable.
 *
 * To drop the first `$n` elements, see `Vec\drop()`.
 *
 * Time complexity: O(n), where n is `$n`
 * Space complexity: O(n), where n is `$n`
 */
function take<Tv>(
  Traversable<Tv> $traversable,
  int $n,
)[]: vec<Tv>;
```

With a reasonable use (pushing and popping 30000 elements) the `Vec\take()` variant takes 1.8 seconds, whereas the `$this->stackPtr` variant takes 8 milliseconds. This makes the difference between having a fast page render and a human perceivable slowdown when rendering. Yes, this stack is not used in this repository. However, it is not unthinkable that someone browsing the docs will copy-paste this stack into their own project. Trying to chase down what is slowing down the page will be a nightmare, since no one suspects this simple class to be the main performance bottleneck on a page. This is likely to only happen in production as well, since the datasets are larger there.

Keeping track of a stack pointer, and adding an `if (vec is full)` check in `push()` is not pretty, but cuts the runtime of pop down to constant time from linear time. (benchmark provided in the comment below).

The assignment `$this->elements[$this->stackPtr] = $this->elements[0];` prevents leaking objects past the stack pointer. We free up the actual value held at that index, at the cost of incrementing the reference count on the first element.
```
buffer vec[]
ptr -1 ^
push obj(A), buffer vec[A]
ptr 0                   ^
push obj(B), buffer vec[A, B]
ptr 1                      ^
push obj(C), buffer vec[A, B, C]
ptr 2                         ^
pop -obj(C), buffer vec[A, B, A]
ptr 1                      ^
pop -obj(B), buffer vec[A, A, A]
ptr 0                   ^
pop -obj(A), buffer vec[A, A, A]
ptr -1              ^
```

The copied A values behind the stack pointer can not be observed, since the only way to increment the stack pointer is by pushing an element. Pushing overwrites the copied value with a fresh value. The A value can not be freed until either the entire stack is freed or until we push D, E, and F (overwriting all the copied A's). The alternative would be adding an `if( stack is emptied by this pop ) { $this->elements = vec[]; }`. This would free A. But this means the vec would have to dynamically regrow again when pushing new values, so leaking A has an advantage when the stack is reused.


AndrewDiMola:
I'm happy with this update and it looks good to me, but will defer to @fredemmott or @Wilfred in case there are any style updates, or other pieces of feedback on the code.

Atry:
I think we should use `C\pop_back` for best performance.

https://docs.hhvm.com/hsl/reference/function/HH.Lib.C.pop_back/

Atry:
@lexidor Did you run your benchmark in [Repo mode](https://docs.hhvm.com/hhvm/advanced-usage/repo-authoritative)? The performance would be different from CLI.

AndrewDiMola:
some more context: https://github.com/hhvm/user-documentation/pull/1173#discussion_r827453871

Atry:
I think we can use a generator to pass the `inout` parameter to `pop_backx`, given that class properties are not allowed for `inout` parameters.

``` hack
function stackGenerator<T as nonnull>(vec<T> $stack): Generator<mixed, ?T, ?T> {
  $pop_or_push = yield;
  for (;;) {
    if ($pop_or_push is null) {
      $pop_or_push = yield C\pop_backx($stack);
    } else {
      $stack[] = $pop_or_push;
      $pop_or_push = yield;
    }
  }
}
```

lexidor:
> I think we can use a generator to pass the `inout` parameter to `pop_backx`, given that class properties are not allowed for `inout` parameters.
> 
> ```hack
> function stackGenerator<T as nonnull>(vec<T> $stack): Generator<mixed, ?T, ?T> {
>   $pop_or_push = yield;
>   for (;;) {
>     if ($pop_or_push is null) {
>       $pop_or_push = yield C\pop_backx($stack);
>     } else {
>       $stack[] = $pop_or_push;
>       $pop_or_push = yield;
>     }
>   }
> }
> ```

```HACK
$s = new GeneratorStack();
$s->push(null); // StackUnderflowException
```

You'd need to either make `->push()` send an operation struct or restrict `T as nonnull`. The Generator you proposed does not know whether the null is a value or the absence of a value.

PR #1186, opened 2022-03-25
Update to HHVM 4.154

fredemmott:
- trait changes were just text changes to the error text
- AsyncMysqlClient::adoptConnection was removed


AndrewDiMola (approved):
Version and everything else looks fine. Seems good.

PR #1188, opened 2022-03-30
Don't expect Cannot use unstable feature: `abstract_enum_class`

lexidor:
This expect does not match the behavior of 4.154.
This feature is not unstable anymore.
The expected output is an empty file.


Wilfred:
LGTM, thanks.

PR #1190, opened 2022-03-30
Improve sidebar user experience

AndrewDiMola:
This PR affects the sidebar in these ways:
* indents subpages
* adds line separators between section topics

## Current
image

## Proposed
image

## Issues
There is a line above the background-color when an element is :hover, which on close examination looks a bit weird, because I can't select the previous sibling to remove the :after pseudo-element. 

image

### DIFF
--- a/sass/navigation.scss
+++ b/sass/navigation.scss
@@ -27,6 +27,14 @@
       }
     }
 
+    .navGroup:after {
+      content: '';
+      display: block;
+      margin: 0 auto;
+      width: 80%;
+      border-bottom: 1px solid #DADADA;
+    }
+
     li {
       a {
         color: $lightgrey;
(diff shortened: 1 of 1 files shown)


lexidor (approved):
The indent is nice for sure. 👍

I have [no real preference](https://c.tenor.com/MJh5L-hYLDMAAAAd/futurama-no.gif) between the line less before and the after.

AndrewDiMola:
> The indent is nice for sure. 👍
> 
> I have no real preference between the line less before and the after.

haha, re: lines, @Wilfred said the same, but on an earlier iteration. I do feel like it helps if, as an exercise, you go to the live site and scroll down the sidebar. It feels a bit monotonous. While a better solution might be grouping by section (like /index), I feel like the lines are a good first step.

fredemmott (approved):

PR #1191, opened 2022-03-30
XHP Content Rearrange + Attribute Selection Operator [WIP]

### LINKED ISSUE #614: No documentation for XHP attribute selection operator

# Where is the problem?

Operator and expression documentation

# What is the problem?

no entry for `->:`

 - Build ID: 2019-03-12T19:42:23+0000:1d93eb8a38ec302c196e2aff88e3b8a3a10423ca
 - Page requested: /hack/expressions-and-operators/member-selection
 - Page requested at: Wed, 13 Mar 2019 19:02:48 +0000

AndrewDiMola:
# XHP
* Fixes #614 
* Moves Setup to before Introduction
* Reorders content on Setup page to make it clear Composer dependencies are necessary
* New section headers in Names page
* Remove [XHP-Bootstrap](https://github.com/facebookarchive/xhp-bootstrap) references

# Expressions and Operators
* New page on XHP attribute selection
* Refactor links so that they're added to the test suite
* Content (page links) grouped into new sections, rather than a long bulleted list
* Content (page links) changed to be more descriptive


guides/hack/21-XHP/01-setup.md
```diff
@@ -1,13 +1,41 @@
...
+  exception of `Facebook\XHP\ChildValidation`.
+
+All the following guides are written with the assumption that XHP namespace
+support is enabled and XHP-Lib v4 is used, but there are notes pointing out any
+major differences&mdash;look for **Historical note** sections.
+
+<span class="fbOnly fbIcon">XHP namespaces are not enabled in Facebook's WWW
+repository, so all **Historical note** sections apply.</span>
+
+## Enable XHP for older HHVM versions
```
lexidor, line 35: The heading isn't quite accurate here. This switch does not enable/disable XHP. It enables the `xhp class` rules, which makes xhp classes interoperate with namespaces.

guides/hack/21-XHP/01-setup.md
```diff
@@ -22,10 +50,16 @@ hhvm.hack.lang.enable_xhp_class_modifier=true
 hhvm.hack.lang.disable_xhp_element_mangling=true
 ```
 
+## Disable XHP
```
lexidor, line 53: Ditto

guides/hack/21-XHP/01-setup.md
lexidor: With mangling enabled, you declare classes in the v3/fb manner, using a colon prefix to indicate xhp-ness. This mangling turns `class :fb:hello-world` into `class xhp_fb__hello_world`. Xhp depends on the fact that this is `\xhp_fb__hello_world` and not `\SomeNamespace\xhp_fb__hello_world`. With mangling disabled (the default), you declare classes in the v4 manner, using the xhp class modifier. `xhp class fb:hello_world` becomes `CurrentNamespace\fb\hello_world`. Xhp no longer depends on the fact that the names get dumped into the root namespace. By changing this setting, you change the parsing rules. You are not disabling xhp (xhp is always enabled). In mangling mode `function f(:fb:block $b): void;` becomes `f(xhp_fb__block $b): void;`. In no-mangling mode, it become `function f(\CurrentNamespace\fb\block $b): void;`.

AndrewDiMola:
I took a lot of liberties on retitling page links in the operators section. Expecting much debate, feedback, and correction.

guides/hack/02-source-code-fundamentals/13-names.md
```diff
@@ -23,15 +23,19 @@ function compute(int $val): void {
...
 The name `$this` is predefined inside any instance method or constructor when that method is called from within an object context.
 `$this` is read-only and designates the object on which the method is being called, or the object being constructed. The type of
 `$this` is [`this`](../built-in-types/this.md).
 
+## Reserved Names
 Names beginning with two underscores (__) are reserved by the Hack language.
 
+## XHP Name Constraints
 Note that [XHP classes](../XHP/introduction) have different name
 constraints. Class names may contain `:`, and must start with
```
lexidor, line 40: Not your fault. This content describes the mangling variant of xhp classes. Their source names had to start with a colon and may contain intermediary colons and minuses. These days, a colon in an xhp type name is interpreted as a backslash and minuses are disallowed.

Before `:fb:display` meant `xhp_fb__display`.
Now it means `\fb\display`.

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,43 @@
...
-* [Scope resolution, `::`](scope-resolution.md)
-* [Spaceship, `<=>`](equality.md)
-* [String concatenation, `.`](string-concatenation.md)
-* [Subscripting, `[]`](subscript.md)
-* [Ternary, `?` `:`](ternary.md)
-* [Type Assertions `is`, `as`, `?as`](type-assertions.md)
-* [Yield](yield.md)
+## Assignment Operators
+* [Assignment](/hack/expressions-and-operators/assignment) (`=`, `+=`, and more)
+* [Bitwise](/hack/expressions-and-operators/bitwise-operators) (`&`, `|`, `^`, `<<`, `>>`, `~`)
```
lexidor, line 5: The operators in this list are not assignment operators. Did you mean `&=` and such? I do not know for sure which operators can be combined with assignment, so maybe `&=` does not exist.

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,43 @@
...
+* [Coalesce](/hack/expressions-and-operators/coalesce) (`??`, `??=`)
```
lexidor, line 6: `?? ` does not assign, only `??=` does.

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,43 @@
...
+* [Pipe](/hack/expressions-and-operators/pipe) (`|>` then stored in `$$`)
+* [Spaceship](/hack/expressions-and-operators/equality#the-spaceship-operator) (`<=>`)
```
lexidor, line 8: Spaceship does not assign. It returns an integer where the sign represents the greater than / less than relationship between the LHS and the RHS.

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,43 @@
...
+* [Ternary](/hack/expressions-and-operators/ternary) (`?` `:`, and `?:`)
```
lexidor, line 9: Ternary and elvis don't assign.

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,43 @@
...
+
+## Comparison Operators
+* [Comparison](/hack/expressions-and-operators/comparisons) (`>`, `>=`,`<`,`<=`)
+* [Equality Comparison](/hack/expressions-and-operators/equality) (`==`, `!=`, `===`, and more)
```
lexidor, line 13: Missing `!==`?

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,43 @@
...
+* [Unpack Tuples](/hack/expressions-and-operators/list) (`list`)
+
+## Function Operators
+* [Async: Suspend Execution](/hack/expressions-and-operators/await) (`await`)
+* [Define a Generator](/hack/expressions-and-operators/yield) (`yield`)
+* [Write to Standard Output](/hack/expressions-and-operators/echo) (`echo`)
+* [Terminate](/hack/expressions-and-operators/exit) (`exit`)
+
+## Error Operators
+* [Indicate Errors](/hack/expressions-and-operators/invariant) (`invariant`)
```
lexidor, line 42: invariant() is an unconditional assert().
Suggestion for the link text "check an invariant / assert and invariant"

guides/hack/03-expressions-and-operators/99-XHP-attribute-selection.md
```diff
@@ -0,0 +1,17 @@
+When working with [XHP](/hack/XHP/introduction), use the `->:` operator to retrieve an XHP Class [attribute](/hack/XHP/basic-usage#attributes) value.
+
+You can also use `$this->:` in place of the named class instance.
```
lexidor, line 3: Why does `$this` get called out here? `->:` can work on arbitrary expressions that resolve to an xhp object. E.g. function calls and operations like ($a ?? $b)->:.

guides/hack/21-XHP/01-setup.md
```diff
@@ -1,13 +1,41 @@
-## Namespace Support
-
-Support for namespaced XHP classes (elements) is enabled by default since
+Support for XHP and namespaced XHP classes (elements like `<p>`) is enabled by default since
```
lexidor, line 1: Hhvm has always supported xhp and it is unconditionally enable. 4.73 made no mangling mode the default.

guides/hack/21-XHP/01-setup.md
```diff
@@ -1,13 +1,41 @@
...
+While the XHP syntax is part of Hack, a large part of the implementation is in a
+library called [XHP-Lib](https://github.com/hhvm/xhp-lib/) that needs to be
+installed via [Composer](https://getcomposer.org/) (add `facebook/xhp-lib` to your `composer.json` manually
+or by running `composer require facebook/xhp-lib ^4.0` or `^3.0`).
+
+This includes the base classes and interfaces, and definitions of standard HTML elements.
+
+### XHP-Lib Version
+
+There are currently two major supported versions of XHP-Lib:
```
lexidor, line 17: Facebook stopped supporting v3 a while ago. It might be a good idea to reflect this in the docs. @fredemmott knows the state of v3. My interpretation of the project status is: Facebook does not recommend you use v3 as it requires non-default hhvm ini settings. V4 is strongly recommended. If v3 falls over in newer hhvm versions, it is up to the community to open a PR and to request a new release be made.

guides/hack/21-XHP/25-further-resources.md
```diff
@@ -1,5 +1,3 @@
 Besides our documentation, there are some other very useful resources to enhance your experience with XHP.
 * [XHP Library](https://github.com/facebook/xhp-lib): The class libraries for XHP.
-* [XHP Bootstrap Library](http://github.com/hhvm/xhp-bootstrap): A framework providing XHP components for web pages.
-* [Using XHP with Bootstrap](http://hhvm.com/blog/6515/using-xhp-with-bootstrap): A blog post about using XHP with the bootstrap library.
```
lexidor, line 4: If you removed this link because it 404'ed, please replace it with https://github.com/facebookarchive/xhp-bootstrap/ instead. From the project status:

We are not planning to actively maintain this project; however, we still believe it has value as an example.

guides/hack/21-XHP/07-basic-usage.md
```diff
@@ -154,6 +154,8 @@ is rendered (`toStringAsync()` is called) with any required attributes missing.
 With `check_xhp_attribute=true` (available since HHVM 4.8) this is also a
 typechecker error.
 
+Use the [`->:` operator](/hack/expressions-and-operators/XHP-attribute-selection) to select an attribute, or `$this->:` in place of the named class instance.
```
lexidor, line 157: Ditto, why is `$this->:` getting a special mention here?

guides/hack/02-source-code-fundamentals/13-names.md
lexidor: "must start with" is outdated and `:` is not a part of the name anymore. It is a namespace separator.

AndrewDiMola: oh, this is a good catch. I'll make a note about v3 distinctions.

guides/hack/21-XHP/25-further-resources.md
fredemmott: I feel these should be entirely removed - as well as being unmaintained:
- they're now considered a bad idea
- they target v3

fredemmott (approved):
Please address lexidor's comments, but nothing new to add

guides/hack/21-XHP/25-further-resources.md
AndrewDiMola: @lexidor - as already said, Fred and I had a conversation offline and are opting to fully remove.

guides/hack/21-XHP/01-setup.md
AndrewDiMola: Changing to "Enable XHP Class Rules" and "Disable XHP Class Rules" respectively.

guides/hack/02-source-code-fundamentals/13-names.md
AndrewDiMola: Removing the entire note on the Names page because current lib differences are more in the Namespace syntax.

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: whoops. good catch. got used to the "and more" phrase.

guides/hack/03-expressions-and-operators/99-XHP-attribute-selection.md
AndrewDiMola: Just providing additional context in case a given reader doesn't know of the alternative option. Will add your note and remove the special mention.

guides/hack/21-XHP/01-setup.md
AndrewDiMola: yes, my mistake. reverting.

guides/hack/21-XHP/01-setup.md
AndrewDiMola: Added a note on the status of both libraries.

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: In all of these, I was perhaps using too loose a description of assignment, as in anything that writes or updates a value.

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: Moved. Thank you!

guides/hack/21-XHP/07-basic-usage.md
AndrewDiMola: Removed.

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: If I could push back, aren't we using `??` to conditionally assign a value, depending on the result of the expressions?

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: I say this because we are "assigning" values in all these cases. 

If I could push back, In `+=` I consider the assignment to be the increment to the value at the variable. In `<<` the shift assigns a different value by shifting, etc. Same for other bitwise. Assign a value based on the result of binary operations.

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: Ditto for above notes. Should these instead be in an "Evaluation Operators" section? Perhaps with the two Coalesce operators split between Assignment and Evaluation?

AndrewDiMola:
Will push latest commit when I figure out why I'm causing the build to file? Likely I pushed the wrong expect files.

guides/hack/21-XHP/25-further-resources.md
lexidor: If we remove links to xhp-bootstrap, I should probably remove this one too.

https://github.com/hhvm/xhp-lib/commit/453325f48cf8be799d7332e0e90cbec127b88c4e#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R16

guides/hack/03-expressions-and-operators/01-introduction.md
lexidor: I see where our definitions differ. I don't consider "produces a new value" to be assignment. I consider giving a new value to a name assignment. We both agree that `$x += 5` is an assignment. In don't consider `$x << 5` to be an assignment, because `$x` is not altered. `$y = $x << 5` would be an assignment by my definition, because of the `$y = ...` part.

Hack disallows using assignment as an expression. So if `if (...) 0;` emits a typechecker error like `Assignments can no longer be used as expressions. Pull the assignment out into a separate statement.`, the syntax at the `...` is an assignment. `$x += 5` emits this error, `$x << 5` does not, and `$y = $x << 5` emits the error again. Calling `<<` an assignment may be cause for confusion.

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: changed mind and agree.

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: changed mind and agree.

lexidor (approved):
✅ You get a heading, you get a heading, every section gets a heading. Very useful when linking to the docs, as they can be linked to by their anchor (the hash at the end of urls like https://docs.hhvm.com/hack/silencing-errors/introduction#best-practices).

AndrewDiMola:
(Also some other minor updates because I realized I was unhappy with my initial rewrite of the Setup page)

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,46 @@
...
+* [Arithmetic](/hack/expressions-and-operators/arithmetic) (`+`, `-`, `*`, `/`, and more)
+* [Increment](/hack/expressions-and-operators/incrementing-and-decrementing) (`++`)
+* [Decrement](/hack/expressions-and-operators/incrementing-and-decrementing) (`--`)
+
+## Bitwise and Bit Shift Operators
+* [Bitwise](/hack/expressions-and-operators/bitwise-operators) (`&`, `|`, `^`, `<<`, `>>`, `~`)
+
+## Class and Member Operators
+* [Create an Object](/hack/expressions-and-operators/new) (`new`)
+* [Access Instance Members](/hack/expressions-and-operators/member-selection) (`->`, `?->`)
```
Wilfred, line 23: I think 'member' is a term best avoided FWIW. PHP uses the term, but it's sufficiently vague that it can cause confusion. 'methods and properties' means the same thing, and you can often specifically say 'methods' or 'properties'.

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,46 @@
...
+* [Access Static Members / Enum Properties](/hack/expressions-and-operators/scope-resolution) (`::`)
+* [Access XHP Attributes](/hack/expressions-and-operators/XHP-attribute-selection) (`->:`)
+
+## Built-in Type Operators
+* [Cast/Convert Types](/hack/expressions-and-operators/casting) (`(int)`, `(string)`, and more)
+* [Check Types](/hack/expressions-and-operators/type-assertions#checking-types-with-is) (`is`)
+* [Enforce Types](/hack/expressions-and-operators/type-assertions#enforcing-types-with-as-and-as) (`as`, `?as`)
+* [Concatenate a String](/hack/expressions-and-operators/string-concatenation) (`.`)
```
Wilfred, line 31: Super pedantic: I'd say "concatenating strings" or "string concatenation". String concatenation requires two strings.

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,46 @@
...
+* [Index Hack Arrays and Strings](/hack/expressions-and-operators/subscript) (`[]`)
```
Wilfred, line 32: How about 'index into hack arrays and strings'? Otherwise this sentence is a little hard to parse, because you have to work out if you're talking about index as a noun or verb.

guides/hack/03-expressions-and-operators/01-introduction.md
```diff
@@ -1,29 +1,46 @@
...
+* [Unpack Tuples](/hack/expressions-and-operators/list) (`list`)
```
Wilfred, line 33: Maybe just 'unpack'? You can use `list()`with both tuples and vecs.

Wilfred (approved):
Left some nitpicks, but this is a really nice improvement.

guides/hack/03-expressions-and-operators/01-introduction.md
lexidor: Agreed, I'd prefer the latter (string concatenation).

<sup>A small technicality, the concatenation operator accepts arraykeys, not just strings. This means "concatenating strings" would be an incomplete description.</sup>

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: You're right. I felt weird including it here and I should have stuck to my instinct.

guides/hack/03-expressions-and-operators/99-XHP-attribute-selection.md
```diff
@@ -0,0 +1,17 @@
+When working with [XHP](/hack/XHP/introduction), use the `->:` operator to retrieve an XHP Class [attribute](/hack/XHP/basic-usage#attributes) value.
```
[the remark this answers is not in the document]
AndrewDiMola: fixed

guides/hack/03-expressions-and-operators/01-introduction.md
AndrewDiMola: My intent was keeping the parallelism with all the other verbs and I found writing that to be a little weird too. Good feedback.

PR #1192, opened 2022-04-03
fix repository of 3.30

cisiqo:
The `ubuntu-3.30-lts` can not found, the correct is `ubuntu-lts-3.30`.


AndrewDiMola:
Thanks! Seems good but don't have access to a VM or linux machine to test. Can someone verify?

fredemmott:
Confirmed by checking with https://dl.hhvm.com/ubuntu/dists/

PR #1194, opened 2022-04-04
Rewrite of Namespaces; Formalization of the "root" namespace

AndrewDiMola:
## Changes
* Added section headers
* Added information on XHP (HTML namespace)
* Rewrites on content on the Namespaces page
* Formalization of "the root namespace", in place of the "global" or "default" namespace

## Before (abridged)
image

## After
image
image
image


guides/hack/02-source-code-fundamentals/37-namespaces.md
```diff
@@ -1,30 +1,39 @@
...
-To refer unambiguously to such names, one can prefix them with a backslash (`\`), as in `\Exception` and `\sqrt`. The names of
-the standard types that are introduced with Hack belong to namespace `HH`.
+Apart from their common prefix, a namespace and its sub-namespaces have no special relationship. For example, `NS1\Sub` can exist without `NS1`. A named top-level namespace does not need to exist for a sub-namespace to exist.
 
+## Reserved Namespaces
 The namespaces `HH`, `PHP`, `php`, and sub-namespaces beginning with those prefixes are reserved for use by Hack.
 
-The pre-defined constant `__NAMESPACE__` contains the name of the current namespace.
+### The `HH` Namespace
+Built-in types like `vec`, `keyset`, `dict`, and `shape` exist in the `HH` namespace (e.g. `HH\Lib\Vec`).
```
lexidor, line 24: The example provided is misconception. The fully qualified name for `vec<T>` is `\HH\vec<T>`. The name vec is automatically imported into all Hack files. The `\HH\Lib\Vec` namespace is related, but not where `vec<T>` is declared.

guides/hack/02-source-code-fundamentals/37-namespaces.md
```diff
@@ -1,30 +1,39 @@
...
 
-When the same namespace is defined in multiple scripts, and those scripts are combined into the same program, the namespace
-is considered the merger of its individual contributions.
+Other types, such as `Exception`, and constants, and library functions (such as `sqrt`) are in `HH` too. Prefix a backslash (`\`) to refer to these types or functions; for example: `\Exception`, `\sqrt`.
```
lexidor, line 26: The type Exception and the sqrt function are not declared in the HH namespace. The fully qualitied name for Exception is `\Exception`, not `\HH\Exception`. These names were inherited from PHP.

guides/hack/02-source-code-fundamentals/37-namespaces.md
```diff
@@ -1,30 +1,39 @@
...
+
+### XHP and the `HTML` Namespace
+As of HHVM 4.73 and XHP-Lib v4, standard XHP elements like `<p>` are defined in `Facebook\XHP\HTML` (for this example, specifically `Facebook\XHP\p`).
```
lexidor, line 29:
```suggestion
As of HHVM 4.73 and XHP-Lib v4, standard XHP elements like `<p>` are defined in `Facebook\XHP\HTML` (for this example, specifically `Facebook\XHP\HTML\p`).
```

guides/hack/02-source-code-fundamentals/37-namespaces.md
```diff
@@ -1,30 +1,39 @@
...
-A *namespace* is a container for a set of (typically related) definitions of classes, interfaces, traits, functions, and constants.
+When the same namespace is defined in multiple scripts, and those scripts are combined into the same program, the namespace
+is the union of all individual components.
 
-A namespace may have *sub-namespaces*, where a sub-namespace name shares a common prefix with another namespace. For example, the
-namespace `Graphics` might have sub-namespaces `Graphics\TwoD` and `Graphics\ThreeD`, for two- and three-dimensional facilities,
-respectively. Apart from their common prefix, a namespace and its sub-namespaces have no special relationship. The namespace
-whose prefix is part of a sub-namespace need not actually exist for the sub-namespace to exist. That is, `NS1\Sub` can exist
-without `NS1`.
+The predefined constant `__NAMESPACE__` contains the name of the current namespace.
```
fredemmott, line 6: Probably should mention:
- `__NAMESPACE__` is primarily for debugging
- `namespace\Foo` (literally lowercase 'namespace') always refers to `Foo` in the current namespace. This can be useful for:
  - clarity, especially e.g. `namespace\map()` is arguably clearer than `map()` in `HH\Lib\Vec` (though it's probabbly better to do a `use namespace HH\Lib\Vec;` and `Vec\map()`, even in that namespace)
  - disambiguating when there is an autoimport

guides/hack/02-source-code-fundamentals/37-namespaces.md
fredemmott, line 26: These are in the root namespcae, not `HH`; if they were in the `HH` namespace, it would be `\HH\Exception`, not `\Exception`.

guides/hack/02-source-code-fundamentals/37-namespaces.md
```diff
@@ -1,30 +1,39 @@
...
+Other types, such as `Exception`, and constants, and library functions (such as `sqrt`) are in `HH` too. Prefix a backslash (`\`) to refer to these types or functions; for example: `\Exception`, `\sqrt`.
+
+### XHP and the `HTML` Namespace
+As of HHVM 4.73 and XHP-Lib v4, standard XHP elements like `<p>` are defined in `Facebook\XHP\HTML` (for this example, specifically `Facebook\XHP\p`).
 
-A namespace directive has two forms.  For example, Script1.php uses the semi-colon form:
+For more information, see [XHP Namespace Syntax](/hack/XHP/basic-usage#namespace-syntax).
+
+## Defining a Namespace
+Namespaces can be defined with semicolons (`;`) or brace delimiters (`{ ... }`).
```
fredemmott, line 34: This feels a bit weird; how about:

namespace declarations can be file-scoped with `namespace MyNS;`, or block-scoped with `namespace MyNS { ... } ` ?

guides/hack/02-source-code-fundamentals/37-namespaces.md
```diff
@@ -52,11 +58,16 @@ namespace NS3\Sub1;
...
 }
 ```
+## Importing from other Namespaces
+With the `use` keyword, a namespace can import one or more member names into a scope, optionally giving them each an alias.
 
-In this case, the namespace extends from the opening brace to the closing brace.
+When importing many names, use `{ ... }`.
+
+```Hack
+use NS1\{C, I, T}; // instead of `NS1\C, NS1\I, NS1\T`
```
fredemmott, line 67: We want to remove raw `use`, and only support `use <kind>`; for example:

```
use namespace HH\Lib\Vec;
use type SomeLib\SomeType;
use const SomeLib\SomeConst;
use function Facebook\Expect\expect;
```

`use Foo;` is essentially `use namespace Foo;` and `use class Foo;` (with some conflict resolution rules), for PHP compatibility which is no longer needed.

guides/hack/02-source-code-fundamentals/37-namespaces.md
```diff
@@ -52,11 +58,16 @@ namespace NS3\Sub1;
...
+```
 
-A namespace can import&mdash;that is, get access to&mdash;one or more names into a scope, optionally giving them each an alias.
-Each of those names may designate a namespace, a sub-namespace, a class, an interface, or a trait.  For example:
+Imported names can designate a namespace, a sub-namespace, a class, an interface, or a trait, as shown in the following example:
```
fredemmott, line 70: - also functions
- any type, not just classes, interfaces, and traits. e.g. type alises to shapes are OK too.

guides/hack/02-source-code-fundamentals/37-namespaces.md
lexidor, line 67: Please prefer `use type` or `use namespace`. This is generally preferred to avoid ambiguous situations. Hhast encourages the `use [kind]` style too. https://github.com/hhvm/hhast/blob/main/src/Linters/UseStatementWIthoutKindLinter.hack

fredemmott (changes requested):

guides/hack/21-XHP/07-basic-usage.md
```diff
@@ -63,12 +63,12 @@ $xhp = <h:p>Hello, world</h:p>;
...
 
 ```
-// from global namespace:
+// exists in the root namespace:
 $xhp = <Facebook:XHP:HTML:p>Hello, world</Facebook:XHP:HTML:p>;
 ```
 
 ```
-namespace CustomNamespace;  // from any namespace:
+namespace CustomNamespace;  // from a defined namespace:
```
lexidor, line 71: The original `any` had meaning here. A name the begins with a namespace separator is fully qualified. These names unambiguously refer to the same thing, whether in the root namespace or CustomNamespace.

guides/hack/02-source-code-fundamentals/37-namespaces.md
lexidor: Thank you, you taught me something today. I didn't know that use clauses could refer to the current namespace.

lexidor:
My comments were unintentionally made at the same time as Fred was writing his review. There is some overlap in our feedback.

guides/hack/02-source-code-fundamentals/37-namespaces.md
AndrewDiMola: ah, I conflated paragraphs when moving. Should've noticed by the lone `\`. Thanks.

AndrewDiMola: Fixed

guides/hack/02-source-code-fundamentals/37-namespaces.md
AndrewDiMola: Apologies, thank you.

guides/hack/02-source-code-fundamentals/37-namespaces.md
AndrewDiMola: ah, can't believe I messed that one up. Thanks.

guides/hack/02-source-code-fundamentals/37-namespaces.md
AndrewDiMola: I learned something today too :-)

guides/hack/02-source-code-fundamentals/37-namespaces.md
AndrewDiMola: Related: Do you think the word 'defined' is wrong in this context?

guides/hack/02-source-code-fundamentals/37-namespaces.md
AndrewDiMola: Good to know. Fixed and will be on the lookout for other examples of ambiguity in the docs.

guides/hack/21-XHP/07-basic-usage.md
AndrewDiMola: This is one of those changes I meant to reverse but accidentally left in the commit. Agreed.

guides/hack/02-source-code-fundamentals/37-namespaces.md
lexidor: The syntax is named `namespace_declaration` in [hhast/NamespaceDeclaration](https://github.com/hhvm/hhast/blob/main/codegen/syntax/NamespaceDeclaration.hack#L15). Definition does indeed sound a bit off, but I am biased by the internal names exposed via hhast.

AndrewDiMola: The reason I changed was to bring more unity: for example. if we say the phrase "define a function", I think we should also use the phrase "define a namespace", or change to "declare a function", etc.

guides/hack/02-source-code-fundamentals/37-namespaces.md
AndrewDiMola: Is this technically correct?

> For clarity, you can use the word "namespace" and form `namespace\Foo` to refer to properties or methods in the current namespace. For example, an alternative way of referring to `Vec\map()`in `HH\Lib\Vec` is `namespace\map()`. This syntax can be useful when you're importing namespaces with many same-named functions and want to refer to the specific function or method in the scope of the current namespace.

lexidor: Namespaces do not have properties or methods. They have functions, type aliases, classishes, and constants. The umbrella term for those things is entity.

The use for `namespace\map()` is as follows. You are in a file that has a namespace header `namespace HH\Lib\Vec`. The usual way to refer to a function from the same namespace would be by calling `map()`. However, many namespaces declare a map() function (Keyset, Dict). You might want to draw attention and say "I mean map from my current namespace". These two (`map()`, `namespace\map()`) are identical if no use clauses are in play. If you wrote `use function HH\Lib\Dict\map;` in the file, now `map()` resolves to the Dict version. You now need to use the `namespace\map()` variant to say "I don't want you to resolve using that use clause, I actually mean `map()` as declared in the `HH\Lib\Vec` namespace."

AndrewDiMola: > Namespaces do not have properties or methods. They have functions, type aliases, classishes, and constants. The umbrella term for those things is entity.
> 
> The use for `namespace\map()` is as follows. You are in a file that has a namespace header `namespace HH\Lib\Vec`. The usual way to refer to a function from the same namespace would be by calling `map()`. However, many namespaces declare a map() function (Keyset, Dict). You might want to draw attention and say "I mean map from my current namespace". These two (`map()`, `namespace\map()`) are identical if no use clauses are in play. If you wrote `use function HH\Lib\Dict\map;` in the file, now `map()` resolves to the Dict version. You now need to use the `namespace\map()` variant to say "I don't want you to resolve using that use clause, I actually mean `map()` as declared in the `HH\Lib\Vec` namespace."

Yes, I should have said constants and functions. Mixing up Class syntax.

guides/hack/02-source-code-fundamentals/37-namespaces.md
lexidor: Indeed, I see `define` being used in [the function introduction](https://docs.hhvm.com/hack/functions/introduction) as well.

&lt;speculation&gt;
This might make a lot of sense to a person who works a lot with C/C++. In those languages a declaration and definition are notably different. You can give multiple declarations for the same function, but only one definition.
```CPP
/* declaration */
std::vector<uint_16> get_elements();
/* definition */
std::vector<uint_16> get_elements() {
  std::vector<int> elements = {1, 2, 3};
  return elements;
}
```
So if my_file.cpp defines the get_elements function, my_file.h would declare it. Other files may include the header file or repeat the declaration.
&lt;/speculation&gt;

In Hack, every definition of a function / classish / type alias / constant is also the only declaration. You can therefore use define and declare interchangeably in Hack.

AndrewDiMola: I think I'm going to keep it as declare in the context of namespaces and define in the context of functions.

guides/hack/02-source-code-fundamentals/37-namespaces.md
AndrewDiMola: @lexidor aside from the issue of redeclaring an existing library function, which will raise an error, it looks like there's an error for trying to do the scenario you described above. That is, if my example is correct.
image

---

Intent is to add a code example to the docs:
```
namespace HH\Lib\Vec;
use function HH\Lib\Dict\map; // map() resolves to `HH\Lib\Dict\map`

// example implementation of HH\Lib\Vec\map
function map<Tv1, Tv2>(
  Traversable<Tv1> $traversable,
  (function(Tv1)[_]: Tv2) $value_func,
)[ctx $value_func]: vec<Tv2> {
  $result = vec[];
  foreach ($traversable as $value) {
    $result[] = $value_func($value);
  }
  return $result;
}

function multiply_vec_by_two(vec<int> $v): vec<int> {
  // resolves to `HH\Lib\Vec\map`, not `HH\Lib\Dict\map`
  return namespace\map($v, (int $value): int ==> { return 2 * $value; },);
}

<<__EntryPoint>>
function main(): void {
  $simple_vec = vec[1, 2, 3];
  $new_vec = multiply_vec_by_two($simple_vec); // a vec of [2, 4, 6]
}
```

lexidor: You can remove the declaration of Vec\map. If you remove that declaration, and hover over the call namespace\map, you'll see you are calling HH\Lib\Vec\map instead of HH\Lib\Dict\map from the use clause.

guides/hack/02-source-code-fundamentals/37-namespaces.md
```diff
@@ -1,30 +1,39 @@
...
-The pre-defined constant `__NAMESPACE__` contains the name of the current namespace.
+## XHP and the `HTML` Namespace
+As of HHVM 4.73 and XHP-Lib v4, standard XHP elements like `<p>` are defined in `Facebook\XHP\HTML` (for this example, specifically `Facebook\XHP\HTML\p`).
+
+For more information, see [XHP Namespace Syntax](/hack/XHP/basic-usage#namespace-syntax).
 
-When the same namespace is defined in multiple scripts, and those scripts are combined into the same program, the namespace
-is considered the merger of its individual contributions.
+## Special Constants
+Use the word "namespace" in the form of `namespace\Foo` to refer to functions, type aliases, classes, and constants in the current namespace.
```
lexidor, line 29: The lowercase `namespace` token is not a constant.

AndrewDiMola: yeah, I guess I'm not entire sure where it should live.

lexidor: This content could live right below the explanation of use clauses. If a use clause makes referring to a name from your own namespace ambiguous...
_You are importing function foo(), but you wish to call the foo() function from your own namespace_
You can prefix the call with `namespace\`.

This is not limited to functions. Type aliases, classishes, constants and sub namespaces can also be ambiguous.

fredemmott (approved):

PR #1195, opened 2022-04-05
Sort Statements pages alphabetically

AndrewDiMola:
I'd like to sort the lists in Statements, Operators, and Built-In Types.
* https://docs.hhvm.com/hack/statements/
* https://docs.hhvm.com/hack/expressions-and-operators
* https://docs.hhvm.com/hack/built-in-types/introduction

Right now the ordering (to me) doesn't make much sense. For these sections, which are just independent references, I think it would be easier to read if they were sorted. The Statements section is shorter than the other sections so the effect isn't as noticeable here as it would be for the other sections.
* Update: See https://github.com/hhvm/user-documentation/pull/1195#issuecomment-1089240106.

URLs should remain the same. If this goes well and we think that it's useful, I'll do the same for the other sections. Just changed the numbers on the file names. Not sure if this will cause unintended effects elsewhere. On Introduction, changing format of links according to best practices.

## Before (check sidebar)
image

## After (check sidebar)
image


lexidor:
I can't speak for the author, but I believe the order / grouping from before may have put some thought into it.

Ordering this alphabetically, although nicer when navigating the sidebar, affects browsing using the next page button too.

Here is an if -> here is a switch (a powerful if) -> here is a while -> here is an alternate while (do-while) -> here is a for (a while with an initial statements and a post statements) -> here is foreach (a more limited, but safer for) -> we have now discussed all things that continue and break can affect, so here are continue and break -> and here are some misc statements.

AndrewDiMola:
> I can't speak for the author, but I believe the order / grouping from before may have put some thought into it.
> 
> Ordering this alphabetically, although nicer when navigating the sidebar, affects browsing using the next page button too.
> 
> Here is an if -> here is a switch (a powerful if) -> here is a while -> here is an alternate while (do-while) -> here is a for (a while with an initial statements and a post statements) -> here is foreach (a more limited, but safer for) -> we have now discussed all things that continue and break can affect, so here are continue and break -> and here are some misc statements.

Ah, perhaps it wasn't fair to say there was no intentionality. It might be worth checking on our back-end analytics to see how many sessions follow the pages in order. My assumption is that Hack isn't anybody's first language which would imply that all of these statements that are not Hack-specific are already familiar and are just used as reference when they need them.

lexidor:
> It might be worth checking on our back-end analytics to see how many sessions follow the pages in order.

👍 That would be useful to know. I may very well be an outlier. I browse quite frequently.

AndrewDiMola:
For the other sections, I would love to be able to add a further section within Built-in Types, at least in the sidebar, to arrange the content in a manner like:

Built-In Types
- Introduction
- Primitive Types 
-- int
-- bool
- Hack Arrays
-- vec
-- keyset
-- dict

Even knowing the dangers of nesting too deep, I wonder if it's possible to create a third layer of abstraction there.

image

AndrewDiMola:
> > It might be worth checking on our back-end analytics to see how many sessions follow the pages in order.
> 
> 👍 That would be useful to know. I may very well be an outlier. I browse quite frequently.

Another outlier here, checking in...

lexidor:
https://github.com/hhvm/user-documentation/pull/1195#issuecomment-1089259975

Definitely a plus one from me on that suggestion. Would you put arraykey and num under the primitive types section?

AndrewDiMola:
> [#1195 (comment)](https://github.com/hhvm/user-documentation/pull/1195#issuecomment-1089259975)
> 
> Definitely a plus one from me on that suggestion. Would you put arraykey and num under the primitive types section?

Not entirely sure. Was just thinking about those two. Was thinking about checking the older O'Reilly book to see how it indexed all this content.

AndrewDiMola:
Looking at the past year (4/9/21 to 4/9/22):
* the `foreach` page had the most hits (~12k); followed by the Introduction page (~5k); 
* followed by the `switch` page (~5k); 
* followed by `for` and `try` (~4k); 
* followed by `if` (~2k), which is also the first in the index if going page-by-page.

Here's what I see for this date on next entrances from the Introduction page:

image

Followed by this data on the `if` page:

image

And, finally, if we stop at the third path (Introduction -> `if` -> `switch`), we see this data:

image

I think there is clear use of going page-by-page. The question is whether it's significant in the totality of views and usage.

What do you think @lexidor and @Wilfred?

Wilfred (approved):
I'm in favour of this. When you're using the sidebar, you're often trying to find something specific, and alphabetical ordering makes that easier.

I agree that navigating by theme is important too, but we can do that by linking between pages. This solves problems like `break` occurring in both loops and `switch`, but `continue` only occurring in loops.

PR #1205, opened 2022-04-22
Rewrite of HSL main page

AndrewDiMola:
I'm working on detailing out all the HSL Namespaces.

- Mostly new content. Cut out some (IMO) extraneous information. In general, trying to remove philosophical details in favor of actual details.
- Long-term goals are to create index pages at places like `/hsl/reference/function/HH.Lib.C`. Have tried implementing this a few times. Not quite there yet.
- Short-term goals are to just write about what we have, since the current page was written a while back, back when the HSL was still packaged separately.

image
image
image

(Ideally the names in the first column would link to index pages, but, again, not quite there yet.)

**Aside:** the 1.5 em spacing has been driving me a bit mad, also in places with bulleted lists. Thinking about cutting that bottom padding / 2.


AndrewDiMola:
For the namespace index, I was thinking of doing something like creating a keyset of tokens split by `\` so you could get a list of the unique namespace names (like `Vec\), and then appending here: https://github.com/hhvm/user-documentation/blob/main/src/site/xhp/api-list.php#L73

I've approached this from a bunch of different angles in the last few months, but can't think of a good way.

PR #1206, opened 2022-04-25
[WIP] Change fbonly to metaonly for internal notes

AndrewDiMola:
- Change `fbonly` and `fbOnly` to `metaonly` and `metaOnly`
- Add new note on Hack Strict for www
- Update Introduction in Source Code Fundamentals


guides/hack/02-source-code-fundamentals/01-introduction.md
```diff
@@ -1,11 +1,12 @@
...
-* [comments](comments.md)
-* [constants](constants.md)
-* [keywords](keywords.md)
-* [literals](literals.md)
-* [names, user-defined](names.md)
-* [namespaces](namespaces.md)
-* [scope](scope.md)
-* [script inclusion](script-inclusion.md)
-* [variables](variables.md)
+* [program structure](/hack/source-code-fundamentals/program-structure)
```
alexeyt, line 3: Does adding this really belong on this PR?

guides/hack/02-source-code-fundamentals/04-program-structure.md
```diff
@@ -32,6 +32,13 @@ have been `run`, `do_it`, or `make_magic`. (What makes `main` the start-up funct
 A file can import another file via [inclusion](script-inclusion.md).
 
 ## Legacy Files
+```yamlmeta
+{
+  "metaonly messages": [
+    "All Hack source code at Meta should use the .php file extension and Hack strict mode (plain <?hh header line)."
+  ]
+}
+```
```
alexeyt, line 41: Does adding this really belong on this PR?

sass/16-extending.md
```diff
@@ -0,0 +1,482 @@
+XHP comes with classes for all standard HTML tags, but since these are first-class objects, you can create your own XHP classes for rendering
```
alexeyt, line 1: Does adding this file really belong on the rename PR?

alexeyt:
The changes mostly make sense to me, but it would be nice if this PR was only for renaming and the new additions were in a separate PR.

sass/16-extending.md
AndrewDiMola: huh -- bizarre. Not sure why this was a new addition.

guides/hack/02-source-code-fundamentals/04-program-structure.md
AndrewDiMola: Someone internally reached out to me to write a note and I realized that we should rebrand as well. So I kind of just started adding and it became something else. You're right in that it should be separated.

AndrewDiMola:
I'll close this out and separate into new PRs. Still WIP on the rebranding though. I think the `fbonly messages` tag is being added something to the reference documentation, and I can't figure out where that is, whether it's internal, or in a separate project, or whether I have to do some other sort of rebuild first.

guides/hack/02-source-code-fundamentals/01-introduction.md
AndrewDiMola: Well, the program structure page, which was referenced by the original note I was adding wasn't a part of the actual index. And I figured while I was here I would change the format into a better one -- a format that the test suite can read and cause errors on if a page is referenced that was renamed or deleted, etc. So they are related to my original intention -- adding the fb-only (meta-only) note, but then I got too involved with refactoring the code to the meta branding (also another request).

AndrewDiMola:
Closing in favor of #1208 and #1209 .

PR #1209, opened 2022-04-27
Note on hack strict for Meta engineers

AndrewDiMola:
A note on hack strict, from internal feedback.

`Facebook` -> `Meta` refactor at a later date.

image

### DIFF
--- a/guides/hack/02-source-code-fundamentals/04-program-structure.md
+++ b/guides/hack/02-source-code-fundamentals/04-program-structure.md
@@ -32,6 +32,13 @@ have been `run`, `do_it`, or `make_magic`. (What makes `main` the start-up funct
 A file can import another file via [inclusion](script-inclusion.md).
 
 ## Legacy Files
+```yamlmeta
+{
+  "fbonly messages": [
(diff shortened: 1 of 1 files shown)


Wilfred (approved):
Looks good, but all files in Hack are strict now. `<?hh // strict` and `<?hh // partial` have no effect. The only way you can have a partial mode file is using the `.hackpartial` extension.

AndrewDiMola:
> Looks good, but all files in Hack are strict now. `<?hh // strict` and `<?hh // partial` have no effect over `<?hh`. The only way you can have a partial mode file is using the `.hackpartial` extension.

Do you think it would be better to say... "All Hack source code at Meta should use the .php file extension and, by default, Hack Strict mode (with the plain <?hh header line)."

And also just add a general note on the existence of `.hackpartial`?

Atry:
`Facebook Engineer?` vs `Meta Platforms Engineer?`

AndrewDiMola:
> Facebook Engineer? -> Meta Engineer?

@Atry - I'm going to refactor this in a separate PR as there are a lot of file changes.

PR #1210, opened 2022-04-28
Add Tip, Warning, and Danger message macros

AndrewDiMola:
Adds a Tip, Warning, and Danger macro message, through a YAML code embed, which involves refactoring the `apiMessageTop` class.

Let me know what you think!

image

Renders a bit different than the live version, but I think it's better spaced here:
image


sass/basics.scss
```diff
@@ -37,6 +37,33 @@ a {
...
+  font-size: smaller;
+  padding: 0.5em;
+  margin-bottom: 0.5em;
+}
+
+.dangerMessage {
+  border: 1px solid #e69a92;
+  background-color: #fff8f7;
+  color: #526d8a;
+  font-size: smaller;
```
Wilfred, line 62: I think I'd keep the enclosing font size rather than using a smaller one.

sass/basics.scss
```diff
@@ -37,6 +37,33 @@ a {
   color: red;
 }
 
+.tipMessage {
+  border: 1px solid #c3e0c9;
+  background-color: #f3f9f4;
+  color: #526d8a;
+  font-size: smaller;
+  padding: 0.5em;
+  margin-bottom: 0.5em;
```
Wilfred, line 46: Nit: You might want to consider factoring out a `.message` class, along these lines:

```
.message {
  padding: 0.5em;

  .tip {
    background-color: #f3f9f4
  }
}
```

Wilfred (approved):

sass/basics.scss
```diff
@@ -37,6 +37,33 @@ a {
...
+}
+
+.warningMessage {
+  border: 1px solid #feedbb;
+  background-color: #fffdf7;
+  color: #526d8a;
```
Wilfred, line 52: Do we use this text colour elsewhere? If not, it might look more coherent if we use the existing text paragraph colour.

sass/basics.scss
AndrewDiMola: I feel that is looks a bit weird in the same size.

Here's how it looks with `small`, removed, and `medium`

image

AndrewDiMola:
Will merge in on Monday, unless there's other feedback.

PR #1223, opened 2022-06-03
Increate MinSize of the autoscaling group to 2

### LINKED ISSUE #1222: Cannot deploy the document site
According to [the Github Action log](https://github.com/hhvm/user-documentation/runs/6699209905?check_suite_focus=true), we encountered the following error when deploying a new version of this document site:

```
Instance deployment: The Docker container unexpectedly ended after it was started.
```

I investigated the events on the AWS web console, and it seems what get killed is not only the newly deployed site, but also the previously deployed server (in an existing Docker container), then the autoscaling group terminated the EC2 instance and restarted another EC2 instance, and finally deployed the new site successfully. However, the document site was temporarily down during the redeployment and Terraform considered it as a failure and terminated the Github Action flow.

I guess the reason why these containers get killed is out-of-memory when AWS temporarily runs two Docker containers on a `t3.micro` EC2 instance. Now we have switched to `t3.small` in c7137c27e1a1d3911bb830077692bc2867198b1a, which has more memory, and should have mitigated the issue.

## Background

Currently we had configured an autoscaling group running Docker containers managed by AWS Elastic Beanstalk. However, the autoscaling group always contains a single EC2 instance because it does not support scaling up based on number of running Docker containers.

## Future work

It seems ECS on Fargate would be better than Elastic Beanstalk for our use case, because Fargate is a public cluster shared with other AWS users, which should always have available memory to deploy a new container.

The cons is that Fargate is more expensive than Elastic Beanstalk for low-traffic applications according to [the comparison](https://blog.boltops.com/2018/04/22/heroku-vs-ecs-fargate-vs-ec2-on-demand-vs-ec2-spot-pricing-comparison/#comparison-list), but it might be worth if it could save our time fighting with the private autoscaling groups.

We can try to migrate from Elastic Beanstalk to ECS on Fargate if the failures appear continuously even when we are using `t3.small` instances.

Another option is to use EKS to declare memory and CPU requirement and set up an autoscaler according to [this document](https://docs.aws.amazon.com/eks/latest/userguide/autoscaling.html), but it is over-complicated given that what we need is just to allocate a new EC2 instance during deployment.

Atry:
This PR should fix #1222 by increate the minimal size of the cluster to 2, so that there will be always an EC2 instance standby for deployment. It looks like a stupid fix but the cost of two `t3.micro` instances. is close to one `t3.small` instance.

### DIFF
--- a/.deploy/terraform/main.tf
+++ b/.deploy/terraform/main.tf
@@ -189,7 +189,7 @@ resource "aws_elastic_beanstalk_configuration_template" "docs" {
   setting {
     namespace = "aws:autoscaling:launchconfiguration"
     name = "InstanceType"
-    value = "t3.small"
+    value = "t3.micro"
   }
   setting {
     namespace = "aws:autoscaling:launchconfiguration"
@@ -235,7 +235,7 @@ resource "aws_elastic_beanstalk_configuration_template" "docs" {
   setting {
     namespace = "aws:autoscaling:asg"
     name = "MinSize"
-    value = "1"
+    value = "2"
   }
   setting {
     namespace = "aws:autoscaling:asg"


Wilfred (approved):
Looks good to me.

PR #1226, opened 2022-06-07
Revert "Revert "Comment out the IPAddressType setting 

Atry:
`IPAddressType` is an unsupported setting and it will result in failures in the newer version of Terraform AWS Provider (previously it was ignored)

This reverts commit e8efe59e9f28db7e4a944e5e3806d7c4a5f2687e.

This PR is a mitigation of #1222


alexeyt (approved):
Seems reasonable.

PR #1227, opened 2022-06-07
Randomize the Elastic Beanstalk environment names

Atry:
According to [the log](https://github.com/hhvm/user-documentation/runs/6734843871?check_suite_focus=true), Elastic Beanstalk environment cannot be updated under our current settings. This PR forces it to be recreated instead of being updated. Therefore this PR will result in some downtime when deploying a new version.

Note that a sophisticated solution to the immutable problem would be to configure a rolling policy for the Auto Scaling Group, but it would be too complicated and need further debugging. As a mitigation, this PR is simpler.


Atry:
I am merging this PR because anyway we should not include unsupported settings

PR #1228, opened 2022-06-07
Revert "Revert "Randomize the Elastic Beanstalk environment names""

Atry:
#1227 was merged unexpectedly and has been reverted. This is a recreation of #1227

This reverts commit 2b04410663e12f1512fcc3fb1469218c0a5a2468.


Atry:
As discussed in WorkChat, we always have the down time because of OOM. So this PR at least does not make things worse.

PR #1229, opened 2022-06-13
Deploy this docs site to isolated environments according to branch name

Atry:
What would this PR do:
1. revert all the mitigation commits
2. Choose Terraform workspace according to branch name, so that it would not interfere existing AWS resources
3. Set the fallback value for PROD_TAG, when there is no existing EB environment


Atry:
I don't understand the test failure. This PR does not change source code or the test job in GitHub Actions.

Atry:
I think the test failure is irrelevant to this PR because the test failure also occurs when I rerun nightly CI for the main branch:

https://github.com/hhvm/user-documentation/runs/6870181081?check_suite_focus=true

Atry:
The test failure seems flaky because it passed when I reran the CI https://github.com/hhvm/user-documentation/runs/6870193613?check_suite_focus=true

PR #1231, opened 2022-06-15
Update Built-In Types Landing Page

AndrewDiMola:
* Adding `null` as a primitive type
* Clarifying that `num` and `arraykey` are union types.
* Reworking description of `mixed`

image

### DIFF
--- a/guides/hack/11-built-in-types/01-introduction.md
+++ b/guides/hack/11-built-in-types/01-introduction.md
@@ -7,9 +7,10 @@ Hack has the following primitive types:
 [float](/hack/built-in-types/float), and 
 [string](/hack/built-in-types/string).
 
-* Some types act as parent, container types, like [num (`int`/`float`)](/hack/built-in-types/num) and [arraykey (`int`/`string`)](/hack/built-in-types/arraykey).
-* Others like [mixed](/hack/built-in-types/mixed) act as supertypes, or, like [nothing](/hack/built-in-types/nothing), act as the complete opposite (no type).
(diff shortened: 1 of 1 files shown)


lexidor:
Strange to find `nonnull` and `mixed` under the `Primitive types` heading. Not something you added, just something I noticed.

I agree that container types was not the right name for the set `arraykey`, `num`. It is easily confused with `Container<_>`.

`int` is a subtype of `arraykey`, true. These types have been explained to me as intersection types. In the same way that `?int` is the intersection `null|int` is `num` `int|float`. Would it be an idea to refer to these types as intersections?

Is `null` not a primitive type? In Javascript it is. https://developer.mozilla.org/en-US/docs/Glossary/Primitive

AndrewDiMola:
> Strange to find nonnull and mixed under the Primitive types heading. Not something you added, just something I noticed.

Yeah, you're right. It is weird. There should be a section separator. 

> I agree that container types was not the right name for the set arraykey, num. It is easily confused with Container<_>.

That was the exact issue 😄 

> int is a subtype of arraykey, true. These types have been explained to me as intersection types. In the same way that ?int is the intersection null|int is num int|float. Would it be an idea to refer to these types as intersections?

I think by implementation these would instead be called union types. Will update.

> Is null not a primitive type? In Javascript it is. https://developer.mozilla.org/en-US/docs/Glossary/Primitive

Good point. Asked some team members internally and they agreed. Updating.

lexidor (approved):
💯 Looks good to me

PR #1243, opened 2022-06-18
Fix certificate error when accessing https://staging.docs.hhvm.com

Atry:


alexeyt (approved):
This seems reasonable, especially since we have test coverage of this code.

PR #1252, opened 2022-07-05
Add version number to the cn-terraform/networking/aws module

Atry:


Atry:
The server to test this PR branch is deployed at docs-Atry-patch-2-lb-97044275.us-west-2.elb.amazonaws.com, according to the log output https://github.com/hhvm/user-documentation/runs/7202840631?check_suite_focus=true#step:8:1312

PR #1257, opened 2022-07-05
Provide context on operators and fix incorrect precedence listings

### LINKED ISSUE #1212: Operator precedence

# Where is the problem?

In the operator precedence table, `&&` beats `|`.

# What is the problem?

Would you happen to know if the operator precedence table is correct? I believe it not to be.

 * Given the expression `0 | 0 && 0` 
 * The precedence rules from the docs state: `&&` has a higher precedence than `|`
 * So you would expect the expression `0 | (0 && 0)` to be the same
 * Evaluating `0 | 0 && 0` results in `false`
 * Evaluating `0 | (0 && 0)` results in `InvalidOperationException`

This leads me to believe that `0 | 0 && 0` is parsed as `(0 | 0) && 0`, and not `0 | (0 && 0)`.

 - Build ID: HHVM=HHVM-4.154.0:HSL=v4.108.1:2022-05-03T20:54:00+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/expressions-and-operators/operator-precedence
 - Page requested at: Sat, 07 May 2022 17:52:43 +0000

AndrewDiMola:
## Changes
* Fixes #1212 - `&&` should be lower, was likely confused with `&`.
* Add reference links to explain and give context for the different Hack operators.

## Open Questions / Discussion
* Is `print` effectively [`echo`](https://docs.hhvm.com/hack/expressions-and-operators/echo)? 
* Is there a more appropriate link for `include` and `require`?
* Any keyword or operator that should not be publicized?
* Trying to avoid using "member" across the documentation set. @Wilfred, let me know if there's a better way to phrase `->` and `?->`.

## Before
image

## After
image
image

### DIFF
--- a/guides/hack/03-expressions-and-operators/07-operator-precedence.md
+++ b/guides/hack/03-expressions-and-operators/07-operator-precedence.md
@@ -4,31 +4,33 @@ Operators higher in the table have a higher precedence (binding more
 tightly). Binary operators on the same row are evaluated according to their
 associativity.
 
-Operator | Associativity
------------- | ---------
-`\` | Left
-`[]` | Left
-`->` `?->` | Left
-`new`  | None
-`()` | Left
-`clone` | None
-`await` `++` `--` (postfix) | Right
-`(int)` `(float)` `(string)` `**` `@` `++` `--` (prefix) | Right
-`is` `as` `?as` | Left
-`!` `~` `+` `-` (one argument) | Right
-`*` `/` `%` | Left
-`.` `+` `-` (two arguments) | Left
-`<<` `>>` | Left
-`<` `<=` `>` `>=` `<=>` | None
-`===` `!==` `==` `!=` | None
-`&&` | Left
-`^` | Left
-`\|\|` | Left
-`&` | Left
-`\|` | Left
-`??` | Right
-`?` `:` `?:` | Left
-`\|>` | Left
-`=` `+=` `-=` `.=` `*=` `/=` `%=` `<<=` `>>=` `&=` `^=` `\|=` `??=` | Right
-`print` | Right
-`include` `require` | Left
+Operator | Description | Associativity
+------------ | --------- | ---------
+`\` | [Namespace separator](/hack/source-code-fundamentals/namespaces)  | Left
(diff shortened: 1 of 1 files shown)


AndrewDiMola:
@lexidor - I'm sorry for not including your awesome script that automatically generates the table. I'm opting to instead keep the page as manually created so that reference links can be included for additional context.

Let me know what you think!

lexidor:
> @lexidor - I'm sorry for not including your awesome script that automatically generates the table. I'm opting to instead keep the page as manually created so that reference links can be included for additional context.
> 
> Let me know what you think!

The script was made to provide insight. It presents the data to doc authors and its output is not suitable for inclusion in the docs verbatim anyways.

> Is print effectively [echo](https://docs.hhvm.com/hack/expressions-and-operators/echo)?

Afaik, yes :)

> Any keyword or operator that should not be publicized?

The dollar operator is invalid in Hack. `$x` is a variable. `$$x` is the dollar operator applied to a variable. It deferences the string contents of $x, yuck.
```HACK
$a = 5;
$x = 'a';
// FIXME that Hack errors, yuck.
echo $$x; // prints 5
```

AndrewDiMola:
> The script was made to provide insight. It presents the data to doc authors and its output is not suitable for inclusion in the docs verbatim anyways.

I had been thinking about using its output as a block in the .md file as a way to regression test against the page. I still think it's viable in some way... as a reference, it remains very useful.

> The dollar operator is invalid in Hack. $x is a variable. $$x is the dollar operator applied to a variable. It deferences the string contents of $x, yuck.

Ah, I see. It seems to have been deprecated here:
https://hhvm.com/blog/2019/10/01/deprecating-references.html

Removing. Also linking `echo` and `print`.

lexidor:
> Ah, I see. It seems to have been deprecated here:
https://hhvm.com/blog/2019/10/01/deprecating-references.html

This isn't the removal of the dollar operator. This is the removal of the reference operator. `$$x` =/= `&$x`.

lexidor (approved):

PR #1262, opened 2022-07-13
Rename fb⇝Meta in fbOnly block, try to fix #1129

### LINKED ISSUE #1129: Remove leading 'fbOnly' sections from the DOM if non-FB viewer
This is most important for API reference pages.

Right now, we hide them with CSS. The problem with this is it shows up in Google results (e.g. search for 'vec map hhvm', and 'Facebook Engineer? This is available as Vec\map in the www repository. ' is just a waste of space there and makes google less useful.

This is not about confidentially etc, as the site (and all of its' inputs) are open source and made on public systems. Just about making the text that google shows more useful.

This isn't quite trivial as we currently pregenerate the entire `.html` file. There's a few options:
- filter them out of the HTML at render time (brittle, slow, but straightforward)
- stop putting them in the HTML. Inject at render time (only works for leading fbonly messages, unless very complicated)
- render separate _fb.html files. Show either foo.html or foo_fb.html depending on ASN

For all of these, it would probably be worth adding a markdown extension to explicitly recognize fbonly, instead of using `<div class='fbOnly'>`, so they can be more-reliably matched.

walaura:
This seems like it should work. Fixes #1129 (maybe)

Google's docs on `data-nosnippet`:
https://developers.google.com/search/docs/advanced/robots/robots_meta_tag

Happy to split the commits! i was going for a stacked diff sorta thing here but it's been a while since i used GH ><

### DIFF
--- a/src/markdown-extensions/YamlFrontMatterBlock.php
+++ b/src/markdown-extensions/YamlFrontMatterBlock.php
@@ -128,12 +128,12 @@ abstract class YamlFrontMatterBlock implements UnparsedBlocks\BlockProducer {
       $messages,
       $message ==> new UnparsedBlocks\InlineSequenceBlock(
         '<div class="message api fbOnly">'.
-        "**Facebook Engineer?**\n\n".
+        "**Meta Engineer?**\n\n".
         '<p>'.
         $message.
         "</p>\n".
         '<!-- '.
-        'Not a Facebook engineer... yet? https://www.facebook.com/careers/'.
+        'Not a Meta engineer... yet? https://www.facebook.com/careers/'.
         " -->\n".
         '</div>',
       ),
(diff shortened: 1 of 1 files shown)


src/markdown-extensions/YamlFrontMatterBlock.php
```diff
@@ -127,13 +127,13 @@ private static function getFacebookMessages(
...
-        '<div class="message api fbOnly">'.
-        "**Facebook Engineer?**\n\n".
+        '<div class="message api fbOnly" data-nosnippet>'.
+        "**Meta Engineer?**\n\n".
         '<p>'.
         $message.
         "</p>\n".
         '<!-- '.
-        'Not a Facebook engineer... yet? https://www.facebook.com/careers/'.
+        'Not a Meta engineer... yet? https://www.facebook.com/careers/'.
```
demosdemon, line 136: https://metacareers.com/

walaura: done!

walaura:
hey thanks! hopefully we get a clean merge because i got 0 tooling for this so i used codespaces! best of luck with the fix, hope it works 🥹

AndrewDiMola:
Merged in https://github.com/hhvm/user-documentation/pull/1263

PR #1263, opened 2022-07-13
Add data-nosnippet HTML attribute

AndrewDiMola:
> _This is not about confidentially etc, as the site (and all of its' inputs) are open source and made on public systems, and the FB-specific content will still be in the public docker images. This is just about making the text that google shows more useful._

Inspired by @walaura's great suggestion, add the [`data-nosnippet` HTML attribute](https://developers.google.com/search/docs/advanced/crawling/special-tags#inline-directives), which tells Google's search results to not include targeted sections (with the attribute applied) in their search results.

This attribute is added to internal-focused content that generates in three major areas:
* inline tags,
* information blocks generated in reference pages, and
* YAML-generated macros.

Screen Shot 2022-07-13 at 3 56 56 PM

Improves issues surfaced in https://github.com/hhvm/user-documentation/issues/1129.


Atry (approved):

Atry:
That's a smart attribute. Hope it works!

alexeyt (approved):

PR #1264, opened 2022-07-14
Add instructions on upgrading via codespaces

AndrewDiMola:
There's a few extra steps to ensure that the version of HHVM used on your codespace matches the version that is being targeted for upgrade. The README details those steps.

PR #1265, opened 2022-07-14
Upgrade to 4.164

AndrewDiMola:


AndrewDiMola:
I wrote these instructions on how to do an upgrade via codespaces. Figured I should test it out to make sure that I'm doing everything (and have documented everything) correctly.

https://github.com/hhvm/user-documentation/pull/1264

AndrewDiMola:
@Atry any idea why this is failing?

Atry:
The load balancer name is too long. I think we should let it generate a randomized name instead of just concatenating all the parts, as mentioned in https://github.com/cn-terraform/terraform-aws-ecs-alb/issues/31

As a workaround, just use a shorter branch name.

PR #1266, opened 2022-07-18
Add instructions on upgrading via codespaces

AndrewDiMola:
There's a few extra steps to ensure that the version of HHVM used on your codespace matches the version that is being targeted for upgrade.

### DIFF
--- a/updating.md
+++ b/updating.md
@@ -25,7 +25,7 @@ See `bin/update-versions --help` for more options.
 # Updating to a new version of HHVM and the HSL from a Codespace
 
 First, ensure that your Codespace has a large enough machine type to perform 
-these upgrade operations. At minimum, an 8-core, 64GB machine is usually necessary.
+these upgrade operations. At minimum, an 16-core, 32GB machine is usually necessary.
 
 As usual, run `update-versions` to update the targetted HHVM version. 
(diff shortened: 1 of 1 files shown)


AndrewDiMola:
Renamed branch so that AWS doesn't complain (hopefully).

updating.md
```diff
@@ -21,3 +21,50 @@ $ bin/update-versions --hhvm-tag=HHVM-a.b.c --hsl-tag=vx.y.z
 ```
 
 See `bin/update-versions --help` for more options.
+
+# Updating to a new version of HHVM and the HSL from a Codespace
+
+First, ensure that your Codespace has a large enough machine type to perform 
+these upgrade operations. At minimum, a 16-core, 32GB machine is usually necessary.
+
+As usual, run `update-versions` to update the targetted HHVM version.
```
Wilfred, line 30: "as usual" kinda implies to me that the user ought to know this already, which seems a little discouraging.

Wilfred (approved):

AndrewDiMola:
@Wilfred - You're right. Usually I'm more sensitive to language like this. My intention was that a reader would be familiar with the existing steps (the first section), so I was intending to let the reader know that the following steps would be familiar steps. Will remove. Both audiences are probably mutually exclusive anyway.

PR #1275, opened 2022-07-22
`$_` is not reserved for `list()` and `foreach`

fredemmott:
It's frequently used for them, but is not reserved; it can be used for any mandatory-but-unused assignment.

### DIFF
--- a/guides/hack/02-source-code-fundamentals/13-names.md
+++ b/guides/hack/02-source-code-fundamentals/13-names.md
@@ -23,9 +23,12 @@ function compute(int $val): void {
 }
 ```
 
-## The Placeholder Variable
-The name `$_`, referred to as the *placeholder variable*, is reserved for use in the
-[list intrinsic function](../expressions-and-operators/list.md) and the [foreach statement](../statements/foreach.md).
+## Placeholder Variables
+Variables beginning with `$_` are referred to as *placeholder variables*, and should be used where an assignment is structurally required but the value is unused. This is frequently used for the
+[list intrinsic function](../expressions-and-operators/list.md) and the [foreach statement](../statements/foreach.md) - for example:
+
+* `list($foo, $_, $bar) = baz();` if `baz()` returns a 3-element tuple but the middle element is unused
+* `foreach ($container as $key => $_) { /* ... */ }` when iterating a keyed container and the values are unneeded
 
 ## The Current Instance Variable
 The name `$this` is predefined inside any instance method or constructor when that method is called from within an object context.


fredemmott:
👋 Slack's point of contact has emailed cla@ to add me to their corporate CLA

AndrewDiMola:
@fredemmott - Was thinking about adding actual examples to make it more useful.

image

Do these examples convey what you're trying to teach?
```placeholder-list.hack
$tuple = tuple('H', 'a', 'c', 'k', 'l', 'a', 'n', 'g');

// create names for $a, $b, $c, and $d, but ignore everything else
list($a, $b, $c, $d, $_, $_, $_, $_) = $tuple;
echo("I love ".$a.$b.$c.$d); // "I love Hack"
```
```placeholder-foreach.hack
$keyed_container = dict['H' => 8, 'a' => 1, 'c' => 3, 'k' => 11];

// use the placeholder variable for keys instead of values
foreach ($keyed_container as $key => $_) {
  echo ($key); // prints keys: 'Hack'
}

echo(" ... and associated values: ");

// a regular example for contrast
foreach ($keyed_container as $key) {
  echo ($key); // prints values: 81311
}
```

fredemmott:
The for-each example does; the tuple example technically does, however:
- it's showing using a tuple as a vec, which is a bit of a smell; while it's tangental, it'd be better not to include it
- I feel that skipping an element in the middle of a tuple is a clearer example

fredemmott:
> `foreach ($keyed_container as $key) {`

Also that should be `$value`, not `$key`

edwinsmith:
Hi Fred!

fredemmott:
@AndrewDiMola would you like me to change these to full examples, or is that a follow-up you'd like to do?

> what you're trying to teach

The main thing I want to do with this PR is to remove "**is reserved for use in** the
 list intrinsic function and the foreach statement"; just added some examples in passing while I was here to hopefully make 'why does this exist?' clearer :)

AndrewDiMola:
> @AndrewDiMola would you like me to change these to full examples, or is that a follow-up you'd like to do?
> 
> > what you're trying to teach
> 
> The main thing I want to do with this PR is to remove "**is reserved for use in** the list intrinsic function and the foreach statement"; just added some examples in passing while I was here to hopefully make 'why does this exist?' clearer :)

We can merge and then I'll add improvements in a separate PR (maybe just for the `foreach`). Sorry for the delay! In Hack on Hack this week!

AndrewDiMola (approved):

PR #1280, opened 2022-08-09
Async guidelines: discourage `HH\Asio\join()`

fredemmott:
This currently appears to encourage `HH\Asio\join()`, and doesn't mention the downsides

### DIFF
--- a/guides/hack/15-asynchronous-operations/28-guidelines.md
+++ b/guides/hack/15-asynchronous-operations/28-guidelines.md
@@ -493,10 +493,11 @@ function main(): void {
 }
 ```
 
-## Use `join` in Non-async Functions
+## Integrating async and non-async functions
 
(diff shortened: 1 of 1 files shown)


guides/hack/15-asynchronous-operations/28-guidelines.md
```diff
@@ -514,7 +515,9 @@ function main(): void {
 }
 ```
 
-This scenario normally occurs in the global scope, but can occur anywhere.
+**THIS IS A COUNTEREXAMPLE**: in real-world code, the entrypoint should be made async instead.
```
lexidor, line 518: You removed one reference to "the global scope" (hhvm 4.6x pseudo mains), but this one survived.
`// In a non-async function, or the global scope, you can`

Atry:
Looks good to me

Issue #1216, opened 2022-05-18
Mention postfix and prefix on increment / decrement guide


AndrewDiMola:
Please complete the information below:

# Where is the problem?

Incrementing / Decrementing guide.

# What is the problem?

We never talk of the consequences of a post- or pre-increment/decrement. 

If Hack is someone's first language (maybe one day), they should also learn what this means conceptually.

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: HHVM=HHVM-4.154.0:HSL=v4.108.1:2022-05-16T17:51:53+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/expressions-and-operators/incrementing-and-decrementing
 - Page requested at: Wed, 18 May 2022 21:13:52 +0000
 - Controller: GuidePageController

thesourabh:
Looking into this!

fredemmott:
As I understand it, there is no significant difference in Hack: in other languages, it matters when it is used in an expression, but `++$x` and `$x++` are both **statements** in Hack, not expressions, so the difference can't come up in valid hack code:

```
$x = 123;
var_dump($x++); // Parsing[1002] Assignments can no longer be used as expressions.
```

EvanHelstrom:
@fredemmott 

I ran some test code and I think it makes a difference with this specific example, but I can't think of any other reason why since you can't use them for assignments to other variables or such.

fredemmott:
@EvanHelstrom the runtime still understands it (it wasn't always banned), but as the typechecker bans it, it's not considered valid Hack code - especially as it's 1002 'parse error', which *no* projects should permit HH_FIXMEs for.

EvanHelstrom:
I guess we can just close this issue then since there isn't really a need for it to be adjusted. Or should it be explained with no use case?

AndrewDiMola:
If there's no real consequence in Hack and you think readers will know ++ and -- by proxy of other languages, I'm fine with you closing this issue.

Wilfred:
Per discussion above, there's no significant difference as this syntax is only allowed in statements.

PR #1286, opened 2022-09-07
Mention postfix/prefix on increment guide

EvanHelstrom:
As a part of #1216: Mention postfix and prefix on increment/decrement guide

Added a short text explaining the difference between a prefix and postfix use of the increment and decrement operators. Located at the bottom of the page, it uses two examples and gives a brief explanation.


fredemmott (changes requested):
Will comment on the original issue, but these examples are invalid: `Parsing[1002] Assignments can no longer be used as expressions.`

PR #1287, opened 2022-09-12
CLI server vs repo auth

lexidor:


Wilfred:
Definitely worth discussing more in the docs, thanks for doing this :)

PR #1289, opened 2022-09-12
Update redirect for new async-blocks path

Wilfred:


Atry:
Is the test failure relevant?

Wilfred:
I'm not sure, and I can't replicate it on a codespace. I'm going to merge to confirm that the master branch is consistent, and because this is an improvement.

PR #1304, opened 2022-10-19
Add examples for the function C\find

salvere002:
Screen Shot 2022-10-19 at 3 27 33 PM


kunalb6:
This is great, can you please add an example that highlights that first value is returned. Basically running the method on a vec with multiple values meeting the criteria and showing that the first is returned
Returns the "first" value of the given Traversable for which the predicate returns true, or null if no such value is found

salvere002:
Screen Shot 2022-10-19 at 5 47 18 PM

PR #1327, opened 2022-12-28
Initial modules documentation

jamesjwu:
This PR adds some initial sections on the modules feature. I will be continuously adding more parts to this and more information before we officially launch the feature.


guides/hack/56-modules/03-using-internal.md
```diff
@@ -0,0 +1,95 @@
+# Internal and module level visibility
+One goal of using modules is to separate a code unit's public and private interfaces. If a file is part of a module, its toplevel functions, classes, interfaces, traits, enums, and typedefs can be marked `internal`. An internal toplevel symbol cannot be referenced outside of its module. If a symbol is not internal, it is `public`, and it's part of the public API of the module. You can optionally use the `public` keyword on toplevel symbols to denote them as public.
```
Wilfred, line 2:
```suggestion
One goal of using modules is to separate public and private APIs. If a file is part of a module, its toplevel functions, classes, interfaces, traits, enums, and typedefs can be marked `internal`. An internal toplevel symbol cannot be referenced outside of its module. If a symbol is not internal, it is `public`, and it's part of the public API of the module. You can optionally use the `public` keyword on toplevel symbols to denote them as public. 
```

'Code unit' isn't a concept we use elsewhere in the docs (I think hackc uses the term internally a little bit, but its notion of a code unit will change with decl-based codegen). I'd also avoid 'interface' here to avoid confusion with `interface` declarations.

guides/hack/56-modules/03-using-internal.md
```diff
@@ -0,0 +1,95 @@
...
+        echo "foo!\n";
+    }
+    internal int $prop = 5;
+    public function bar(): void {
+        echo "bar!\n";
+    }
+}
+```
+
+An internal method or property can be called anywhere from within a module. `internal` replaces the visibility keyword of the method or propertiy (i.e. `protected`, `public` or `private`). Note that **class members' visibilities do not have to match the visibility of the class itself**: an internal class can have public methods, and a public class can have internal methods. You can think of the visibility on a toplevel symbol to represent where a symbol is allowed to be denoted, whereas the visibility of a method or property to represent where that individual member can be accessed.
```
Wilfred, line 33:
```suggestion
An internal method or property can be called anywhere from within a module. `internal` replaces the visibility keyword of the method or property (i.e. `protected`, `public` or `private`). Note that **class members' visibilities do not have to match the visibility of the class itself**: an internal class can have public methods, and a public class can have internal methods. You can think of the visibility on a toplevel symbol to represent where a symbol is allowed to be denoted, whereas the visibility of a method or property to represent where that individual member can be accessed. 
```

I don't understand "allowed to be denoted". Aren't both `FooInternal` and `$foo->internalProp` just access?

guides/hack/56-modules/03-using-internal.md
```diff
@@ -0,0 +1,95 @@
...
+    internal int $prop = 5;
+    public function bar(): void {
+        echo "bar!\n";
+    }
+}
+```
+
+An internal method or property can be called anywhere from within a module. `internal` replaces the visibility keyword of the method or propertiy (i.e. `protected`, `public` or `private`). Note that **class members' visibilities do not have to match the visibility of the class itself**: an internal class can have public methods, and a public class can have internal methods. You can think of the visibility on a toplevel symbol to represent where a symbol is allowed to be denoted, whereas the visibility of a method or property to represent where that individual member can be accessed. 
+
+If you try calling an internal method or accessing a property from outside of the module, you'll get a runtime error.
```
Wilfred, line 35: Presumably hh will catch this too?

guides/hack/56-modules/03-using-internal.md
```diff
@@ -0,0 +1,95 @@
...
+<<__EntryPoint>>
+function test(): void {
+    $x = new Foo(); // ok since Foo is a public class
+    $x->bar(); // prints "bar!"
+    $x->foo(); // error, foo is an internal method being called from outside the module.
+    $x->prop = 5; // error, $x::prop is an internal property being accessed outside of the module.
+}
+```
+
+We will go over the inheritance and override rules in a different section.
```
Wilfred, line 48:
```suggestion
```

guides/hack/56-modules/03-using-internal.md
```diff
@@ -0,0 +1,95 @@
...
+
+
+### Referencing internal symbols in your public API
+Public symbols within your module generally cannot reference internal symbols directly.
```
Wilfred, line 52:
```suggestion
Public definitions within your module generally cannot use internal symbols directly. 
```

guides/hack/56-modules/03-using-internal.md
```diff
@@ -0,0 +1,95 @@
...
+### Referencing internal symbols in your public API
+Public symbols within your module generally cannot reference internal symbols directly. 
+
+```hack
+module foo;
+internal class Secret {
+    internal function mySecret(): int {
+        return 42;
+    }
+}
```
Wilfred, line 60: Do you need the method in this example?

Also, it might be clearer to say `FooInternal` to avoid getting any crypto/password associations.

guides/hack/56-modules/03-using-internal.md
```diff
@@ -0,0 +1,95 @@
...
+    }
+    internal function mySecret(): int {
+        return 42;
+    }
+}
+public function foo(): Public { // 
+    return new Secret();
+}
+```
+At runtime, if code from another module calls `foo`, it will receive a Secret object. However, statically, any function that calls foo must respect the defined Public interface.
```
Wilfred, line 84: Perhaps `PublicFoo`? It's confusing to talk about a public type called `Public`.

```suggestion
At runtime, if code from another module calls `foo`, it will receive a Secret object. However, statically, any function that calls foo must respect the defined public interface.
```

guides/hack/56-modules/04-inheritance.md
```diff
@@ -0,0 +1,67 @@
+# Module level visibility and inheritance
+
+## Inheriting methods and properties
+This section discusses the relationships between Hack's object hierarchy and `internal`. In general, override rules for internal class members are consistent with that of private and protected: you can only override a class member with a visibility that's at least as visible as the previous.
```
Wilfred, line 4:
```suggestion
In general, override rules for internal class members are consistent with that of private and protected: you can only override a class member with a visibility that's at least as visible as the previous. 
```

guides/hack/56-modules/04-inheritance.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Inheriting toplevel symbols
+Unlike with methods and properties, within the same module, an internal class can override a public one, and vice versa. 
+
+```hack
+public class Bar {}
+internal class Foo extends Bar {}
+public class Baz extends Foo {}
+```
+You can think of this behavior as freely choosing which symbols in your module to export to your public API. Since the visibility of a toplevel entity only affects how it is statically referenced, no inheritance rules need to be applied when overriding them. Note that even if the toplevel symbol visibilities do not need to match any relationship, the individual members of those symbols do need to meet inheritance rules as discussed previously.
```
Wilfred, line 52:
```suggestion
You can think of this behavior as freely choosing which symbols in your module to export to your public API. Since the visibility of a toplevel entity only affects how it is statically referenced, no inheritance rules need to be applied when overriding them.
```

This last sentence kinda sounds like something that's relevant when implementing modules, but less useful for users.

guides/hack/56-modules/05-traits.md
```diff
@@ -0,0 +1,44 @@
+Traits have special behavior in Hack when used with modules. 
+
+## Marking traits internal 
+An internal trait can only be used within the module it's defined. Internal traits can have public, protected, or internal methods.
```
Wilfred, line 4:
```suggestion
An internal trait can only be used within the module where it's defined. Internal traits can have public, protected, or internal methods. 
```

Presumably trait methods can be private too?

guides/hack/56-modules/05-traits.md
```diff
@@ -0,0 +1,44 @@
...
+module bar;
+public class Bar {
+    use TFoo; // error, TFoo is internal
+}
+```
+
+
+
+## Public traits
+Since public traits have "copy-paste" behavior in that their implementations can be copied to any other module, these traits must be treated as if they are outside of any module for safety. Therefore, public traits cannot access any internal symbols or implementations from the module they are defined in.
```
Wilfred, line 24:
```suggestion
Since public traits have "copy-paste" behavior in that their implementations can be copied to any other module, these traits are treated as if they are outside of any module. Therefore, public traits cannot access any internal symbols or implementations from the module they are defined in. 
```

Rationale makes total sense for Hacklang folks, but less for users :)

guides/hack/56-modules/05-traits.md
```diff
@@ -0,0 +1,44 @@
...
+internal class Foo {}
+// Assume TFoo is defined as before
+public trait TBar {
+    use TFoo; // error, TBar is a public trait, it cannot access internal symbols in module foo
+    public function foo(): mixed {
+        return new Foo(); // error, TBar is a public trait, it cannot access internal symbols in module foo
+    }
+}
+```
+Public traits also cannot have internal members, as they may be used by classes outside of any module.
```
Wilfred, line 37: We don't generally use the term members for users, prefer 'methods or properties'.

guides/hack/56-modules/06-type-aliases.md
```diff
@@ -0,0 +1,44 @@
+Type aliases have special rules when it comes to internal types depending on which kind of type alias is used. 
+
+## Internal type aliases
+An opaque or normal type alias can be marked internal. Just as with classes, this means that the type can only be referenced from within the module.
```
Wilfred, line 4: I'd use concrete syntax rather than saying 'opaque or normal', as you're referring to a single kind of type type alias.

```suggestion
A type alias declared with `type` can be marked internal. Just as with classes, this means that the type can only be referenced from within the module.
```

guides/hack/56-modules/06-type-aliases.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Module-level type aliases
+We also introduce a new type alias known as a **module-level** type alias. You can create one using the syntax `module newtype TFoo as ... = ...`.
+A module-level type alias is opaque outside of the module it's defined in, and transparent inside (rather than just being opaque outside of the file it's defined in). Since they hide their implementations from outside the module, you can use internal types from within a module to implement them. They still cannot be constrained by internal types. 
+
+```hack
+module newtype FooModule = FooInternal; // ok 
+module newtype FooModuleErr as FooInternal = FooInternal; // error, FooInternal is an internal type, cannot be used as a constraint 
+```
+Since the entire point of module newtypes is to create an interface surrounding a module boundary, you cannot mark module newtypes themselves internal.
```
Wilfred, line 38:
```suggestion
Since the purpose of module newtypes is to create an interface surrounding a module boundary, you cannot mark module newtypes themselves internal. 
```

'The entire point' implies the user is being a little dumb to me.

guides/hack/56-modules/07-advanced-features.md
```diff
@@ -0,0 +1,53 @@
+This section details some advanced semantics and behaviors of modules.
```
Wilfred, line 1: Could you name this page 'closures and function pointers'? When users see 'advanced features' in the docs sidebar they won't know if the page is relevant to whatever question they have in mind.

guides/hack/56-modules/07-advanced-features.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Closures and function pointers
+
+Within a module, you can create a pointer to an internal function and pass it anywhere for use. Pointers to internal functions can only be created from within a module, but can be called and acccessed from anywhere.
+```hack
+module foo;
+internal function f() : void {
+    echo "Internal f\n";
+}
+public function getF(): (function():void) {
```
Wilfred, line 11:
```suggestion
public function get_f(): (function():void) {
```

Follow Hack naming conventions.

Wilfred (approved):
Great job: Lots of good examples and impressive thoroughness!

I've made some suggestions to make it even better, but feel free to merge whenever you're happy,

guides/hack/56-modules/02-defining-modules.md
```diff
@@ -0,0 +1,38 @@
...
+Currently, module bodies are empty. This will change in future versions of Hack as we build more support for modules and organizing their relationships. 
+
+# Module membership
+A module can be thought of as a logical structure of code organized into a list of files. To add a file to a module, use the `module foo;` syntax at the top of the file. 
+
+```hack
+// This file is now a member of module foo
+module foo;
+```
+A file can have at most one module membership statement, and the statement must be before any symbol in the file (File attributes and namespace declarations can appear before module membership statements, as they are not referenceable symbols). For clarity, module definitions (i.e. `new module`) cannot be placed within any module.
```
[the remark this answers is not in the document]
jamesjwu: Module definitions can technically be placed in files outside of any module, so you could put a module def and a class that doesn't have a module associated with it in the same file. Conventionally, this doesn't seem to be a very good code practice however.

Issue #1351, opened 2023-07-17
Mistake in Documentation, Getting Started: Starting a Real Project, Adding Functions or Classes


tnev:
Please complete the information below:

# Where is the problem?

Documentation, Getting Started: Starting a Real Project, Adding Functions or Classes section, code example

# What is the problem?

The code example is 

```hack
use namespace HH\Lib\Vec;

function square_vec(vec<num> $numbers): vec<num> {
  return Vec\map($numbers, $number ==> $number * $number);
}
```

It then notes "If you then run hh_client, it will tell you of a mistake... To fix this, change the return type of the function from vec<int> to vec<num>."

However the return type is already `num` in the example.

--------------------------------

Please don't change anything below this point.

--------------------------------

 - Build ID: HHVM=HHVM-4.164.0:HSL=v4.108.1:2023-07-17T02:20:09+0000:1fa47f258c6b68f8ec01899aa82fd6ffa0957109
 - Page requested: /hack/getting-started/starting-a-real-project
 - Page requested at: Mon, 17 Jul 2023 12:07:46 +0000
 - Controller: GuidePageController

PR #1352, opened 2023-07-17
Fix mistake in documentation, Getting Started: Starting a Real Project, Adding Functions or Classes

tnev:
https://github.com/hhvm/user-documentation/issues/1351

## Where is the problem?
Documentation, Getting Started: Starting a Real Project, Adding Functions or Classes section, code example

## What is the problem?
The code example is

```hack
use namespace HH\Lib\Vec;

function square_vec(vec<num> $numbers): vec<num> {
  return Vec\map($numbers, $number ==> $number * $number);
}
```

It then notes "If you then run hh_client, it will tell you of a mistake... To fix this, change the return type of the function from vec to vec."

However the return type is already num in the example.


Wilfred:
Thanks, D47734583.

